PR #2, opened 2016-09-29
Issue 1

lfaucon:
#1


README.md
```diff
@@ -3,7 +3,14 @@ Fabricating and running orchestration graphs
 
 ## Set up
 
-...
+- install meteor
+
+- run the commands
+`meteor npm install`
+`meteor npm install --save react react-dom`
```
houshuang: You don't need to run these commands, because --save means that they get written to package.json, and will be installed on meteor npm install in the future.

frog/imports/ui/Analytics.jsx
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react';
+ 
+export default class Analytics extends Component {
+  render() {
+    return (
+      <p>You are on the Analytics app</p>
+    );
+  }
+}
```
houshuang: If it is a stateless component, a cleaner way could be like this:

```js
const analytics = () => { 
  return(
      <p>You are on the Analytics app</p>
  )
}

export default analytics
```

frog/imports/ui/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+    super(props);
+    // The state of the App component indicates what the app is showing
+    // The display automatically updates itself on a call of setState()
+    // The Body component has app={this.state.app} as a property
+    this.state = {
+      app: "home",
+    };
+  }
+
+  renderButtons() {
```
houshuang: Here's an attempt at a rewrite using more pure components, and splitting out some of the logic:

```js
import React, { Component } from 'react';

import Body from './Body.jsx';
import HeaderButton from './HeaderButton.jsx';

const apps = ["home","repository","editor","engine","analytics"];

const AppButton = ({app, changeFn}) => { 
  const appobj = {
    id: app,
    onClick: () => changeFn(app)
  }
  return(
    <HeaderButton data={appobj}  />
  )
}

const Buttons = ({apps, changeFn}) => { 
  return(
    <div>
      {apps.map(app => <AppButton app={app} changeFn={changeFn} key={app}/>)}
    </div>
  )
}


// App component - represents the whole app
export default class App extends Component {
  constructor(props) {
    super(props);
    // The state of the App component indicates what the app is showing
    // The display automatically updates itself on a call of setState()
    // The Body component has app={this.state.app} as a property
    this.state = {
      app: "home",
    };
  }

  render() {
    console.log(this.state)
    return (
      <div>
        <div id="header">
          <h1>FROG - FABRICATING AND RUNNING ORCHESTRATION GRAPHS</h1>
          <Buttons apps={apps} changeFn={(app) => this.setState({app: app})}/>
        </div>
        <div id="body">
          <Body app={this.state.app} />
        </div>
      </div>
    );
  }
}
```

frog/imports/ui/Editor.jsx
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react';
+ 
+export default class Editor extends Component {
+  render() {
+    return (
+      <p>You are on the Editor app</p>
+    );
+  }
+}
```
houshuang: Prefer pure components :) (same goes for all the files below)

README.md
lfaucon: Done

frog/imports/ui/Analytics.jsx
lfaucon: For this I simply reproduced the way they do it in the Meteor + REACT tutorial. Furthermore These files are just the basis on which we will work and will become more complex in a near future.

houshuang: All the more reason to keep it simple and separate concerns... One thing we need to talk about is how to deal with internal state in the app. I need to read a bit more about how people in Meteor do that - I have good experiences with Redux and saw some blog posts about using Meteor with Redux, but need to read them before I recommend it :)

PR #9, opened 2016-10-05
Issue 4

palexlee:


frog/imports/ui/Repository.jsx
```diff
@@ -1,9 +1,230 @@
...
+export default class Repository extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      nameFilterText:"",
+      //The elements in the following lists are the attributes of the activities
+      //we want to filter
+      plane: [],
+      type: [],
```
lfaucon: It would have felt more natural to me to have `plane` and `type` be something like:
``` 
{
    Video: True/False,
    Quizz: True/False,
    Lecture: True/False,
}
{
    1: True/False,
    2: True/False,
    3: True/False
}
```

frog/imports/ui/Repository.jsx
```diff
@@ -1,9 +1,230 @@
...
+            {this.renderListActivities()}
+          </ul>
+        </div>
+      );
+  }
+
+
+}
+
+class Activity extends Component {
```
lfaucon: I would suggest having a separate file for this class, simply to factor the code. It's good to start like this now not to end up with 1000 lines long files in three weeks. The best would be to create a new folder `imports/ui/repository` containing `Repository.jsx` and `Activity.jsx`.

frog/imports/ui/Repository.jsx
```diff
@@ -1,9 +1,230 @@
...
+            //Iterates over properties of the object (as we don't know what they are)
+            //These special keys are created so that they are unique for each element of the array
+            properties.push(React.DOM.br({key: key + "-br" + i}));
+            properties.push(React.DOM.a({key: key + "-a" + i}, prop + ": "+ object[prop]));
+            ++i;
+          }
+        }
+
+    return properties;
+  }
```
lfaucon: Hey, I don't really like the use of `React.DOM.a` and `React.DOM.br` instead of the xml tags. What do you think of the code below ?

```
  renderObjectProperties(object) {
    class Property extends Component { 
      render () {
        return(
          <div>
            <br/>
            <a>{this.props.data.prop + ": "+ this.props.data.value}</a>
          </div>
        )
      }
    }

    return(Object.keys(object).map((prop)=>(
      <Property key={prop} data={{prop:prop, value:object[prop]}}/>
    )))
  }
```

frog/imports/ui/Repository.jsx
```diff
@@ -1,9 +1,230 @@
...
+    this.setState({
+      isClicked: !this.state.isClicked,
+    });
+  }
+
+  renderObjectProperties(object, key) {
+        var properties = []
+        var i = 0;
+        for(var prop in object) {
+          if(object.hasOwnProperty(prop)) {
```
lfaucon: Is this necessary ? `prop` just comes from `var prop in object`

frog/imports/ui/Repository.jsx
palexlee: I used the lists to make it easier to add a new filter (by just adding one line of code), but now that I think about, the way you suggested makes the code clearer about what kind of filters are available. I'll make the necessary changes.

frog/imports/ui/Repository.jsx
houshuang: I agree, I think we should stay consistent on using JSX

mj2905: I agree, that is better than before, there are no more strange unique keys to add to the React.DOM.a and .br components, and it's cleaner.

PR #15, opened 2016-10-10
Issue 4

mj2905:
I've modified just the onInput to the onChange methods used for the filter. (See commit message). Just to be done with this issue.


mj2905:
You can try now, I've merged both in the issue-4 branch.

PR #26, opened 2016-10-27
Cleaning up package.json, adding bcrypt, sorting

houshuang:
It's always complaining about bcrypt. When I did npm i --save, it also sorted the package.json, which I guess is nice.

### DIFF
--- a/frog/package.json
+++ b/frog/package.json
@@ -2,20 +2,18 @@
   "name": "frog",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


lfaucon:
Thanks, I think I did the exact same thing at the same time as you. You can pull the develop branch. My problem with bcrypt came from the python version (incompatibilities with python3), solved using `meteor npm install --python=python2.7`

PR #42, opened 2016-11-23
Using yarn and symlinks to decrease install time and size

houshuang:
I reorganized the mono-repo to use yarn, and to only install the dev-dependencies once, and symlink them into all the other repositories. 

This approach takes 50 seconds, and results in a total size of 200 MB. The old approach takes 4 minutes, and generates 560 MB of data. Note that this will only increase once we add twenty more operators and activities etc :)

I couldn't figure out how to check for the presence of a binary, but this script should also work perfectly well with npm instead of yarn, so could you add a test that it uses npm if yarn is not installed?

### DIFF
--- a/ac/ac-ck-board/package.json
+++ b/ac/ac-ck-board/package.json
@@ -7,24 +7,16 @@
     "watch": "watch 'npm run build' ./src",
     "prepublish": "npm run build"
   },
-  "devDependencies": {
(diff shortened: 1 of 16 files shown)


houshuang:
Enabling npm should also make the tests pass again.

houshuang:
Tests pass! :)

PR #43, opened 2016-11-28
Issue 41

KhalilMrini:


frog/imports/ui/StudentView.jsx
```diff
@@ -108,6 +109,62 @@ const StudentView = ( { user, sessions, products } ) => {
   </div>
 )}
 
+var TimerExample = React.createClass({
```
lfaucon: Can you rewrite this class to be consistent with the rest of the code? 
``` javascript
class Timer extends component {
    constructor(props){ 
        super(props);
        this.state = { elapsed: 0 }
    }
    componentDidMount() { ... }
    ...
}
```

KhalilMrini:
Sure! I will address this soon

frog/imports/ui/StudentView.jsx
```diff
@@ -108,6 +109,62 @@ const StudentView = ( { user, sessions, products } ) => {
...
+        clearInterval(this.timer);
+    },
+
+    tick: function(){
+
+        // This function is called every 50 ms. It updates the 
+        // elapsed counter. Calling setState causes the component to be re-rendered
+
+        this.setState({elapsed: new Date() - this.props.start});
+    },
```
houshuang: Although given that we might have several components that need to show clock information, it would be interesting to write a higher-level function which takes a React component and makes it regularly update. Not a priority, but I might play with this :)

houshuang: OK, I had a bit of time in a boring presentation - check out #44 :)

houshuang:
We don't need server-side code unless we want the server to take some action based on time (ie. automatically switch to the next activity based on time, etc).

houshuang:
@lfaucon I think we need to think about separating activity parameters and OG parameters. Every activity needs a name, and perhaps a start/end date etc, but these will not be passed to the activity, only useful for engine execution. So no point in each activity specifying these. In Operators, I'm currently merging in "source" and "target" into the form provided by the operator config, but this is not a great long-term strategy - some operators/activities might provide their own parametrization component, and anyway this data should be stored separately... perhaps

```
activity: {
  _id:
  created_at:
  activity_package: 'ac-form',
  package_parameters: {quiz questions etc),
  og_parameters: {start_time, end_time, name etc}
}
```

?

PR #46, opened 2016-12-01
Graphs

lfaucon:
This version seems to not have too many bugs, but I still have a lot on my todo list to improve it. We can merge it. 

- [ ] Enable to select an element of the graph (activity/operator) to see its config/edit it 

- [ ] Delete button for in-graph activities

- [ ] Remove out-operators

- [ ] Have a field `sessionId` in results and products OR make copies of all the content of a graph for each session

- [ ] Lines instead of boxes for planes

- [ ] cleaning up the api

- [ ]  remove usage of `status: 'OUT'/'IN' `


frog/imports/api/graphs.js
```diff
@@ -3,8 +3,16 @@ import { uuid } from 'frog-utils'
  
 export const Graphs = new Mongo.Collection('graphs');
 
-export const addActivity = () =>
+export const addGraph = () => {
+  const id = uuid()
```
houshuang: ah, I see you are returning the id... makes sense. carry on :)

if we're doing this in all functions, we could make a little HOF

```
const withID = (fn) => {
    const id = uuid()
    return (
        (args) => {
            fn(...[args] + [id])
            return (id)
        }
    )
}
```

and then we can do 

```
export const addGraph = withId( (id) => 
  Graphs.insert({
    _id: id,
    name: 'untitled',
    created_at: new Date()
    }) 
)
```

just for fun :)

frog/imports/api/activities.js
```diff
@@ -14,13 +15,45 @@ export const addActivity = (activity_type, data, id) => {
...
+
+export const addGraphOperator = (params) => {
+  const id = uuid()
+  Operators.insert({...params, created_at: new Date(), _id:id, status: 'IN'})
+  return(id)
+}
+
+export const copyActivityIntoGraphActivity = (graphActivityId, fromActivityId) => {
+  const fromActivity = Activities.findOne({_id:fromActivityId})
+  Activities.update(graphActivityId, {$set: {data: fromActivity.data, activity_type: fromActivity.activity_type}})
```
houshuang: Maybe we should have a parent tag that points to where the activity was copied from? So we can say "someone updated this activity, merge to graph etc" in the future?

frog/imports/api/graphs.js
```diff
@@ -3,8 +3,16 @@ import { uuid } from 'frog-utils'
...
   Graphs.insert({
-    _id: uuid(),
-    created_at: new Date() 
+    _id: id,
+    name: 'untitled',
```
houshuang: might be cleaner to use a default argument

`addGraph = (name = 'untitled') => {}`

that way we can use the function in the future if we decide to create a new graph that already has a name etc.

frog/imports/api/activities.js
lfaucon: Good idea! It's done

frog/imports/api/activities.js
```diff
@@ -14,13 +15,45 @@ export const addActivity = (activity_type, data, id) => {
...
+}
+
+export const copyOperatorIntoGraphOperator = (graphOperatorId, fromOperatorId) => {
+  const fromOperator = Operators.findOne({_id:fromOperatorId})
+  Operators.update(graphOperatorId, {$set: {data: fromOperator.data, operator_type: fromOperator.operator_type, type:fromOperator.type}})
+}
+
+export const removeGraph = ( graphId ) => {
```
lfaucon: I was wondering, because it also uses Operators and Activities (both in api/activities.js). I agree with the need of restructuring/cleaning. Added to todo list

frog/imports/api/graphs.js
lfaucon: Good point! I've never used default arguments in javascript so it does not come naturally for me. Done.

frog/imports/ui/ActivityEditor.jsx
```diff
@@ -126,7 +126,7 @@ class ActivityBody extends Component {
 
 export default createContainer(() => {
   return {
-    activities: Activities.find({}).fetch(),
-    operators: Operators.find({}).fetch(),
+    activities: Activities.find({status:'OUT'}).fetch(),
+    operators: Operators.find({status:'OUT'}).fetch(),
```
lfaucon: Added to the todo list. I even want to get rid of it and make the difference on wether an activity has a field `graphId` or not

frog/imports/ui/TeacherView.jsx
```diff
@@ -83,25 +85,49 @@ const SessionController = ( { session, activities } ) => {
...
-)}
+const SessionControllerContainer = createContainer(({ session }) => {
+  return({
+    session: session,
+    activities: session? Activities.find({graphId: session.graphId}).fetch() :null
+  })
+}, SessionController)
+
+const SessionList = ( { sessions, graphs } ) => { 
+  var selectedGraph = graphs[0] ? graphs[0]._id: null
```
lfaucon: If the list of graphs is empty graph[0] is undefined, so it crashes on graphs[0]._id

frog/imports/ui/TeacherView.jsx
```diff
@@ -83,25 +85,49 @@ const SessionController = ( { session, activities } ) => {
...
-        </li>
-      ) 
-    } </ul>
-    <button className='btn btn-primary btn-sm' onClick={addSession}>Add session</button>
-  </div>
-)}
+const SessionControllerContainer = createContainer(({ session }) => {
+  return({
+    session: session,
+    activities: session? Activities.find({graphId: session.graphId}).fetch() :null
```
lfaucon: Space before `?`? You're becoming so French ;) 

I won't fix it everywhere, but I'll go for the more spaces going forward.

frog/imports/ui/TeacherView.jsx
```diff
@@ -121,26 +147,26 @@ const DashView = ({ user, logs }) => {
   }
 }
 
-const TeacherView = ( { activities, sessions, logs, user } ) => { return(
+const TeacherView = ( { graphs, sessions, logs, user } ) => { return(
```
lfaucon: Done in a few places, will do going forward

frog/imports/ui/TeacherView.jsx
```diff
@@ -83,25 +85,49 @@ const SessionController = ( { session, activities } ) => {
...
+  return({
+    session: session,
+    activities: session? Activities.find({graphId: session.graphId}).fetch() :null
+  })
+}, SessionController)
+
+const SessionList = ( { sessions, graphs } ) => { 
+  var selectedGraph = graphs[0] ? graphs[0]._id: null
+
+  const changeGraph = (event) => {
```
lfaucon: I changed it to a class temporarily.

frog/imports/ui/GraphEditor.jsx
```diff
@@ -0,0 +1,252 @@
+import React, { Component } from 'react';
+import { createContainer } from 'meteor/react-meteor-data'
+import { Meteor } from 'meteor/meteor';
+import Draggable from 'react-draggable';
+
+import { Activities, Operators, addGraphActivity, addGraphOperator, copyActivityIntoGraphActivity, copyOperatorIntoGraphOperator, dragGraphActivity, removeGraph } from '../api/activities';
+import { Graphs, addGraph, renameGraph } from '../api/graphs';
+
+import jsPlump from 'jsplumb';
+import { $ } from 'meteor/jquery';
```
lfaucon: Removed it in package.json. Yes, JsPlumb uses it

houshuang:
I approved your changes - I leave it up to you to decide if there is anything else you need to do before merging. Then you can merge my pull request (if you approve) and we can continue improving it.

PR #48, opened 2016-12-08
Graphs fixes

lfaucon:
I implemented some of the changes that I mentioned in the last PR:

- Allow to remove activities added to a graph
- Not using a field `status = 'IN' | 'OUT'` in activities to distinguish in-graph and out-graph activities
- Lines instead of boxes for the editor. Now that I see, I am even more convinced that I was right 😄  
- Copying activities and operators of a graph when creating an activity


frog/imports/ui/GraphEditor.jsx
```diff
@@ -28,20 +28,43 @@ jsPlumbRemoveAll = () => {
 
 jsPlumbDrawAll = (activities,operators) => {
   if(jsPlumbInstance){
+    console.log('for each activity')
```
houshuang: argh:)

frog/imports/ui/GraphEditor.jsx
```diff
@@ -135,11 +172,13 @@ class GraphEditorClass extends Component {
     // creating an activity in the graph
     planeNames.forEach(plane => {
       $('#'+plane).dblclick(event => {
+        console.log('hello')
```
houshuang: booh

PR #52, opened 2017-01-09
Graph creator

mj2905:
We are fixing some lint issues, but you can review.


lfaucon:
Thanks a lot! I start looking at it now. As it's quite big, expect my feedback only tomorrow :)

frog/imports/ui/graph/Graph.jsx
```diff
@@ -0,0 +1,745 @@
...
+  handleStop,
+  sourceOperator,
+  targetOperator,
+  loaded,
+  plane,
+  graphId,
+  moveCursor,
+  cursor,
+  scale,
+  activitySourceClicked}) => {
```
lfaucon: That's a lot of props ;)  I understand that there are many thing going on. In that case, be sure to not use props when other solutions would be more efficient/readable. For example you can use something like Redux (http://redux.js.org/, I think Stian mentioned it) to save global states to be used by your react components. That way, you don't have to pass 20 props through 5 layers of nested components.

frog/imports/ui/graph/DraggableAc.jsx
```diff
@@ -0,0 +1,215 @@
...
+                <span>
+                  {this.props.editorMode && duration >= minRealBox ?
+                    <a
+                      onClick={(event) => this.props.delete(activity)}
+                      >
+                      <i className='fa fa-trash'  style= { { color: 'white' } }/>
+                    </a> : ''
+                  }
+                </span>
+              }
```
lfaucon: Here you have:
```Javascript
 +              {
 +                <span>
 +                  {this.props.editorMode && duration >= minRealBox ?
 +                    <a
 +                      onClick={(event) => this.props.delete(activity)}
 +                      >
 +                      <i className='fa fa-trash'  style= { { color: 'white' } }/>
 +                    </a> : ''
 +                  }
 +                </span>
 +              }
```
Wouldn't it be the same to have:
```Javascript
 +                  {this.props.editorMode && duration >= minRealBox ?
 +                    <a onClick={(event) => this.props.delete(activity)}>
 +                      <i className='fa fa-trash'  style= { { color: 'white' } }/>
 +                    </a> : ''
 +                  }
```
?

I find a lot of useless brackets or divs in your code.

frog/imports/api/activities.js
```diff
@@ -99,6 +150,15 @@ Meteor.methods({
     Activities.remove({})
   },
 
+  'graph.flush.operator': (operatorId) => {
+    Operators.remove({ _id: operatorId })
+  },
+
+  'graph.flush.operators': (activityId) => {
+    Operators.remove({ 'from._id': activityId })
+    Operators.remove({ 'to._id': activityId })
```
mj2905: We only copied the plane and the _id field of the source and target activities, as an activity position, duration and other properties may change. In that case we would have to always modify the operators which is tedious. Thus, we copied only the plane and the id as both were used a lot. 
In Graph.jsx :
```
addNewOperator = (target) => {
    if (this.state.currentSource != null && target != this.state.currentSource) {
      this.setState({currentSource:null, loaded:false});
      const fromAc = {plane: this.state.currentSource.plane, _id: this.state.currentSource._id}
      const toAc = {plane: target.plane, _id: target._id}
      addGraphOperator({_id: uuid(), graphId: this.props.graphId, from: fromAc, to:toAc})
    }
  }
```

frog/imports/ui/graph/DraggableAc.jsx
```diff
@@ -0,0 +1,215 @@
...
+
+const defaultTime = 10
+
+const circleRadius = 6
+
+const rndMargin = 12
+
+const minRealBox = 45
+// #fe8181 ligth red
+const divStyle = (duration) =>
```
mj2905: Since a month we want to do it, but everytime there was something else to do, so that after being reported everytime it was never done.

frog/imports/ui/graph/DraggableAc.jsx
mj2905: For this case I think too that it gives the same result. But sometimes for example we have useful divs as we need to have absolute positioning into relative positioning.

frog/imports/api/activities.js
lfaucon: Ok, I see, so you will use the field 'from._id' to retrieve the corresponding activity and find its position, duration, etc. I would have found more logical to get the plane information from the activity instead of the operator (planes are in the definition of activities, not of operators). But I understand that you had to find work-arounds against our incomplete always-changing implementation ;)

PR #53, opened 2017-01-20
Issue 49

KhalilMrini:


KhalilMrini:
I don't know why it fails? The project works on my computer

houshuang:
It fails because of a bunch of eslint error. Try running eslint locally.

KhalilMrini:
I cannot fix the last two problems:
   26:20  error  'MSBlobBuilder' is not defined  no-undef
  253:20  error  'FileReader' is not defined     no-undef
Because these types are needed in the code. I saw the eslint documentation but I don't see how to fix it?

KhalilMrini:
I applied an ignore comment on the advice of Louis.

PR #65, opened 2017-02-06
Rewrite timescale

lfaucon:
Just for code prettiness. Hoping it would give you a smile @houshuang


houshuang:
Beautiful! Feel free to rewrite all my code like this :)

Of course this code will also change when you parametrize the length of the graph... 

PS: Not sure how you did your pull requests, that caused every single commit to show up in all of them, but as long as the "files changed" view shows the correct diff, I guess it's OK.

PR #67, opened 2017-02-07
Graph list

lfaucon:
Add graph list next to graph editor and allow saving multiple graphs

### DIFF
--- a/frog/imports/ui/svg/index.js
+++ b/frog/imports/ui/svg/index.js
@@ -2,15 +2,12 @@ import React, { Component } from 'react';
 import { Provider } from 'mobx-react';
 import Form from 'react-jsonschema-form';
 import styled from 'styled-components';
-import { createContainer } from 'meteor/react-meteor-data';
-
-import { removeGraph } from '../../api/activities';
-import { Graphs, addGraph } from '../../api/graphs';
 
 import { connect, store } from './store';
 import Graph from './Graph';
 import Rename from './Rename';
 import SidePanel from './SidePanel';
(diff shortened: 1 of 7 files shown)


frog/imports/ui/svg/index.js
```diff
@@ -2,33 +2,38 @@ import React, { Component } from 'react';
 import { Provider } from 'mobx-react';
```
houshuang: I would prefer if you split this file into three. Have a top level container, which links to the actual graph component (separate file), and perhaps it does the switching between the graph list in the sidebar, or the component in the sidebar (which could be two separate files). Not sure if it's cleaner to do the data fetching in the main file, or in each separate file (connect / createContainer) ... anyway, let's try to keep each component as short and easy to read as possible. Feel free to create subdirs, for example Mainpage/index.js, Mainpage/GraphComponent.js, Mainpage/ComponentSidebar.js and Mainpage/Graphlist.js or something like that.

Also I think we should redesign the graph list - perhaps it could be a drop down menu instead, or a separate page etc. BUt that can be done after merging this one.

lfaucon:
When you have no graph selected or created, it lets you work on graph with id=1, which you can find again when refreshing :) I'll take my code from the previous version that write in the user object (in Mongo) your current graph and takes you directly to it

lfaucon:
Fixed, previous issue:
The choice of current graph editing is done in the following way:
```
Was editing a graph that still exists ? => graphId :
  Is there at least one graph in Mongo ? => graphId :
    createNewGraph 
    => graphId
```

PR #71, opened 2017-02-15
Upgrading packages

houshuang:
I needed to upgrade some Meteor packages to enable proper parsing of Flow type syntax, and I decided to upgrade other packages at the same time.

Also had to make a tiny change to the Flow type of ActivityRunnerT to enable tests to pass (same issue you remarked).

Might be good idea to rebase your branch off develop again after accepting this, so we're both working with the newest versions of packages, and can make sure everything works well etc. 

PS: Meteor keeps bugging me to upgrade with --patch, but the newer version gives lots' of errors. I have filed a bug, but let's wait with upgrading Meteor for now.


houshuang:
PS: After merging, you should do git clean -fdx, and ./initial_setup.sh again.

PR #80, opened 2017-02-21
Argue dataset

lfaucon:
DONE:
- New sample dataset (to be improved) for the ArgueGraph script
- Modifications to `ac-quiz` to make use of the new Engine API
- Modifications to `ac-quiz` to make it a MCQ with justifications (more suitable for argue script)
- New op-argue (still to be implemented)
- Example of connections between activities (when an operator does not make much sense, but it gives somehow more work to the activity packages to handle the data)  

TO TEST:
- **Admin > Load Argue Graph**
- Enjoy the view in **Graph Editor**
- Have 2 to 4 students joining the session
- Have teacher click **Start**, then **Next** and students fill up the forms as they come
- Enjoy the new collaboration layout on the third activity (edit only on the left side)

COMMENTS:
- Activities Packages receive as props a function `saveProducts` that they must use if they want their produced data to be used somewhere. We could imagine adding to the API in the same way `saveLog` and `saveStatus` for the traces and the dashboard.
- For closing an activity to an individual studnet once he submitted, you previously checked for a result (*same thing as product*?). I expect that an activity package specifies if it should be open or not by itself: For example when a students submits the form, we use `reactiveFn` to write to a key `STATUS+studentID` that the form is completed and then hide the activity.

### DIFF
--- a/ac/ac-quiz/src/index.js
+++ b/ac/ac-quiz/src/index.js
@@ -107,11 +107,13 @@ export const ActivityRunner = (props: ActivityRunnerT) => {
     ? reactiveKey['DATA' + userInfo.id]
     : null
 
-  const partnerId =  socialStructure
+  const findPartner = socialStructure
     ? Object.keys(socialStructure).find(
       id => !(id === userInfo.id) && socialStructure[id].group === groupId
     )
-    : userInfo.id
+    : null
+
+  const partnerId = findPartner ? findPartner : 'NO_ID'
 
   const partnerFormData = reactiveKey
     ? reactiveKey['DATA' + partnerId]
--- a/ac/ac-quiz/src/index.js
+++ b/ac/ac-quiz/src/index.js
@@ -1,23 +1,24 @@
 // @flow
+
 import React from 'react';
 import Form from "react-jsonschema-form";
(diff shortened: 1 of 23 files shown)


lfaucon:
I planned to do that on the next PR, so you will not have a hard time finding my modifications in the middle of thousands of syntax modifications.

ac/ac-quiz/src/index.js
```diff
@@ -39,52 +40,126 @@ export const config = {
...
+
+  const findPartner = socialStructure
+    ? Object.keys(socialStructure).find(
+      id => !(id === userInfo.id) && socialStructure[id].group === groupId
+    )
+    : null
+
+  const partnerId = findPartner ? findPartner : 'NO_ID'
+
+  const partnerFormData = reactiveKey
```
houshuang: For these kinds of things you don't need a ternary, you can do

```
const partnerId = findPartner || 'NO_ID'
const partnerFormData = reactiveKey && reactiveKey['DATA' + partnerId]
```

lfaucon: I'm surprised to see that flow does not complain on that. I remember it did not like that I multiply Boolean with Integers :smile:

Issue #92, opened 2017-03-14
TODO before Boston


lfaucon:
## TODO
- [ ] -> Have a User module which is called instead of Meteor.loggedInUser (maybe use MobX?)
- [ ] -> Parse user from URL and implement user logging in through for example /#graph?user=Stian
- [ ] -> Script for the demo (Stian)
- [ ] -> Technical implementation of simulation for the demo (Stian) 

## BONUS
- [ ] -> Word cloud / differential word cloud dashboard (Louis, bonus)
- [ ] -> Read only version of the graph with a progress bar for teacher orchestration and visualization of students groups (any) (Stian, very bonus)

Issue #81, opened 2017-02-22
Fixing drawPath to look nicer


houshuang:
Right now there is a problem when paths are drawn with the start later than the beginning, like this operator

![screen shot 2017-02-22 at 11 27 55](https://cloud.githubusercontent.com/assets/61575/23207572/0210e796-f8f2-11e6-9286-41f501b3767f.png)

should be quite easy to fix. 

Probably other things we can do too, for example links to activities should go to their front, not their back (like this link which merges with the activity body at the middle)

![screen shot 2017-02-22 at 11 29 28](https://cloud.githubusercontent.com/assets/61575/23207614/3183a07c-f8f2-11e6-8944-eb95b2dc20c5.png)

I'm not touching drawPath.js in my rewrite, so feel free to play with it :)

lfaucon:
I will gladly play with draw path, once we have a bit more robust graph editor. I see only one path with the start later than the beginning which is the connection between the first individual quiz and the social operator and I can't see what is the problem with this connection.

The main problem I see concerns the second point that you raise; Connections **to** activities/operators should connect to the left side (input), while connections **from** activities/operators should connect to the right side (output).

PR #93, opened 2017-03-14
Fix paths

lfaucon:
#81 #92 

Draw connections in a way that makes clear inputs and outputs for our dataflow

## comments
- The mouseOver bug on activities was due to the mouse mouse being over the connection and not over the activity. Thus the weird behavior of removing the highlight when the mouse goes backward.

- I tried drawing connections on top of activities, but it looks bad.

- The curves drawn by the `<path>` are bezier curves. They take 4 points to be described, starting point, endpoint and direction of start and direction of end (image from wikipedia). 

bezier

### DIFF
--- a/frog/imports/ui/GraphEditor/store/activity.js
+++ b/frog/imports/ui/GraphEditor/store/activity.js
@@ -1,9 +1,11 @@
 // @flow
+
 import { observable, computed, action } from 'mobx';
 import cuid from 'cuid';
 import { store } from './index';
 import Elem from './elemClass';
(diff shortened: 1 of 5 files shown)


houshuang:
Looks quite good, however there is a problem in the quite common case where two adjacent activities are connected by an operator... With some space, it looks like this

screen shot 2017-03-15 at 06 56 47

But with the activities being adjacent, like this:

screen shot 2017-03-15 at 06 56 53

which looks misleading...

PR #94, opened 2017-03-14
Fix graph movement

houshuang:
Graph startTime is always an integer (not float), which should help with activity alignment. Activities should not be above each other when they don't need to, and dragging activities "across" (swap) should work better. The activity swap still needs better UI feedback, but at least it should be fully functional.


lfaucon:
The swap works much better than before, but I still can find some rounding bugs:

I think the picture below shows two different rounding bugs: First for the x coordinate of the dragline and second for the distance between the start of the activity being dragged and the end of the previous. 

screenshot from 2017-03-15 11-34-48

houshuang:
Thanks for your comments, they were really useful. I think the latest version should be much more robust, let me know if you can break it :)

PR #97, opened 2017-03-15
This fixes some bugs with keyboard shortcuts

houshuang:
- Only allows keyboard shortcuts while mouse is over main graph (not minimap)
- Filters keyboard shortcuts with Cmd etc (to avoid Cmd-P for printing triggering operator)
- Avoid default, to avoid both creating operator and entering into the rename field, if rename field is active

### DIFF
--- a/frog/imports/ui/GraphEditor/Graph.js
+++ b/frog/imports/ui/GraphEditor/Graph.js
@@ -75,30 +75,3 @@ export default connect((
       <ScrollFields width={width} height={height} />}
   </svg>
 ));
-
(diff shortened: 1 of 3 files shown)


lfaucon:
I have the issue that when I create a new activity, if the mouse leaves there naming text area, then all keys work fine for typing the name except for s, p and backspace.

houshuang:
Regarding your issue, once we have a nicer rename box for graphs, we can track whether a user is renaming a graph. Then we'll probably have to import the store to let the rename component update ui state or something.

PR #98, opened 2017-03-15
This adds a help modal for the graph editor

houshuang:
The modal is pulled up with '?'. We can adapt the text going forwards.
We could also automatically show the modal the first time a user is
editing a graph (not currently the case).

### DIFF
--- a/frog/imports/ui/GraphEditor/index.js
+++ b/frog/imports/ui/GraphEditor/index.js
@@ -31,7 +31,7 @@ const keyDown = e => {
   if (store.mode === 'rename') {
     return;
   }
-  if (e.ctrlKey || e.altKey || e.metaKey || e.shiftKey) {
+  if (e.ctrlKey || e.altKey || e.metaKey) {
     return;
   }
   if (e.keyCode === 27) {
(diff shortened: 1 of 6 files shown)


houshuang:
PS This is development with nice libraries/framework etc at its best - it took me something like 15 minutes to add this feature (most of it was writing the text :)), the code changes are minimal, and it works great and looks great... Not brain surgery, but still something that would have been non-trivial with jquery, without Bootstrap styling etc.

frog/imports/ui/GraphEditor/index.js
```diff
@@ -28,28 +28,35 @@ export default class AppClass extends Component {
...
     store.ui.cancelAll();
     store.ui.unselect();
   }
   if (e.keyCode === 8) {
-    // backspace
+    // backspace - delete
     e.preventDefault();
     store.deleteSelected();
   }
+  if (e.keyCode === 191 && e.shiftKey) {
```
lfaucon: Is there an other way than using `keyCode` ? On my swiss keyboard, the help opens if I click on the key showing `§` and `°` and not `?`. Works with the expected key if I switch to English keyboard. 

Suggested fixes:
- Open the help with a button instead of a keyboard shortcut
- use the letter `H`
- see what we can do with `charCode` instead of `keyCode`

houshuang:
Good point - seems like it's actually worth adding a dependency here to handle this properly... you need to do npm install in frog/frog

frog/imports/ui/GraphEditor/index.js
houshuang: Addresses by using proper library

PR #99, opened 2017-03-16
Making it possible to create read only graph store

houshuang:
I turned state into a setter and getter, which checks whether store is read-only or not. All actions check whether
they are in the correct "mode", so this basically makes the entire store read-only (or not). 

To test, change new Store() to new Store(true) in /store/index.js


houshuang:
It now uses setId instead, which means that it can be turned on and off at will. You can test it in the console, by checking that all actions (moving, dragging, adding etc work) and then typing
```
window.store.setId(window.store.graphId, true)
```

and seeing that you can do nothing (other than zooming, and panning).

Issue #75, opened 2017-02-15
Adapt operators and activities to the new engine


lfaucon:
## Activities
- ck-board
    - [ ] robust
    - [ ] correct behavior
- form
    - [X] robust
    - [X] correct behavior
- iframe
    - [X] robust
    - [X] correct behavior
- jigsaw
    - [ ] robust
    - [X] correct behavior
- quiz
    - [ ] robust
    - [X] correct behavior
- text
    - [X] robust
    - [X] correct behavior
- video
    - [X] robust
    - [X] correct behavior

## Operators
- agg-ck-board
    - [ ] robust
    - [ ] correct behavior
- agg-text
    - [ ] robust
    - [ ] correct behavior
- argue
    - [X] robust
    - [X] correct behavior
- jigsaw
    - [X] robust
    - [X] correct behavior
- like-with-like
    - [X] robust
    - [X] correct behavior

houshuang:
Are you happy with this one now? @lfaucon

lfaucon:
Not yet. Still need to work on the activity types

Issue #74, opened 2017-02-15
Clean - Remove code that is not used anymore


lfaucon:

houshuang:
Can we close this?

lfaucon:
I removed a lot of unused code in #100 but certainly not all of it

PR #100, opened 2017-03-17
Operators

lfaucon:
Work for #74, #75 and #92 

- Jigsaw operator working as expected (robust and efficient)
- Argue Operator working approximately as expected (robust and efficient)
- rework on TeacherView and StudentView
- removed a lot of dead code


frog-utils/src/index.js
```diff
@@ -47,8 +47,18 @@ export const splitAt = (i: number, xs: Array<any>): Array<Array<any>> => {
   return [first, second];
 };
 
-export const shuffleList = (xs: Array<any>): Array<any> =>
-  xs.slice(0).sort(() => 0.5 - Math.random());
+export const shuffleList = (xs: Array<any>): Array<any> => {
```
houshuang: Would it make sense to replace this with lodash.shuffle? https://lodash.com/docs/4.17.4#shuffle

op/op-argue/src/index.js
```diff
@@ -13,15 +13,68 @@ export const config = {
   properties: {}
 };
 
+const shuffleList = (xs: Array<any>): Array<any> => {
```
houshuang: Why isn't this imported from frog-utils?

PR #105, opened 2017-03-20
Sidepanel for editing activities

houshuang:
First work on showing activities in a sidebar, currently works - you can choose a type, or edit an existing type. 

- [x] Save only happens on submit (probably autosave on change is more appropriate)
- [x] Looks pretty bad in narrow sidebar, do we need minimal config + expanded config in modal?
- [ ] Should revisit config objects in activities, name and time should be removed, etc.
- [x] Need to add operator configuration

(You don't need to merge this yet, but this PR let's you track my work - it should currently be fully functional, albeit ugly)


houshuang:
Tried to just tighten up the layout a little bit - can be made much nicer. It would be nice to have configurable size of the graph (it could extend if we are not editing any activities, etc).

Here's how it looks now:
screen shot 2017-03-20 at 17 18 29

screen shot 2017-03-20 at 17 18 22

houshuang:
(one day we should get the login name back on the toolbar line too... it can't be that hard :))

PR #107, opened 2017-03-22
Nicer sidebars for activities and operators

houshuang:
Mainly adding rename, which is in sync with graph, and cleaning up display a bit. The code is a bit convoluted, I might go back and refactor later, but for now it should work fine, and I think it's better to merge this and begin playing with it (next step is trying to build graphs using operators and activities, and making sure ac/op work well, but in the process, I might decide to change the views further) - I might go back and refactor a bit later.


frog/imports/ui/GraphEditor/store/uiStore.js
```diff
@@ -1,14 +1,23 @@
...
 import type { Elem } from './store';
 import Operator from './operator';
 
 export default class uiStore {
+  constructor() {
+    autorun(() => {
+      if (this.selected) {
+        this.setStickySelected(this.selected);
+      }
+    });
```
lfaucon: - Why do we need stickyselected ? 
- What does this piece of code in the constructor do? 

The sidePanel crashes if the graph is deleted while something is selected

houshuang: Right now, I am quite liberal about using cancelAllSelections or what it's called, and the result is for example that the sidebar (which relies on the currently selected Activity/Operator) would disappear - for example by double-clicking to rename an activity, and pressing enter to complete the rename (not so common in real life perhaps, but common when I want to demonstrate how the two title fields are linked ;))...

autorun is a mobx functionality that tracks all the dependencies (ie. all the mobx variables accessed when the function is first run), and then is automatically rerun every time any of them change - in this case, it's evaluated any time this.selected changes. The idea of stickySelected was to have a variable that tracks the last thing that was selected, whether or not the selection was cancelled afterwards.

It's quite possible we could do this more elegantly in the future :) Thanks for fixing the crashing when deleting graph - I'm thinking that problem is also related to the fact that we have components that span stores (Meteor and MobX), and I think it's quite possible that we need to do another smaller refactor some time in the future. But I think for now, as long as it works, and let's us rapidly refactor the kind of interfaces we want, it's OK - I will aim for as good code as possible of course, but I will not be too worried if it ends up being a bit ugly or copy and paste for now, as long as user experience goes up...

Issue #111, opened 2017-03-28
More granular/well placed time marks on graph


houshuang:
Also, the last minute indication sometimes half disappears (when it's 120 minutes for example), or sometimes there is none (because the end time is not divisible by 15)... One thing we could do is to have more time labels at a higher zoom level (if you're zoomed all the way in, 15 minutes might take the whole width of the screen), and perhaps we also special case that we render the final minute count, no matter how much it is (even if it's 97), so that it's properly displayed. A final idea is to add a little bit of "margin" at the end of the graph, so that the final line can be clearly shown.

PR #123, opened 2017-03-30
Adaptive axis labeling, fixing #111

houshuang:


frog/imports/ui/GraphEditor/fixedComponents.js
```diff
@@ -58,21 +58,29 @@ export const LevelLines = connect(({
...
-        </g>
-      );
-    })}
-  </g>
-));
+}) => {
+  let divider = Math.round(5 / scale) % 5 * 5;
+  if (divider === 0) {
+    divider = 20;
+  }
```
lfaucon: The `% 5` followed by handling the case of `0` can be made a lot more clear with the `between` method. Example of code that does the same thing: 
```
const divider = Math.round(between(1, 4, 5 / scale)) * 5
```

frog/imports/ui/GraphEditor/fixedComponents.js
```diff
@@ -58,21 +58,29 @@ export const LevelLines = connect(({
...
+  return (
+    <g>
+      {[...Array(graphDuration).keys()].map(index => {
+        const i = index + 1;
+        const boolToBit = b => b ? 1 : 0;
+        const length = boolToBit(i % 15 === 0) * 15 +
+          boolToBit(i % 5 === 0) * 10 +
+          5;
```
lfaucon: I find that having the displayed text not matching (every `divider` minutes) with the longer ticks (every 15 minutes) is misleading.

```
const length = (i % divider ? 15 : 0) + (i % 5 ? 10 : 0) + 5
```
Also, we can remove this `boolToBit` function

PR #132, opened 2017-04-05
Make selections, renaming, moving work cleaner

houshuang:
The click event we get from the browser is not very clean, if you attach both click and dblClick to the same item, both will fire when you dblClick. You have to do some manual work to wait a bit, and cancel a single click event if a double click is fired. I did that here. 

I also cleaned up a bit, so that it clearly distinguishes between selecting, renaming, and moving. Previously multiple events where fired at the same time (when you move, it would also be selected, when you click to select, it would show the moving bars etc). These should now all be clearly distinguished. There is a tiny lag when clicking, before the sidebar appears, but I think it's manageable.

I also removed stickySelect, and am more careful about using cancelAll, so that you can for example rename an activity, without losing selection. This does add some code and complexity, but all the interactions should now work well, and it should be much easier to work with the sidebar etc.

By mistake, I based this pull-request on the work I had done in hideandshow (PR 131), and by the time I realized, I had already merged in develop, making it too difficult to rebase onto develop. Once you merge 131, the diff view should be much cleaner.


lfaucon:
DblClicking on an activity does not open the rename box for me. Is it the same bug as #136 ?

houshuang:
Everything should work now. The only thing still missing is that if you resize an activity, and the mouse is over the activity when you let go of the button (which is not always the case, because of the rounding), it will still select... The problem is the order in which events arrive - even if I check for state.mode=normal in the keyUp, by that time, the DragEvent has already registered the keyUp, and the mode has already been reset to normal... And if I set some kind of global "hasBeenResizing", when do I cancel that property? Tricky.

PR #139, opened 2017-04-06
Engine fix

lfaucon:
## What works
I fixed and made more robust many activities and operators so that even if the graph is not designed correctly, the app should not crash, but instead for example group all students in one group. I am sure that there are still several bugs left, so share them with me if you encounter any.

## What does not yet work:
- **op-agg-ck-board** & **op-agg-text**: given the new data structure, these two do not make sense anymore. I think we should change them into some *filed-renaming/extracting* operators to help flowing data from one activity to the other when the outputs and inputs are not perfect matches. Do you see what I mean? 
- **ac-ck-board**: I will fix it as it is a very interesting activity for our demo :)


lfaucon:
I modified the ac-jigsaw to be nicer for the demo. I think you can easily make a graph with the editor and run it without problem. Please try and tell me if you encounter any bug :) 

The hard coded jigsaw will not work anymore

There is a problem with some font of ReactJsonSchemaForm that I do not know how to fix

PR #152, opened 2017-04-11
Creates and log in a user with default password from url

lfaucon:
I did not find a way to have many user connected within the same browser.


houshuang:
Great, we should have done this long time ago! :)

PR #159, opened 2017-04-18
Boston scenario

lfaucon:
I think I am getting very close to being able to run the scenario for Boston. The last task will be to enable the list ranking activity to read the input from an operator.

## TODO
- A lot of debugging, trying to build more scenario with the current set of ACs/OPs and fixing all the bugs we encounter. I'll take care of it.

- Automating the demo. You've already been working on it, could you try to see if it will be difficult?

## BONUS
- Improve TeacherView


houshuang:
Why don't we keep work on this branch for now - I don't have time to review in detail and would rather focus on the demo, this frees us up to write exactly how we need to, for the demo, and then after the conference we can review what we actually want to merge into the main develop branch.

lfaucon:
Yes, sure !

houshuang:
Superseded by #198

PR #160, opened 2017-04-19
Boston scenario nightmare

lfaucon:
Can you fix the bug in there? I receive the following error:

```
> node nightmare/script.js 
Error: Cannot read property 'focus' of null
Error: Cannot read property 'focus' of null
undefined
```


lfaucon:
@houshuang The problem was due to the teacher changing the activity too fast, thus nightmare crashed when students try to fill up the form that already disappeared

PR #167, opened 2017-04-25
Nightmare test

lfaucon:
Add a test using nightmare, that simply visits all the tabs by changing urls.


lfaucon:
I realise that to run the nightmare tests on the CI server, we need to run meteor on it. Do you think that we should simply remove the `SHUTDOWN` command that we are currently using? But in that case, should we stop meteor after the tests ran?

houshuang:
Good point... the most important thing is that the timing works right (that the tests are not run before Meteor is ready), and that they complete in time (that they don't wait forever for Meteor to shut down)... If Meteor is still running by the time the container shuts down, that's fine, since it always starts from a clean container... You'll just have to experiment a bit. But if this works well, we could still accept the pull request as something that can be used locally, and you could do another pull request where you integrate it with CI (which should be the goal).

PR #172, opened 2017-05-30
Compound activities

lfaucon:


frog/package.json
```diff
@@ -20,6 +20,7 @@
     "frog-utils": "*",
     "json-stable-stringify": "^1.0.1",
     "json-stringify-pretty-compact": "^1.0.3",
+    "less": "^2.7.2",
```
lfaucon: Forgot it there after experimenting different thing, will remove it.

houshuang:
I'll let you merge it when you're ready. In the future, maybe better to make two different pull requests though - hard to review if so much extra code is added in the middle :)

lfaucon:
I realized too late that the update of the activity packages could have been done separately.

PR #187, opened 2017-06-05
Add vertical lines to display progress

lfaucon:


houshuang:
Specific nits:
- the timer currently starts running the moment I load the graph, before I ever click Start, which seems unintuitive
- did we remove the opportunity to pause/restart graphs?
- I think we need to rethink the design a little bit, the timer is really the least important thing (it's a nice visual reminder, but nothing more), so it should be far less salient. Perhaps a grey indicator only covering the lower 1/3 of the graph, or something like that. And maybe even a checkbox to turn it on/off)
- my instinct would be for the actual execution line (or what we call it, orchestration indicator? :)), to be located between t0 and t1, and not at t0, where it is located now. (Where t0 is the current time point, and t1 is the next - man we need better vocabulary for these things).

Anyway I leave it up to you to judge when we should integrate this pull request - what should be done in this one, and what can be done in a future PR. If you give it a second pass and clean up the timer starting etc, we can probably integrate it - it's new functionality that only adds, but doesn't disturb anything existing, and we can always keep making it better. Also since you're moving stuff in the graph editor, it can create merge conflicts in the future.

houshuang:
Great job.

PR #207, opened 2017-06-09
175

lfaucon:
#175 

The input of panDelta was multiplied by 2 without any reason. Removing it fixed the speed of scroll by dragging the panBox, but made scroll by bringing ac/op/con on an edge very slow, so I also sped that up.


houshuang:
Please better PR titles :)

PR #212, opened 2017-06-13
Adding function and button to restart session

houshuang:
Useful for testing


frog/imports/api/sessions.js
```diff
@@ -129,16 +132,29 @@ Meteor.methods({
...
+  'sessions.restart': session => {
+    const graphId = session.fromGraphId;
+    if (!graphId || !session) {
+      return;
+    }
+    Meteor.call('flush.session', session._id);
+    const newSessionId = Meteor.call('add.session', graphId);
+    Meteor.call('session.joinall', newSessionId);
+    runSession(newSessionId);
+    nextActivity(newSessionId);
```
lfaucon: I understand it is the fastest way to implement this functionality and it works as a temporary solution.

The two concerns I see with this implementation is 
- The list of subscribed students. We could avoid to call joinAll and only join the already subscribed students
- Having modified the session graph before deciding to restart it; With this implementation you lose these modifications

A different strategy for restarting a session could be:
- Delete saved reactive data and operator's products, and update nodes for being not computed
- Reset session's state, timeInGraph, startedAt, openActivities

In general I believe that restarting a session will not be an often used feature (Do you imagine a professor restarting his lecture after 5 minutes? ). Only us as developers feel the need to do so, I believe.

houshuang: I agree, I suggest we leave it in for now, it's useful for development (although it would be great to move towards more automated tests etc, but we'll still need it). In the future, we can replace it with something more robust, or remove it altogether.

PR #223, opened 2017-06-14
Flowstruct

houshuang:
Not ready, but look at flow-utils/types.js


houshuang:
This should be ready now.

PR #224, opened 2017-06-15
This adds the hypothesis operator

houshuang:
I add and update the hypothesis operator to adhere to the new types, and add Jest checking, which mocks the http fetch to a pre-set state, and compares it to a saved snapshot. I also add some functions to frog-utils to work with activityData, for the common case of an operator returning a single unit, not caring about social structures


houshuang:
I added a test for the case that a search returns no results, and also adding ID to all objects, to make it easier to work with. Still OK to merge?

PR #225, opened 2017-06-16
Figuring out how to test with Meteor

houshuang:
I'm trying to write some tests for a function in a file that imports 'meteor/meteor' and 'meteor/mongo', but Jest cannot find these files. We've already added moduleNameMapping to .flowconfig, which works, so I wanted to do the same (also based on Googling), but it still isn't working. If you want to have a look at it, feel free. We need to fix it, otherwise we won't be able to test any files relying on Meteor/Mongo.


lfaucon:
With my last commit, jest finds the file, but there is still an error with it:

```
 FAIL  frog/imports/api/__tests__/activities.js
  ● Test suite failed to run

    ReferenceError: Package is not defined
      
      at frog/.meteor/local/build/programs/server/packages/meteor.js:4:11
      at Object.<anonymous> (frog/.meteor/local/build/programs/server/packages/meteor.js:1380:3)
      at Object.<anonymous> (frog/imports/api/activities.js:1:1011)
      at Object.<anonymous> (frog/imports/api/__tests__/activities.js:3:19)
```

lfaucon:
Done by #461

PR #231, opened 2017-06-20
Fix flow in api/activities

lfaucon:


frog/imports/api/activities.js
```diff
@@ -18,13 +24,18 @@ export const Results = new Mongo.Collection('results');
 
 export const getInstances = (
   activityId: string
-): [?(string[]), structureDefT] => {
+): { groups: string[], structure: structureDefT } => {
```
houshuang: I agree, this is clearer. It was fine while I was using const [groups, structure] = getInstances, but when I saw you using getInstances[1], I realized that it was very hard to understand.

frog/imports/api/activities.js
```diff
@@ -39,50 +50,50 @@ export const addActivity = (activityType, data = {}, id, groupingKey) => {
   }
 };
 
-export const duplicateActivity = activity =>
+export const duplicateActivity = (activity: ActivityDbT) =>
   Activities.insert({
     ...activity,
     _id: uuid(),
     data: { ...activity.data, name: activity.data.name + ' (copy)' }
```
houshuang: I thought we had already separated it out, but I think we should move name to the "top-level", just like grouping key is (I think?)... data should be an opaque config that is only understandable by the activityType.

frog/imports/api/activities.js
```diff
@@ -39,50 +50,50 @@ export const addActivity = (activityType, data = {}, id, groupingKey) => {
...
   });
 
-export const addGraphActivity = params =>
+export const addGraphActivity = (params: Object) =>
```
houshuang: What should params be here? Is it the data: {} object, or other things? If we know what it contains, we could be more specific than Object.

frog/imports/api/activities.js
```diff
@@ -39,50 +50,50 @@ export const addActivity = (activityType, data = {}, id, groupingKey) => {
...
   Activities.insert({
     ...params,
     graphId: params.graphId,
```
houshuang: Isn't this already taken care of by ...params? :)

frog/imports/api/activities.js
```diff
@@ -39,50 +50,50 @@ export const addActivity = (activityType, data = {}, id, groupingKey) => {
...
     createdAt: new Date(),
     _id: uuid()
```
houshuang: This is just a note for future study, but I read some article saying that storing uuids as strings in Mongo was extremely inefficient, and that it would be much more efficient to store it as a binary field or something - something to explore in the future.

frog/imports/api/activities.js
```diff
@@ -39,50 +50,50 @@ export const addActivity = (activityType, data = {}, id, groupingKey) => {
...
   });
 
-export const importActivity = params =>
+export const importActivity = (params: Object) =>
   Activities.insert({ ...params, createdAt: new Date(), _id: params._id });
```
houshuang: Same comment about _id: params._id. WHat does this add over ...params?

frog/imports/api/activities.js
```diff
@@ -105,8 +116,8 @@ export const copyActivityIntoGraphActivity = (
 };
 
 export const copyOperatorIntoGraphOperator = (
-  graphOperatorId,
-  fromOperatorId
+  graphOperatorId: string,
+  fromOperatorId: string
```
houshuang: I am eagerly waiting for the day Flow allows nominal typing, so that we can say that this field takes an operatorId, and Flow will complain if you try to put a GraphId there... :)

frog/imports/api/activities.js
```diff
@@ -118,15 +129,20 @@ export const copyOperatorIntoGraphOperator = (
...
 };
 
-export const removeGraph = graphId => Meteor.call('graph.flush.all', graphId);
+export const removeGraph = (graphId: string) =>
+  Meteor.call('graph.flush.all', graphId);
 
 export const deleteDatabase = () => Meteor.call('graph.flush.db');
 
-export const dragGraphActivity = (id, xPosition) => {
+export const dragGraphActivity = (id: string, xPosition: number) => {
```
houshuang: This function is not used anywhere anymore, should be removed (I guess it's from our old implementation of graph editor)

frog/imports/api/activities.js
```diff
@@ -118,15 +129,20 @@ export const copyOperatorIntoGraphOperator = (
...
   Activities.update(id, { $inc: { xPosition } });
 };
 
-export const addOperator = (operatorType, data = {}, id) => {
+export const addOperator = (
+  operatorType: string,
+  data: Object = {},
+  id: string
+) => {
   if (id) {
```
houshuang: Either you should make id optional in the type, or you should remove the if (id) check :) Seems like we do two different things based on the existence of id, though, so you should make it optional?

frog/imports/api/doGetInstances.js
```diff
@@ -1,18 +1,26 @@
...
-import { getAttributeValues, type ObjectT } from 'frog-utils';
+import {
+  getAttributeValues,
+  type ObjectT,
+  type structureDefT
+} from 'frog-utils';
 
 // given an activityEntry and an objectEntry, calculates which instances
 // to create
-export default (activity: { groupingKey?: string }, object: ObjectT) => {
```
houshuang: Should activity here have the ActivityDbT type? or something more specific?

frog/server/mergeData.js
```diff
@@ -12,7 +12,7 @@ export default (activityId: string, object: ObjectT) => {
   const { activityData } = object;
   const activity = Activities.findOne(activityId);
 
-  const [groups, activityStructure] = getInstances(activityId);
+  const { groups, structure } = getInstances(activityId);
```
houshuang: Very nice that we can change the calling code so little, and get so much more clarity! :)

frog/imports/api/activities.js
lfaucon: I agree, I just fixed the type annotation without looking deeply at all the functions. Do we even use this one?

frog/imports/api/activities.js
lfaucon: I believe it is, I just fixed the type annotation without looking deeply at all the functions.

The slight difference could be that we still write a field `graphId: undefined`. Not sure how mongo behaves with this

frog/imports/api/activities.js
lfaucon: Sure ! I love deleting code <3

frog/server/mergeData.js
lfaucon: :)

lfaucon:
The API activity is still very not perfect and we have an issue to take care of it that has been pending for some time now. I think it will be much easier to do anything as soon as flow will check green, so I suggest to merge this as it is a good step towards green state

PR #233, opened 2017-06-21
Fix Flow in ReactiveHOC

lfaucon:


frog/imports/ui/StudentView/ReactiveHOC.jsx
```diff
@@ -1,15 +1,34 @@
...
+    return WrappedComponent.name;
+  } else {
+    return 'Component';
+  }
+};
 
-const ReactiveHOC = (dataStructure, docId) => WrappedComponent => {
+const ReactiveHOC = (dataStructure: any, docId: string) => (
+  WrappedComponent: Class<Component<*, *, *>>
+) => {
```
houshuang: You are not specifying the return value? (Also Class<Component>)? Or am I reading it wrong?

lfaucon: I had a weird flow error: 
```
Expected React component instead of React$Component
```

I found a fix here, but I will need to read it more carefully to understand it
https://github.com/facebook/flow/issues/4017
https://flow.org/en/docs/types/utilities/#toc-class

PR #238, opened 2017-06-21
Upgrading Prettier formatting

houshuang:
This is the result of running the latest Prettier. I don't love the formatting of decorators, but that's what they've decided, and we want to stay with Prettier - it's not the end of the world :)


lfaucon:
I see no update of prettier in package.json. Is it normal that so many files changed ?

houshuang:
It was in the previous PR to upgrade Flow :) I thought it would not make much difference since it was just a point release but I guess it still did.

PR #254, opened 2017-06-22
Fixing CK-Board, and giving config to mergeFunction

houshuang:
This PR updates CK-board to work properly. It makes sure that the mergeFunction receives both the config data and the activity data, and merges the two in the mergeFunction. I add ways of testing the mergeFunction, but having an easy way to create an in-memory reactiveDataFn / doc. 

I also added the option of having the four quadrants, with titles configurable. (I don't think there's an easy way of hiding the four titles, if people click "no", so that's a bit confusing).

I've tested this manually data straight from op-hypothesis, op-hypothesis + config data, or just config data.

I only tested in the all configuration, I didn't test manually if it also works in p1/p2, but it should.

Because I needed access to the generateReactiveData from an operator, I moved it to frog-utils.

### DIFF
--- a/frog-utils/src/__tests__/generateReactiveFn.js
+++ b/frog-utils/src/__tests__/generateReactiveFn.js
@@ -1,6 +1,6 @@
 import ShareDB from 'sharedb';
 import { uuid } from 'frog-utils';
-import generateReactiveFn from '../generateReactiveFn';
+import { generateReactiveFn } from '../generateReactiveFn';
 
 const share = new ShareDB();
 const connection = share.connect();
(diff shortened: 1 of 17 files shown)


frog-utils/src/__tests__/generateReactiveFn.js
```diff
@@ -1,6 +1,6 @@
 import ShareDB from 'sharedb';
 import { uuid } from 'frog-utils';
```
lfaucon: shouldn't this have become `import { uuid } from '..';` ?

houshuang: Indeed. Tests still seem to pass though, but I've fixed it.

PR #257, opened 2017-06-22
Adding status colors

houshuang:
This PR adds code that I had in the hypothesis branch for showing the status of operators etc during runtime - it reads the state tag, to see if an operator/activity has been computed, etc, and displays colors based on this. This code could also be the basis of highlighting which activities can be dragged to, with type-checking etc.

However, there is a weird bug - when I drag a connection onto an operator, it's flickering on and off (it's triggering onLeave a bunch). I don't know why I would have induced this (I don't think this was a problem in my hypothesis branch). I remember we had such a problem previously, did we figure out a way to solve it? Help is welcome.


lfaucon:
The problem is due to the mouse pointer being over the connection line that you are dragging and not over the operator. When you move, you go over the operator for a second, then the connection line catches up with you and the pointer enters it and leaves the operator. 

I don't know how to fix it. The current fix is to put an offset on the end of the connection, which makes dragging connections forward working well, but not dragging backward.

houshuang:
Will try to fix this before proceeding on this PR

houshuang:
Seems like this is not making the problem worse than the default, so I suggest we merge, and look into how to fix the general problem (will file bug). Nice side effect - sessions show up as green also in graph editor, which is visual indication of what has already been executed / that it's a session, etc.

PR #262, opened 2017-06-26
Re-adding dashboard and dashboard to ac-chat

houshuang:
Note, there are currently some dashboards (like ac-chat) which expect each log message to be a separate db entry, and others (ac-video) which expect there to be a single "merged" object. We need to fix this. But at least displaying the dashboard currently works.


lfaucon:
Did you test it? It's not working for me. The dashboard stays on the state `<p>0 logs</p>` even after students have sent messages. I believe the `logger` passed as props might be not working

houshuang:
See my message - the current log format is messed up, and not in sync with dashboards etc. But this PR does correctly display the dashboard. I guess I can fix the chat dashboard before you approve the PR. If you insist :)

PR #265, opened 2017-06-26
Fixing activity moving/clicking

### LINKED ISSUE #192: [graph-editor] Cannot select activity on left edge
Cannot select activity or operator on the left edge of the graph, is graph is scrolled all the way to the left (starting at 0 minutes). Probably because of the drag-zone, possibly turn off drag zone if it's impossible to drag further.

houshuang:
This more or less fixes #192. I made the drag fields much smaller (you basically have to pull the cursor all the way to the edge of the screen to engage), and also made them hidden if pan is positioned exactly at the beginning or end of graph (most common case is at position 0, adding or selecting the first activity). 

A small other issue was that sometimes if you clicked on an activity and slightly moved the cursor, it would be too little to actually move the activity a single minute, but it would internally register as a "move", and not result in a selection. I now detect this, and trigger a select if an activity has a move activity that doesn't result in actual movement.

(A possibility would of course be to have every move event result in a select, but I'm not sure that's desirable).

Please liberally test - this should work perfectly in all cases.


lfaucon:
I wonder why the scrollfields were here when not desired. I see in the code:

https://github.com/chili-epfl/FROG/blob/develop/frog/imports/ui/GraphEditor/Graph.js#L93 
```
      {scaled &&
        scrollEnabled &&
        <ScrollFields width={graphWidth} height={600} />}
```

https://github.com/chili-epfl/FROG/blob/develop/frog/imports/ui/GraphEditor/store/uiStore.js#L175
```
@computed
  get scrollEnabled(): boolean {
    return !!['movingOperator', 'dragging', 'moving', 'resizing'].includes(
      store.state.mode
    );
  }
```

which seems to prevent the scrollfields to be here when we are doing nothing but selecting an activity. Except if we forget somehow to come back to normal mode. Which should be fixed.

I haven't tested the thiner scrollfields yet

lfaucon:
@houshuang I made the scrollfield visible but transparent. I think it's somehow better for the UI as teachers will have a visual indication that they can scroll by dragging (but not sure it's the best indication... still better than nothing?)

I could not find when the scrollfield are here while not expected to. I think the fact that they are now visible will help us track that issue. 

At the same time I made them bigger again as I find the thin version hard to use.

If you are fine with the current state let's merge this PR

PR #272, opened 2017-06-26
Fixing Runner

houshuang:
This is quite an egregious bug, that harks back to your switching from an array to an Object for doGetInstances. Of course, it would have been better for JS to just complain that you're doing [1] on an object, but instead we just silently get undefined, and the error ends up much later down the stack. I'm actually quite surprised that Flow didn't catch this, the files are all flow-checked.

### DIFF
--- a/frog/imports/ui/StudentView/Runner.jsx
+++ b/frog/imports/ui/StudentView/Runner.jsx
@@ -51,7 +51,7 @@ const Runner = ({ activity, object }) => {
     }
 
     const config = activity.data;
-    const activityStructure = getInstances(activity._id)[1];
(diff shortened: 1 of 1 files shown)


houshuang:
I tried a few different ways to see what Flow picks up. Apparently [1] is a blindspot... 

screen shot 2017-06-26 at 22 47 51

It recognizes that { structure } is correct, and that { bullshit } is incorrect, whether I use the destructuring, or the keyed access. But it doesn't complain about [1]. Maybe a bug.

lfaucon:
I'm surprised too that flow doesn't catch this

PR #275, opened 2017-06-27
Removing getInstances, and switching to doGetInstances directly

houshuang:
getInstances wasn't doing any useful job anymore

### DIFF
--- a/frog/imports/ui/StudentView/Runner.jsx
+++ b/frog/imports/ui/StudentView/Runner.jsx
@@ -5,11 +5,13 @@ import { Meteor } from 'meteor/meteor';
 import { createContainer } from 'meteor/react-meteor-data';
 import { MosaicWindow } from 'react-mosaic-component';
 import { focusStudent, getMergedExtractedUnit } from 'frog-utils';
+import { cloneDeep } from 'lodash';
 
 import { activityTypesObj } from '../../activityTypes';
 import { createLogger } from '../../api/logs';
 import { Objects } from '../../api/objects';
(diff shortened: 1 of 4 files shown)


frog/imports/ui/StudentView/Runner.jsx
```diff
@@ -40,9 +42,10 @@ const Runner = ({ activity, object }) => {
 
     const RunComp = activityType.ActivityRunner;
     RunComp.displayName = activity.activityType;
-    const ActivityToRun = ReactiveHOC(activityType.dataStructure, reactiveId)(
-      RunComp
-    );
+    const ActivityToRun = ReactiveHOC(
+      cloneDeep(activityType.dataStructure),
```
lfaucon: Do we still have problems with that? I thought it was fixed already

houshuang: Sorry, was trying to separate out into different PRs, but this one snook in. I'll leave it in for now - this actually got removed, and the problem reappeared. Anyway, I think this is a more solid solution than either {...dataStructure} (which doesn't work for arrays), or Object.freeze, etc.

frog/server/reactiveToProduct.js
```diff
@@ -33,7 +37,7 @@ export default (activityId: string) =>
     activityId,
     {
       $set: {
-        product: getActivityDataFromReactive(activityId),
+        activityData: getActivityDataFromReactive(activityId),
```
houshuang: This was another bug

PR #277, opened 2017-06-27
Fixes activity reactive data being undefined

houshuang:
The activity was loading before the reactive data had been loaded. This was supposed to be fixed by initializing it with dataStructure, but anyway I think it's better to wait until the reactiveData has been loaded, so activities don't begin using dataFn etc before they are ready?


frog/imports/ui/StudentView/ReactiveHOC.jsx
```diff
@@ -27,7 +27,7 @@ const ReactiveHOC = (dataStructure: any, docId: string) => (
     constructor(props: Object) {
       super(props);
       this.state = {
-        data: dataStructure,
+        data: null,
```
lfaucon: I don't understand how this solves the problem of being undefined

frog/imports/ui/StudentView/ReactiveHOC.jsx
```diff
@@ -67,11 +67,15 @@ const ReactiveHOC = (dataStructure: any, docId: string) => (
...
-      />;
+      <div>
+        {this.state.data
+          ? <WrappedComponent
+              dataFn={this.state.dataFn}
+              data={this.state.data}
+              {...this.props}
+            />
+          : 'Loading...'}
+      </div>;
```
lfaucon: This doesn't need to be wrapped in a div

PR #278, opened 2017-06-27
Operators only run once, fixing #276

houshuang:


frog/server/runDataflow.js
```diff
@@ -54,6 +53,8 @@ const runDataflow = (
     return;
   }
 
+  nodeTypes[type].update(nodeId, { $set: { state: 'computing' } });
```
lfaucon: It was very hard for me to understand why moving this line after the if statement helps. You never use the fact that state will be equal to 'computing'. As of now we could remove this line and everything would work the same. 

I thought the problem was due to concurrency issue of code running at the same time, but actually it was only that the state='computed' was being overwritten. 

Will we ever have issue or concurrency? 
If yes, I would suggest replacing the above if statement by `if (node.state === 'computed' || node.state ==='computing')...`

houshuang: No, we should never have issue of concurrency as far as I understand. 

> In Meteor, your server code runs in a single thread per request, not in the asynchronous callback style typical of Node. We find the linear execution model a better fit for the typical server code in a Meteor application.

(https://gist.github.com/possibilities/3443021)

PR #280, opened 2017-06-27
Makes session names much more userfriendly

houshuang:
This PR generates nice names for session graphs, with the original graph name, prefixed by #, and optionally adding a number if there are multiple sessions generated by the same graph. It also adds this name to the session object, and uses this in the TeacherView display.

In general, we should avoid the user seeing any uuids, unless it's explicitly for a debugging purpose.


frog/imports/api/sessions.js
```diff
@@ -85,11 +85,18 @@ Meteor.methods({
   },
   'add.session': graphId => {
     const sessionId = uuid();
-    const copyGraphId = addGraph('SESSION_GRAPH<' + sessionId + '>');
+    const graph = Graphs.findOne(graphId);
+    const graphs = Graphs.find({}).fetch();
```
lfaucon: Not efficient to download all the graphs object from the database to then check all of their names. My next commit suggest a nicer way to do the same thing

lfaucon:
@houshuang As I made the last changes, I'll let you merge it

PR #291, opened 2017-06-29
Script to automatically create and initiate activity packages

houshuang:


houshuang:
Sure, it's an easy change. Of course, then people might write
createActivity ac-short, and get ac-ac-short :) But you could test for that
:)

I don't have time, feel free to update, and also create one for operator.
(It might be better to change this to a generic "create" function, which
takes operator / activity as an argument, because a lot of the commands are
identical, and some of the stuff could be more modularized as well.

houshuang:
Thinking further about this, I think I want people to type in the whole id, because that's the id that is used everywhere, in the import statements, in ac/... etc, so it's more explicit. However, if you want to have the script check that the shortname begins by ac-, and error if not, that would be an easy fix.

I also see there are some small ESLint errors. Feel free to cleanup and pull, I want touch this again.

aperritano:
thanks guys, this worked great.

PR #309, opened 2017-07-06
Fix path to data in ac-brainstorm

lfaucon:
Fixing a few bugs for a small demo of FROG this afternoon

- The path to the reactive data of ac-brainstorm sometime started by 'ideas' and sometimes not. I fixed it to never use 'ideas' in the path.

- According to the types: 
```
obj.data: Object || any[]
```
I disallowed it to be undefined some time ago, but for this I made the server send {} instead of undefined. So we still have to check if data is of type Array or Object in the ac/op packages when we run specific function like forEach for example. I am surprised that flow does not complain about accessing `.forEach(...)` on possibly an object.


houshuang:
Agree with the path change. The nice thing about having generic items in a top level list (if we standardized on only object, we could also standardized on always having {list: [...]} of course... The advantage is that you can have generic operators that operate on items (distribute, aggregate, etc).

lfaucon:
@houshuang  

I made the test using `Array.isArray` which is indeed much better (the previous test on forEach would have failed on an object which has a key 'forEach')

I understand your suggestion about not having to always check the incoming data in operators. This PR fixes a bug in `ac-brainstorm.mergeFunction` when it has no incoming connection. As the incoming data `type dataUnitT = Object | any[];` is wrongly initialized to `{}`

I would suggest:
- Restrict ourselves to `type dataUnitT = Object;` and doing `{ list: [...] }` or `{ 0: ..., 1: ..., 2: ...}` when we want lists. Preferably using the same approach in most op/ac that handle lists. We are already doing 
`{ 0: ..., 1: ..., 2: ...}` in ac-brainstorm for example.
- Require ac packages to set the default empty data to `{}` or `[]`

I prefer the first option. What do you think?

houshuang:
So in this case, if it had correctly specified dataStructure, it would have worked? in that case, it's kind of the activity's responsibility to make sure it correctly specifies. Whether we should always default to objects, I'm more agnostic on. I'm open to it, but I'm not sure it has a huge advantage - if the activity tried to iterate over obj.list, it would still fail, and it would still have to check whether obj has list (which is an array), unless you had dataStructure: { list: []}, in which case you might as well just say dataStructure: []?

lfaucon:
I believe you are mixing up things.

There is the internal reactive data, written by the `dataFn` and initialised by `dataStructure`. 
There is also the data received from product operators, for which activities do not define empty structures, which can be `Object | Array` and which is `{}`if empty.

ac-brainstorm maintains internal data which looks like `{id0: ..., id1: ..., id2: ...}`, but expects an array `[...]` as input.

houshuang:
So actually looking at the code, rather than just scanning the PR and responding, the code in question is in mergeFunction. mergeFunction is called from server/mergeData, which uses `getMergedExtractedUnit` to get the activityData, and then calls the merge function. However, it checks if it has activityData, before calling the mergeFunction. in `extractUnit`, it says:

```
  if (!data) {
    return { data: {}, config: {} };
```

This is obviously problematic, since we cannot assume that {} is valid input data. I wonder if this came about when we began including config in the data structure, before that we could just have an empty object, now we need this whole structure... One choice would be to not actually call the mergeFunction if data: {}, however I guess there might be cases in which you want to merge from config (if someone specifies boxes in the config for example)... Another choice would be to have a better marker of empty data, like data: null. (Which would give immediate errors if someone tries to access a key, but that might be safer than not giving error, but returning undefined). 

In the end, an activity should not receive data in which it cannot deal with, or if it has declared that it can deal with the data, then it should be able to deal with it (through checking etc). Including empty data. 

Anyway I'll let you figure out and resolve this for now, I've said enough :) We can revisit whatever you do in the future when we have strong type-checking etc.

PR #310, opened 2017-07-06
File upload

agentromain:
Upload file feature added to the sidePanel of the Graph Editor
It gives you the data URL of an uploaded file


frog/imports/ui/GraphEditor/SidePanel/UploadFile.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+class UploadFile extends Component {
+  state: { url: string };
+
+  constructor(props: {}) {
+    super(props);
+    this.state = { url: '' };
+  }
+
+  processFile(files: Array<any>) {
```
lfaucon: I don't see this function used anywhere. Are you sure it is necessary?

frog/imports/ui/GraphEditor/SidePanel/UploadFile.jsx
```diff
@@ -0,0 +1,55 @@
+// @flow
+
+import React, { Component } from 'react';
+import Form from 'react-jsonschema-form';
+
+class UploadFile extends Component {
```
lfaucon: This class does not upload files. It converts files to dataURI. The file are uploaded only when we copy this dataURI in the configuration.

frog/imports/ui/GraphEditor/SidePanel/ActivityPanel.js
```diff
@@ -75,6 +76,7 @@ const EditClass = props => {
       >
         <div />
       </Form>
+      <FileForm />
```
lfaucon: Right now it hard to make the difference between what part of the sidePanel is for the config and what part is the fileUploaderTool. Can you:
- Stick the tool to the bottom of the SidePanel
- Have a toggle button for the tool

ac/ac-induction/src/config.js
```diff
@@ -22,11 +22,11 @@ export default {
       }
     },
     imgTrue: {
-      title: 'correct image uri',
+      title: 'Correct image URL',
```
lfaucon: Right now when I paste a large dataURI in a field of the configuration form. It calls `onChange` and takes time to load. During that time I lose control of the app. Uploading files usually is done in the background to not freeze the interface for the user. You need to handle that issue.

For this you should see how to make the Configuration `onChange` asynchronous.

frog/imports/ui/GraphEditor/SidePanel/UploadFile.jsx
```diff
@@ -0,0 +1,55 @@
...
+              'ui:widget': 'file'
+            }
+          }}
+          onChange={this.onChange}
+          liveValidate
+        >
+          <div />
+        </Form>
+        <h4>Submitted file as data URL</h4>
+        <pre>{this.state.url.toString()}</pre>
```
lfaucon: It's not always easy to select and copy this very long url. Can you make a "copy to clipboard" button?

lfaucon:
We can really do cool things with these dataURI.
ac-iframe has become ac-video, ac-image, ac-pdf :)

houshuang:
Since I haven't been part of the discussions, I'm not sure exactly how this is supposed to work, so I won't comment for now.

frog/package.json
```diff
@@ -53,8 +55,9 @@
...
-    "ac-induction": "*"
+    "ws": "^3.0.0"
   },
-  "devDependencies": {"flow-bin": "^0.48.0"}
-}
\ No newline at end of file
+  "devDependencies": {
+    "flow-bin": "^0.48.0"
+  }
+}
```
lfaucon: Thank you for sorting `frog/package.json` in alphabetical order. Did it happen when you ran `meteor add ...` ?

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
+// @flow
+
+import React, { Component } from 'react';
+import Form from 'react-jsonschema-form';
+import CopyToClipboard from 'react-copy-to-clipboard';
+
+import { Uploads } from '/imports/api/uploads';
```
lfaucon: I'm surprised that the path `/imports/api/uploads` works! I was expecting `../../../api/uploads`

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
+// @flow
```
lfaucon: This file is still called `fileConvert`, it makes more sense now to call it back `FileUploader` now :)

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+import React, { Component } from 'react';
+import Form from 'react-jsonschema-form';
+import CopyToClipboard from 'react-copy-to-clipboard';
+
+import { Uploads } from '/imports/api/uploads';
+
+class fileConvert extends Component {
```
lfaucon: rename `FileUploader`

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+    const that = this;
+    Uploads.insert(e.formData.file, (err, fileObj) => {
+      that.setState({
+        url: window.location.origin + '/cfs/files/uploads/' + fileObj._id,
+        copied: false
+      });
+    });
+  };
+
+  onCopy = () => this.setState({ copied: true });
```
lfaucon: I believe you use this method nowhere

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+                }
+              }}
+              uiSchema={{
+                file: {
+                  'ui:widget': 'file'
+                }
+              }}
+              onChange={this.onChange}
+              liveValidate
+            >
```
lfaucon: You could have before the class `fileConvert` a component

```
const FileForm = ({ onChange }) => 
           <Form
￼               schema={{
￼                 type: 'object',
￼                properties: {
￼                 file: {
￼                     type: 'string',
￼                     title: 'Upload an image to get its URI'
￼                   }
￼                 }
￼               }}
￼               uiSchema={{
￼                 file: {
￼                'ui:widget': 'file'
￼                 }
￼               }}
￼               onChange={onChange}
￼               liveValidate
￼             >
```

and then only write in the fileConvert Component:

```
<FileForm onChange={this.onChange} /> 
```

Advantages of this are for example:
- easier to read and understand the code
- easier to write tests for each component separately

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  constructor(props: {}) {
+    super(props);
+    this.state = { url: '', copied: false, display: false };
+  }
+
+  onChange = (e: { formData: { file: string } }) => {
+    const that = this;
+    Uploads.insert(e.formData.file, (err, fileObj) => {
+      that.setState({
```
lfaucon: If I switch tab or close the sidepanel too fast, I receive the following error
```
warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the fileConvert component.
```

To fix that you should use the methods `componentDidMount` and `componentWillUnmount` and for example maintain a variable `this.mounted` and test `if(mounted) { setState(...) }`

https://facebook.github.io/react/docs/react-component.html

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+class fileConvert extends Component {
+  state: { url: string, copied: boolean, display: boolean };
+
+  constructor(props: {}) {
+    super(props);
+    this.state = { url: '', copied: false, display: false };
+  }
+
+  onChange = (e: { formData: { file: string } }) => {
```
lfaucon: If I upload a big file, the UI still gets stuck until the file finishes to upload. I expected CollectionFS to work better for this, but I think you'd do better to use `Promise` as we had seen before.

Also, I'd love if when the file is chosen the url takes the value `Uploading. . .` and then the correct value when the upload is over. For this you can just `setState('Uploading. . .')` at the beginning of `onChange`

frog/imports/ui/GraphEditor/SidePanel/fileConvert.jsx
```diff
@@ -0,0 +1,85 @@
...
+import CopyToClipboard from 'react-copy-to-clipboard';
+
+import { Uploads } from '/imports/api/uploads';
+
+class fileConvert extends Component {
+  state: { url: string, copied: boolean, display: boolean };
+
+  constructor(props: {}) {
+    super(props);
+    this.state = { url: '', copied: false, display: false };
```
lfaucon: Do you need the field `copied` in the state? I don't see how the component changes depending on that property.

frog/package.json
agentromain: I guess

PR #311, opened 2017-07-06
Change joinAll method to be more efficient and exclude teacher

houshuang:


houshuang:
This changes the joinAll method in the session API to be more efficient (using Mongo more), but it also excludes teacher from being included. 

Note that we should also make sure that the teacher is included in p3 activities, maybe by adding anyone with that controlSession as well as currentSession (future PR).

PR #318, opened 2017-07-13
Argue op

agentromain:
the operator has been tested with 4, 5 and 6 students.
It create groups with students who have the most different results from an activity
If there is an odd number of students, we consider only the n-1 first students and at the end we assign the last one with the group that corresponds to him the best.


op/op-argue/src/index.js
```diff
@@ -13,56 +13,134 @@ const config = {
...
+};
+
+const dist = (A, B) => {
+  const s = new Set();
+  const k = Object.keys(Object.assign({}, A, B));
+  for (let i = 0; i < k.length; i = 1 + i) {
+    if (A[k[i]] !== B[k[i]]) {
+      s.add(k[i]);
+    }
+  }
```
lfaucon: Do you like functional programming?

```
const s = new Set();
Object.keys(Object.assign({}, A, B))
    .filter(k => A[k] !== B[k])
    .forEach(k => s.add(k))
return s;
```

or even better

```
return Object.keys(Object.assign({}, A, B))
    .filter(k => A[k] !== B[k])
    .reduce((acc, k) => acc.add(k), new Set())
```

op/op-create-groups/src/index.js
```diff
@@ -33,6 +33,7 @@ const operator = (configData, object) => {
 
   const ids = shuffle(globalStructure.studentIds);
   const struct = chunk(ids, configData.groupsize);
+
```
lfaucon: It's better if you don't modify files when you don't need it. We could want to see the history of modification on `op/op-create-group` and it makes it easier to have the least number of commits possible.

op/op-argue/src/index.js
```diff
@@ -13,56 +13,134 @@ const config = {
...
+  return result;
+};
+
+const dataToArray = (ids, payload) => {
+  const result = [];
+
+  for (let i = 0; i < ids.length; i += 1) {
+    result.push(flatten(payload[ids[i]].data));
+  }
+  return result;
```
lfaucon: I love functional programming :)

```
return ids.map(id => flatten(payload[id].data));
```

op/op-argue/src/__tests__/index.js
```diff
@@ -0,0 +1,240 @@
...
+//         break;
+//     }
+//   }
+//   return success;
+// };
+
+// test('Test 6, with gen and verifier: 4x10 students, 3 answers', () =>
+//   expect(
+//     verifyAnsFromGen4(operator({}, genGroupFromTabADCB(obj3, 10)))
+//   ).toBeTruthy());
```
lfaucon: Why did you comment this test? Doesn't it work anymore?

agentromain: It was created for when the input data of the operator had the form: 
`{ students: [gradesStud1: [], gradesStud2[], …]}` so the operator wouldn't be able to handle it.

op/op-argue/src/__tests__/index.js
```diff
@@ -0,0 +1,252 @@
...
+
+const expected2 = {
+  group: { '1': ['1', '4'], '2': ['2', '3', '5'] }
+};
+
+const expected3 = {
+  group: { '1': ['1', '4'], '2': ['2', '3'], '3': ['5', '6'] }
+};
+
+const expectedMult = res => {
```
lfaucon: I find this function a bit over complex. What do you think of that ?

```JavaScript
const expectedMult = res => {
    const l = Object.keys(res.group).length / 2;
    const group = x => Math.floor(x / l) // returns 0, 1, 2, or 3
    return Object.values(res.group).reduce(
        (acc, [s1, s2]) => acc && (
            (group(Number(s1)) + group(Number(s2))) % 4 === 1 // returns true is [0 and 1] or [2 and 3] but not other pairs
        ),
        true
    )
}
```

PR #319, opened 2017-07-13
Fix a lot of issue with NaN in svg.

lfaucon:
I believe the errors due to wrong usage of `timeToPxScreen`


houshuang:
This is due to a spurious activity, and we should fix the root cause - this will only hide the real bug.

PR #333, opened 2017-07-18
Valid graph

agentromain:
The validator component works fine, the validGraphFn (in the API) might be improved.
For now it only test if the op/ac have a type, if any json-schema form has any problem and if all group activities are always connected to a social op


houshuang:
Hi,
I would suggest that you add some test cases, as we discussed, before I start reviewing your code style. That way, if we make some stylistic changes to the code, we'll know right away that we're still getting the correct results. Ping me when you want me to review.

houshuang:
A bug, when I simply create a single new activity

screen shot 2017-07-20 at 11 40 17

houshuang:
Can you please:

- [ ] Move the red/green circle onto the top menubar, next to the graph name (probably change it from SVG to a simple unicode circle or something)
- [ ] Merge my pull request
- [ ] Add a check to make sure that all activityTypes and operatorTypes are valid (exist in the system) - this error can happen if you import a graph (for example the argueGraph that you can import from admin)

Then we can merge this. Thanks!

PR #334, opened 2017-07-18
Configs required fields

romainAA:
I added a required field to all activities/operators except for chat and argue because they don't need it, and ck-board and distribute (because I don't fully understand them). A also removed a graph from the teacherView creating session list if this graph was invalid.


agentromain:
shouldn't we close this pull request ? I don't think there are many changes and I think @houshuang already modified the sessions filter in the teacherView

houshuang:
Let's keep it open while I think about how to deal with form validation in a better way :)

PR #342, opened 2017-07-21
New act type list

agentromain:
Changed the interface of the choose ac/op list


frog/imports/ui/GraphEditor/SidePanel/ListComponent.jsx
```diff
@@ -0,0 +1,58 @@
...
+    const InOut = () => (inp !== '' && outp !== '' ? <br /> : <div />);
+    const outp = obj.meta.outputs === '' ? '' : 'Outputs : ' + obj.meta.outputs;
+
+    const onOver = () => this.setState({ over: true });
+    const onOut = () => this.setState({ over: false });
+
+    return (
+      <div value="titi" className="list-group-item">
+        <h5 style={{ fontWeight: 'bold' }}>
+          {' '}{obj.meta.name}{' '}
```
houshuang: Don't need the {'  '} spaces here?

agentromain: I think eslint put them here

houshuang: That would be prettier (which is triggered by eslint). I have prettier run on each save in my editor, so I see stuff like this right away.

agentromain: So you mean I should erase it manually every time I run npm start fix ? Or just leave it here ?

PR #346, opened 2017-07-24
Better ac/op list

houshuang:
This PR

- improves the look of the list a bit, by using bootstrap Button instead of button, etc.
- now, you can only expand items, expanding an item collapses all other items

Still not perfect, but seems cleaner to me.

Note the use of recompose to add state to a stateless component. I just don't like classes :) We use this a few other places, including the withVisibility function from frog-utils :)

Also updated/expanded on the descriptions in all ac/op packages.


agentromain:
Everything seems good to me, the expand button looks a lot like the 'choose' one, but I guess it doesn't matter

PR #347, opened 2017-07-24
New act type list

agentromain:
few modifications


houshuang:
The situation is a tiny bit messy right now. We have this PR:

https://github.com/chili-epfl/FROG/pull/346

which builds on your work, reformats the list, and adds better descriptions

and I just pushed

https://github.com/chili-epfl/FROG/pull/348

this is the first step of implementing preview, I reformatted the list display again, and added a preview button.

Perhaps you can review 346 and we can merge it if you're happy with it. Then we could also merge 348 if you like it (it doesn't have the actual preview functionality yet, but at least the change of look). Then we can see if there are any remaining changes in this PR which need to be merged? Otherwise, it's hard for me to tell what is new code here, since it's being compared to develop, and not to all the added code mentioned above.
Thanks!

PR #357, opened 2017-07-26
Fixing auto-login

### LINKED ISSUE #332: Login called when user is logged in
When switching between views (and URLs) with a logged in user, the login functions gets called, creating console errors. It should only be called if there is no logged in user.

houshuang:
Ensures that login only occurs when we're actually changing user, not just because the URL has changed.

Closes #332

### DIFF
--- a/frog/imports/ui/App/index.js
+++ b/frog/imports/ui/App/index.js
@@ -44,7 +44,8 @@ export default class App extends Component {
 
   handleNewHash = () => {
     const [, username, location] = window.location.hash.split('/');
-    const loggedInUsername = Meteor.users.findOne(Meteor.userId()).username;
+    const loggedInUsername =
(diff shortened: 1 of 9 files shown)


frog/imports/ui/App/index.js
```diff
@@ -34,12 +34,19 @@ export default class App extends Component {
...
+        this.handleNewHash();
+        window.addEventListener('hashchange', this.handleNewHash, false);
+      }
+    });
   }
 
   handleNewHash = () => {
     const [, username, location] = window.location.hash.split('/');
-    if (username) {
+    const loggedInUsername = Meteor.users.findOne(Meteor.userId()).username;
```
agentromain: I don't know if it's relevant since we don't really handle different users and since students can only access students view, but with an URL like this one : http://localhost:3000/student1 the window.location.hash will be "". That means that username and location will be set to undefined

Issue #367, opened 2017-08-01
graph editor error: Cannot read property 'toLowerCase' of undefined


aperritano:
Sync'ed with upstream for the latest code changes. Steps to reproduce: 

1. Add an activity to a level and click on it. Left panel doesn't open until menu icon is clicked; shows without options to choose an activity.

<img width="1336" alt="screen shot 2017-07-31 at 11 09 21 pm" src="https://user-images.githubusercontent.com/219031/28809028-29779290-7646-11e7-8784-82c2d52fabfe.png">

houshuang:
Can you reproduce on a clean develop branch - could this be a merge problem? I cannot seem to reproduce on our repo.

Also, I tried installing the latest version of your fork (initial_setup.sh) etc, but it did not compile, some missing libraries etc. Would be happy to add your repo to our automatic continuous integration testing, which could help make sure you're always "green". Let me know.

aperritano:
I have the found the error and fixed it; frog utils. What is your preferred protocol for pushing back into the remote from a fork?

What is the link to continuous build site? What is the error output?

houshuang:
Just make a PR, but make sure it's on our code, not on your fork. If you made the fix in your own code, you can create a branch off our develop and cherry pick the one commit that fixes the bug onto it, and make a PR based on that. It's hard for me to give more details since I don't know what your local setup is like.

The CI is integrated with github and shows up automatically in all commits, PRs etc, like this

<img width="541" alt="screen shot 2017-08-01 at 15 39 30" src="https://user-images.githubusercontent.com/61575/28828014-b19d2b1a-76cf-11e7-84fa-5931b8b04653.png">

currently it's broken, because the server ran out of disk space, but I just got a new 1TB shared drive, which I'll hook up tomorrow. Here's an example of build output from a successful build: http://icchilisrv3.epfl.ch:8080/job/FROG/job/develop/42/console - you an access this from clicking on the green/red marks in our repo - it's public.

houshuang:
Just looking at your commit (Btw you continue to commit your idea.xml files), this is an example of things that can happen because we don't have your activity as part of our repo, so any of our checks etc will not reveal any problems with your activity.

Also I think the flow definition changed to include description and shortDesc, so it should have given you a flow error telling you that your activitity package had a type error?

aperritano:
[...]

PR #368, opened 2017-08-01
fix for issue #367

aperritano:


houshuang:
I will close this, because shortDesc is required in the type definition of activity, and we should expect packages to conform to the type definition.

PR #386, opened 2017-08-11
Improve induc

romainAA:
New smart induction activity that also works by group and produce data


ac/ac-induction/src/index.js
```diff
@@ -46,8 +68,9 @@ const meta = {
 export default ({
   id: 'ac-induction',
   type: 'react-component',
-  meta,
   config,
+  meta,
   ActivityRunner,
+  merge,
```
lfaucon: This should be a field call `mergeFunction`. If you don't give it a correct name it won't be used by the engine. did you make sure it works correctly? I get the following error, after a clean install on your branch, then creating an ac-induction and trying to run it. Also, same error when trying to see the preview. It seems to me it could be due to the non initialization of the reactive data.

screenshot from 2017-08-11 15-13-18

lfaucon: I confirm this error. Fixing it seems to get everything working

romainAA: Ok ,I'm correcting it and I push it

ac/ac-induction/src/mergeFun.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import { shuffle } from 'lodash';
+
+export default (obj: Object, dataFn: Object) => {
+  const listIni = [];
+  for (let i = 0; i < obj.config.examples.length; i += 1) listIni.push(i);
```
lfaucon: Don't use a for loops if you can avoid it. In general, you never need to create an empty list and fill it up within a for loop. Especially for the simple case of `[1, 2, 3, 4, 5, ...]` 
```
const listIni = [...obj.config.examples.keys()]
const listIni = [...Array(obj.config.examples.length).keys()]
```

lfaucon:
I don't have enough time to review your code in detail yet. I'll go over it again next week

houshuang:
@lfaucon I'll let you be responsible for reviewing and merging this code

lfaucon:
This PR is not really mergeable as-is, it will require more work from Romain's. We wondered if it would be a good topic for his semester project

romainAA:
This is a first PR that could be merged, even if it isn't fully functional since it does't create any bug (I think) and passes flow and eslint errors. 
This PR deals with the induction activity, it's divided in few parts depending on the configuration given by the teacher (choosable through a boolean).
1. Presentation: The teacher explains to the students what he will be doing in the activity
2. Examples: The student will see images and be told if they correspond or not to the concept.
3. Tests with feedback: The student will be shown images and he will have to say if they correspond to the concept or not and why. Then when he will submit response, a modal pops-up to correct him. (for now only the image and a blank modal is implemented)
4. Definition: The teacher then gives the general definition to the student to avoid any misconceptions.
5. Tests: The student is then tested as in part 3 but won't receive any feedback to evaluate his understanding of the concept. (implemented as much as part 3)
@lfaucon

lfaucon:
@romainAA I finally looked at what you've done! I like it a lot. 
Could you try to fix the conflicts so I can merge it? I believe the files can have changed because of updating the eslint version. Thanks !

PR #392, opened 2017-08-14
React router

romainAA:
The basics works as it use to but with a better routing system.
But a student can access all apps by the url.


houshuang:
You didn't run ESLint and Prettier :)

houshuang:
I did a substantial rewrite of Romain's work attempting to use as much idiomatic router features as possible. This seems to work, but I haven't tested it extensively. Needs review, and more testing. 

The desired functionality:
- if no user logged in, present option to log in as teacher/student (as before)
- no matter what URL is entered, it goes to student view for students
- you can log in as a random user by appending ?login={username} on any URL, it logs in, and redirects to remove this from the URL bar. This means that we can develop our "real" URL schema, and not worry about the fact that in the future, this quick-login feature will be removed
- it should wait until all data is loaded until it shows the various components
- in graph editor, it updates the URL bar with the ID of the current graph - you can have multiple graphs open in different windows, if you copy a URL with a graph id, it will open on that specific graph id etc.

Let's test thoroughly, and also look at code organization. Then we can think about other places where we want to externalize state into the URL bar.

@lfaucon @romainAA

romainAA:
When you want to create a new session in the teacherView, it doesn't gives you the usual panel but just sends you to the graphEditor and you can't come back to the teacherView.

houshuang:
Also, when clicking twice on "Graph", the second time it will not show the hash (because setId is not triggered?) Maybe make Graph disabled if already on Graph?

houshuang:
With latest push, teacher view is broken (setId pushing - not updated with history, and would push the wrong url - graph/)

houshuang:
OK, would like to get this merged, so please review again @lfaucon @romainAA 

Currently, the URL should update with the current graph in both graph editor and teacher view, and loading a URL with a valid graphId should switch to that graphId. The only caveat is that the teacher view still uses controlSession from the database, which means that if you have two windows open, you can only watch a single session at a time. It requires a bigger rewrite to decouple this, not necessary at this point.

Otherwise everything should work properly as described above. Note that you can no longer click on the currently active tab, it's also shown in a different color.

PR #393, opened 2017-08-15
Using Yarn Workspaces for initial_setup

houshuang:
Yarn has a new feature called [workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) designed for mono-repos. You need a .yarnrc file turning the feature on, and to specify the list of workspaces in the root package.json. Now, we only need to run yarn install in the root directory, and it will look at all required files, and install them a single time instead of many, and also figure out links between packages internally etc.

The only thing we finally add is to symlink all of the root packages into frog, because Meteor linker does not look into parent directories like npm/node. 

This should make builds faster and smaller, and maybe even help with production building.


houshuang:
We should wait for [this](https://github.com/yarnpkg/yarn/issues/4070#issuecomment-322430009) to be resolved before merging - otherwise installing new packages in frog/ will break Meteor.

Issue #388, opened 2017-08-14
Image classifier activity


houshuang:
Given a list of images as input (URLs), this activity should rapidly display each image full-screen, and let the user click a button (letter/number) to tag the image with that letter. The output of the activity is the same list of URLs, with the tag received by the user. Which categories/keys could be configured in config.

It should be possible to do this very quickly, if necessary, some form of pre-loading of URLs might be necessary (see if there's an existing component for this).

Issue #387, opened 2017-08-14
Image viewer activity


houshuang:
We need an image viewer activity with the following characteristics:

- takes a list of images as input (URLs)
- optionally images come with one or several attributes

View 1: List of thumbnails
Images in the list are shown as thumbnails. Clicking on a thumbnail shows the image full-screen. Left/right used to navigate between images (can also have some icons). Esc/clicking X goes back to thumbnail view.

View 2: Overview of categories
If the image URLs have attributes (for example group 1 -5), n number of boxes are shown on the screen. Within each box a few random thumbnails from that "category" are shown. Clicking on the box opens that category in View 1. (Clicking X goes back to overview of thumbnails).

If there are multiple categories, there is a pulldown letting the teacher switch between the different categories to view the organization. 

Interaction (can be configured in the config):
Allow user to submit an image - shows a button that lets user upload/drag and drop/capture from laptop camera

Allow users to select images - Users can toggle selected images by clicking on them. A single click changes border color to green. One more click unselects. 

If the activity is p2, config can specify: minimum number of votes. For example 2 votes, first click turns border yellow, second click turns border green. 

Submit button - show a submit button, once submission is done, no more changes can be done. 

Show basket - this is a different mode from the voting, where a user can have a "basket of thumbnails", and add or remove images from this basket. (Could also be starring images, or some other way of selecting certain images).

All interactions must be logged for learning analytics.

The first step would be implementing View 1. Images can be added in config, or from another activity. Instead of a list of URLs, it could be a list of objects with fullsize: and thumbnail:, if we have resized versions of all thumbnails, might be important for performance. Also look into ways of dealing with many thumbnails - scrolling, on-demand loading, etc.

houshuang:
@romainAA Did you ever implement the "star" and view category? I think the easiest way, as you mentioned, was to have a simple way to star, and then just treat that as another category that can be viewed. Maybe voting behaviour can be used here.

PR #396, opened 2017-08-15
Img class

romainAA:
The activity works, tell me if you want the UI to be modified.
There is a flow error that come from a node_module file : 
61: export const typeFromAST: $FlowFixMe = typeFromASTImpl;
                               ^^^^^^^^^^ identifier `$FlowFixMe`. Could not resolve name


lfaucon:
@romainAA If this relates to an existing issue, can you provide the link to the issue in a comment?

is it this one #388 or this one #387 ?

romainAA:
Related to #388

ac/ac-imgClass/src/imgClassifier.jsx
```diff
@@ -0,0 +1,67 @@
...
+  width: 100%;
+  height: 90%;
+`;
+
+const ImgPanel = (props: { url: string }) =>
+  <div style={{ width: '90%', height: '100%' }}>
+    <ImgBis alt="" src={props.url} />{' '}
+  </div>;
+
+const ShortCutPanel = (props: { categories: Array<string> }) =>
```
lfaucon: Two UI requests:
- Could this short cut panel be centered vertically in its parent's div, so it will be at the same level as the image?
- My first reflex was to click on the shortcuts, but it did nothing. I believe it would be nice to have them clickable and user would chose whether to use the short cuts or just click. what do you think?


screenshot from 2017-08-17 11-10-17

ac/ac-imgClass/src/imgClassifier.jsx
```diff
@@ -0,0 +1,67 @@
...
+      </li>
+    )}
+  </ul>;
+
+export default ({ activityData, data, dataFn }: ActivityRunnerT) => {
+  const categ = activityData.config.categories || [];
+  const imgs = activityData.config.images || [];
+  categ.forEach((x, i) =>
+    Mousetrap.bind((i + 1).toString(), () => {
+      dataFn.objInsert(x, data.index + 1);
```
lfaucon: Right now the reactive data looks like this: 
```JavaScript
{
  1: "Light Side",
  2: "Light Side",
  3: "Light Side",
  4: "Light Side",
  5: "Light Side",
  6: "Light Side",
  7: "Light Side",
  index: 7
}
```
And the corresponding image urls are in the config of the activity. In our dataflow, the config of the activity is not transfered to the product, so operators that would use the output of this activity, would not know the urls. I believe the following format will be more usable:

```JavaScript
{
  1: { url: "...", category: "Light Side"},
  ...,
  index: 7
}
```

Stian mentionned in #388:
>  The output of the activity is the same list of URLs, with the tag received by the user.

ac/ac-imgClass/src/imgClassifier.jsx
```diff
@@ -0,0 +1,67 @@
...
+      </h4>
+      {data.index < imgs.length &&
+        <FlexDiv>
+          <ImgPanel url={imgs[data.index]} />
+          <ShortCutPanel categories={categ} />
+        </FlexDiv>}
+      {data.index >= imgs.length &&
+        <h1>
+          {'End of the activity'}
+        </h1>}
```
lfaucon: I believe
```JavaScript
<h1>
  {'End of the activity'}
 </h1>
```
is the same as:
```JavaScript
<h1>End of the activity</h1>
```

lfaucon:
Possible graph for testing:
```
{
  "graph": {
    "_id": "cj6g7wqjd0008326nic2zjfv2",
    "name": "starWarsImageClassifier",
    "createdAt": "2017-08-17T09:02:35.353Z",
    "duration": 120,
    "broken": false
  },
  "activities": [
    {
      "_id": "0-cj6g7i4bs0001326n94izkuy1",
      "title": "1",
      "startTime": 0,
      "length": 5,
      "plane": 1,
      "activityType": "ac-imgClass",
      "data": {
        "title": "Classify",
        "images": [
          "https://boygeniusreport.files.wordpress.com/2015/08/yoda.jpg?quality=98&strip=all",
          "http://www.bloomeration.com/wp-content/uploads/2017/05/star-wars-rebels2-1200x675.jpg",
          "http://digitalspyuk.cdnds.net/15/50/1600x800/landscape-1449498579-darth-vader-star-wars.jpg",
          "http://www.planwallpaper.com/static/images/star_wars_battlefront_dice.0.jpg",
          "http://a.dilcdn.com/bl/wp-content/uploads/sites/6/2017/07/d23-2017-battlefront-ii-02-1536x512.jpg",
          "https://lumiere-a.akamaihd.net/v1/images/Obi-Wan-Kenobi_6d775533.jpeg?region=0%2C0%2C1536%2C864&width=768",
          "http://static.srcdn.com/wp-content/uploads/Star-Wars-7-Character-Guide-Finn-Rey.jpg"
        ],
        "categories": ["Light Side", "Dark Side"]
      },
      "err": false
    }
  ],
  "operators": [],
  "connections": []
}
```

ac/ac-imgClass/src/imgClassifier.jsx
lfaucon: Also, we might want the list of urls to come from an operator.
In that case you should implement the `mergeFunction` to write in the reactive data both the list from the operator and the list from the config.

In the future we would like to even implement streaming operators which create input data repeatedly. for example half of the students keep generating images and the other half keep categorizing the images. I believe we would manage this through the merge function being called several times.

romainAA: okay, so what data input should I accept ?
I was thinking about something like this : 
`data: {images: [], categories: []}`
With urls in images and possible additional categories from an operator.

lfaucon: I suggest:
```JavaScript
{
  [key:string] : { url: string }
}
```

PR #403, opened 2017-08-17
Img view

romainAA:
Everything works well on my computer, as said in the last commit , even getting data input from imgClassifier works well.


lfaucon:
Work on #387

PR #405, opened 2017-08-18
Small changes

romainAA:
Fixed or improved the following things :
• overflow on graphEditor's sideBar
• removed the autofocus in the chooseActivity panel of the graph Editor
• sort the searchList of the chooseActivity and chooseOperator panels in alphabetic order
• added highlight to the search string even when it's in the long description (only visible when expanded)
• in brainstorm each student can now only vote once for each idea (+color for thumb up/down icons)


romainAA:
@lfaucon Now it seems good for every window size, you can check and can tell me if you notice something else that should be improved it just tell me I would change it

houshuang:
@romainAA In the future, please separate these into different pull requests. Especially the stuff you did in Brainstorm, and modifying the sidepanel in graph editor is completely different, and it would be easier to review if it were separate.

houshuang:
Preview is now broken for Brainstorm.

romainAA:
@houshuang  Can you develop what is broken ? everything works fine on all previews for me. Before choosing with and without data and after choosing with the config

houshuang:
@romainAA Much of this doesn't have to do with you, but since you're editing the brainstorm activity anyway, I cleaned it up a bit - some of it is also your code. Please test to make sure it still works as desired.

houshuang:
@romainAA You can no longer pull the panMap at the bottom all the way until the end, if the sidepanel is open, it does not go further than this. This works well on develop, I checked. It also goes all the way when the sidepanel is closed.

screen shot 2017-08-22 at 13 44 04

PR #410, opened 2017-08-21
Img view ui

lfaucon:
@romainAA: Can you test if the activity still works as it was before? and also review the code? and track possible bugs! Thanks


romainAA:
We need to close the ZoomView when we change the category, because if you open index 3 in 'all' and go to 'sea', the activity crashs

PR #412, opened 2017-08-23
Improving ac-brainstorm

houshuang:
- in brainstorm each student can now only vote once for each idea 
- color for thumb up/down icons
- some restructuring for legibility


houshuang:
I separated out your (and my) changes to ac-brainstorm from smallChanges - let's get this merged first

PR #419, opened 2017-08-25
Countdown

romainAA:


houshuang:
Can you add a -30 minutes button as well (only available if time > 30 seconds). Currently if you add time before starting the countdown, there is no way of undoing.

romainAA:
I'm going to do another PR after to change add time to a dropdown so I can do the - button on the same way

frog/imports/ui/StudentView/SessionBody.jsx
```diff
@@ -16,20 +18,74 @@ const getInitialState = (activities, d = 1) => {
...
 
-const SessionBody = ({ session }: { session: Object }) => {
+const SessionBody = ({
+  session,
+  currentTime
+}: {
+  session: Object,
+  currentTime: number
+}) => {
+  const tmp =
```
houshuang: better to call it a meaningful name, like secondsLeft

frog/imports/api/sessions.js
```diff
@@ -117,6 +127,8 @@ Meteor.methods({
       graphId: copyGraphId,
       state: 'CREATED',
       timeInGraph: -1,
+      countdownStartTime: -1,
```
houshuang: why not use strings, like 'notStarted', 'paused'? Removes need to look it up. We will try to add Flow definitions for database collections too, where we can check this.

frog/imports/api/sessions.js
```diff
@@ -117,6 +127,8 @@ Meteor.methods({
...
+      countdownTimeLeft: 30000,
```
houshuang: Just make this 0 or null, shouldn't matter if countdown is not running?

frog/imports/ui/StudentView/SessionBody.jsx
```diff
@@ -16,20 +18,74 @@ const getInitialState = (activities, d = 1) => {
...
+      <div>
+        {session.countdownStartTime !== -1 &&
+          <Countdown>
+            <h4>
+              {msToString(tmp)}
+            </h4>
+          </Countdown>}
+        <h1>NO ACTIVITY</h1>
+      </div>
+    );
```
houshuang: Make this 
```
  <Countdown>
 +            <h4>
 +              {msToString(tmp)}
 +            </h4>
 +          </Countdown
```

into a component, so you don't have to duplicate the code twice.

houshuang: actually thrice ;)

frog/imports/ui/StudentView/index.js
```diff
@@ -1,28 +1,31 @@
...
       {curSession
-        ? <SessionBody session={curSession} />
+        ? <SessionBody session={curSession} currentTime={currentTime} />
         : <SessionList sessions={sessions} />}
     </div>
   );
 };
 
 export default createContainer(() => {
+  if (TimeSync.serverOffset() > 500) TimeSync.resync();
```
houshuang: Why this? Did you see this in the documentation? Isn't serverOffset static and basically just means the difference between server and client? Nothing wrong with that in itself? As long as its stable...

frog/imports/ui/TeacherView/ButtonList.jsx
```diff
@@ -31,7 +47,11 @@ const ButtonList = ({
     {
       states: ['STARTED'],
       type: 'primary',
-      onClick: () => nextActivity(session._id),
+      onClick: () => {
+        updateSessionCountdownStartTime(session._id, -1);
+        updateSessionCountdownTimeLeft(session._id, DEFAULT_TIME_LEFT);
```
houshuang: Another thought, remember that a variable in Sessions collection can be an object, so you could have

type definition: 
countdown: {state: 'none'} | {state: 'paused', endTime: number}, {state: 'running', endTime: number} 
or something similar. The idea is to "outlaw illegal states"... Having to keep track of two different variables can easily bring bugs.

houshuang:
@romainAA On second thought, don't make it a pulldown for +/-, because it will make it very difficult to click +30 multiple times. just add two buttons :)

ac/ac-imgView/src/ActivityRunner.jsx
```diff
@@ -34,7 +34,6 @@ const ActivityPanel = ({
   webcam,
   setWebcam
 }) => {
-
```
houshuang: Try to check which files you are adding to git, instead of just adding all files. this is obviously not related to this PR.

frog/imports/api/openUploads.js
```diff
@@ -9,7 +9,6 @@ export const OpenUploads = new FS.Collection('openUploads', {
 OpenUploads.allow({ insert: () => true });
 
 export const uploadFile = (files: Array<any>, callback: string => any) => {
```
houshuang: same as above...

houshuang:
@romainAA Ping me when you want me to review.

romainAA:
@houshuang I think it's done, you can review it.
I've tested it and everything seems to work for me

PR #420, opened 2017-08-25
Adds ImageReload 

houshuang:
Adds a ImageReload component, which also takes an src=, but if it gets a 404, will retry downloading every second. This is particularly for ac-ImgView, where the URL might be available before the uploaded file has been fully processed. Also adds this component to ac-ImgView. Needs more testing.


houshuang:
@lfaucon Thanks for noticing. I'm not good at CSS but the latest changes seem to work both for the thumbnail view and the full screen modal.

lfaucon:
The merge function of ac-imgView is broken due to #428, waiting for #437 which fixes this issue

PR #430, opened 2017-08-28
Adding basic LTI functionality

houshuang:
This enables automatic logging through an LTI hook. It still needs to be much improved, adding more security, etc, but it's a first step, which enables further testing.


frog/imports/ui/App/FROGRouter.js
```diff
@@ -24,7 +24,9 @@ const Page = ({ isNotLoggedIn, isRedirect, isStudent, path, ready }) => {
     return <NotLoggedIn />;
   }
   if (!ready) {
-    return <p>Loading...</p>;
+    return process.env.NODE_ENV === 'production'
+      ? <img src="/images/Loading.gif" alt="" />
```
lfaucon: You use everywhere `"images/Loading.gif"`
But the file you added is called `"Spinner.gif"`

Furthermore Clients will not be able to access a file if it is in `frog/client`. I believe you should put this file in `frog/public`

houshuang:
@lfaucon Thanks, should work now

PR #431, opened 2017-08-28
Removing superfluous yarn.lock

houshuang:
With the new way of doing initial_setup, it does not care about the yarn.lock in subdirectories any more. Also upgrades to latest Yarn.


houshuang:
package-lock.json is generated by NPM when we install yarn... I'm not sure if we should commit it, I guess it doesn't make a big difference if we do.

houshuang:
@romainAA Added

PR #438, opened 2017-08-29
Preview on separate permanent URL

### LINKED ISSUE #383: Extract activity preview to a separate top-level tab
Currently the only way to preview an activity is to add a new empty activity in the graph editor. The preview functionality can also be very useful when developing new activities - we should take all the existing code, but also make it usable under a specific URL - like /preview/ac-video/0 (for the first dataset etc). The URL is important, because it means that the activity will stay during reloads (common during development) - and it needs to be updated when a user switches from one example data to another. You could even display the sidepanel open permanently, to let developers switch between trying to hardcoded example data, and a config they enter manually.

houshuang:
This enables previewing activities outside of the context of the graph editor, with a permanent URL that survives reloads. 

- you can directly type /preview/{activity-id} optionally followed by {example-id}. It also keeps track of whether showing data with ?showData=true
- you can click on preview in the top menubar (or just type /preview) to see a list of all activity types. This list is also shown if you close an activity preview
- when previewing an activity in the graph editor, there is a link to open the preview with a permanent URL (note, you cannot get back to the graph view from here)

This seems to work quite well with minimal code change. In the future, I could imagine adding the possibility of custom configurations, as well as the ability to see an activity inside a multi-window view with n number of panes.

Closes #383

### DIFF
--- a/frog/imports/ui/Preview/index.js
+++ b/frog/imports/ui/Preview/index.js
@@ -1,0 +1,51 @@
+// @flow
+import React from 'react';
+import { toObject as queryToObject } from 'query-parse';
+import { withRouter } from 'react-router';
+import { A } from 'frog-utils';
+
+import { StatelessPreview } from '../GraphEditor/Preview';
+import { activityTypes } from '../../activityTypes';
+
+const ActivityList = ({ history }) =>
+  <div>
+    <h2>Choose activity to preview</h2>
+    <ul>
+      {activityTypes.map(act =>
+        <li>
+          <A onClick={() => history.push(`/preview/${act.id}`)}>
+            {act.id}
+          </A>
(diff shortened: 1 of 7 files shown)


frog/imports/ui/Preview/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+import { StatelessPreview } from '../GraphEditor/Preview';
+import { activityTypes } from '../../activityTypes';
+
+const ActivityList = ({ history }) =>
+  <div>
+    <h2>Choose activity to preview</h2>
+    <ul>
+      {activityTypes.map(act =>
+        <li>
```
romainAA: should add the property 'key' => I get a warning in the console

PR #441, opened 2017-08-29
Fixing reactive data for preview

### LINKED ISSUE #382: Restart activity preview
We might want to check how the activity would appear in different cases.
For example, with the simple form, if you go to the preview, answer some questions and hit submit, then you will never be able to watch the preview again. You will only be stuck with the "Form(s) submitted" panel.

houshuang:
This makes the Preview reactive data structure live in memory, and adds a button to reset the data. Since it's in memory, it's  automatically reset on reload as well, of course.

Closes #382

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     jest: true
   },
   rules: {
+    'react/require-default-props': 'off',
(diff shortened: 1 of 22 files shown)


lfaucon:
@houshuang I let you merge this. there are a few conflicts

PR #442, opened 2017-08-30
Upgrading dependencies

houshuang:
Especially Flow and Prettier (which again leads to a ton of cascading changes - they seem to keep taking out and putting in parenthesis, but I guess they are getting more stable as time goes on).


houshuang:
Unfortunately crazy amount of Flow errors added with the new version - will take some time to figure out.

PR #443, opened 2017-08-31
Fix reactive data (second attempt)

### LINKED ISSUE #382: Restart activity preview

houshuang:
This makes the Preview reactive data structure live in memory, and adds a button to reset the data. Since it's in memory, it's automatically reset on reload as well, of course.

Closes #382

### DIFF
--- a/frog/imports/ui/GraphEditor/Preview.js
+++ b/frog/imports/ui/GraphEditor/Preview.js
@@ -1,15 +1,16 @@
 // @flow
-//
+
 import React from 'react';
 import { cloneDeep, uniqBy } from 'lodash';
 import Stringify from 'json-stable-stringify';
-import { A } from 'frog-utils';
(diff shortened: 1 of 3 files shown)


houshuang:
@lfaucon Should be ready to merge now

houshuang:
The collection already exists thing I think only comes up if you switch tabs very quickly - it's very hard to avoid any timing issues, but anyway the collections are only in memory, and it seems to work well. Instead of deleting and recreating collections (which totally didn't work), it's actually keeping track of uuids, and creating new collections whenever you reset data (not a problem, since it's all in memory). 

There are several issues related to how it deals with pre-configured activities, I'll file a separate issue.

PR #444, opened 2017-08-31
Small operator config fixes

houshuang:
Just some small fixes to op-create-groups and op-hypothesis config definitions.


houshuang:
@lfaucon Fixed

PR #445, opened 2017-08-31
Update the ActivityPackage dashboard API

lfaucon:
## What's new
- New field for in `ActivityPackageT` for dashboards
```JavaScript
{
  ...,
  dashboard: { 
    Viewer: ReactComponent,
    mergeLog: Function,
    initData: any
  },
  ...
}
```
- An example of dashboard on `ac-imgView` counting number of votes and number of uploads. No visualization, data pipeline only.

- In `server/mergeData` creates the document for the logs and initialise it with the `initData`

- in `server/mergeLogData` implement `Meteor.Method({'merge.log': ... }]` called from the loggers

- A component to navigate between the available dashboards when several activities are open

## Comments
- Removes the possibility of filtering selected activities (just like categories are filtered) in `ac-imgView`. It was badly implemented and was creating bugs. It can be easily re-implemented. This PR was already big enough!

## Incoming
- Nicer looking Dashboards for most of UNIL activities

### DIFF
--- a/ac/ac-imgView/src/ActivityRunner.jsx
+++ b/ac/ac-imgView/src/ActivityRunner.jsx
@@ -69,7 +69,7 @@ class ActivityRunner extends Component {
   }
 
   render() {
-    const { activityData, data, dataFn, uploadFn, userInfo } = this.props;
+    const { activityData, data, dataFn, userInfo, logger } = this.props;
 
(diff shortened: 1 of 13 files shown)


lfaucon:
@houshuang made the changes you suggested

PR #449, opened 2017-09-01
Countdown uses his own createContainer

### LINKED ISSUE #447: Countdown is hidden behind mosaic
screenshot from 2017-09-01 07-49-32

### LINKED ISSUE #448: StudentView is rerendered every second
Normally the mosaic allows to resize or move around the separate windows, but now, it all goes back to the initial position every second. What else gets recomputed every second? Database queries maybe? I believe this is due to the implementation of the countdown.

lfaucon:
- Moves the countdown component to its own file and give it the currentTime props through createContainer.

- Update the style so the countdown is on top of mosaic windows and with fixed position

- small clean up on SessionBody compoennt

Closes #448
Closes #447

### DIFF
--- a/frog/imports/ui/StudentView/Countdown.jsx
+++ b/frog/imports/ui/StudentView/Countdown.jsx
@@ -1,0 +1,46 @@
+// @flow
+
+import React from 'react';
+import { createContainer } from 'meteor/react-meteor-data';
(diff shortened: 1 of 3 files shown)


lfaucon:
solves #448 and #447

PR #450, opened 2017-09-01
Unifying copying/adding graph

houshuang:
I have a new function, addGraph in api/graphs, which takes an object with graph, activities, operators, connections, or empty. If empty, it just creates a new unnamed graph. If not, it does all the copying logic (updating ids in connections etc). It does all the updating of act/op/conn in memory, before writing the final version to Mongo (the previous case, we first wrote to Mongo, then updated the connections in the database through a loop).

Now the same logic is used for copying a graph when creating a session, in the graph editor, and when importing a graph. This makes it easier to add new logic in the future.

### DIFF
--- a/frog/imports/api/graphs.js
+++ b/frog/imports/api/graphs.js
@@ -7,13 +7,55 @@ import { Activities, Connections, Operators } from './activities';
 
 export const Graphs = new Mongo.Collection('graphs');
 
-export const addGraph = (
-  name: string = 'undefined',
-  graph: ?Object
-): string => {
-  const id = uuid();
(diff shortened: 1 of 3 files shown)


frog/imports/api/graphs.js
```diff
@@ -7,13 +7,55 @@ import { Activities, Connections, Operators } from './activities';
...
+  const newConn = graphObj.connections.map(connection => {
+    const id = uuid();
+    matching[connection._id] = id;
+    return {
+      ...connection,
+      _id: id,
+      graphId,
+      source: {
+        id: matching[connection.source.id],
+        type: connection.source.type
```
lfaucon: I wonder if we still need this field.

PR #452, opened 2017-09-01
Control operator

houshuang:
This is a substantial PR that introduces a new control operator. The output is defined in types.js. We also introduce a new form field, and I have rewritten the logic for custom form fields to support nested custom form fields, which also the socialAttribute field benefits from. We also need logic when copying graphs to ensure the activity reference is up to date.

I have included an operator which let's you explicitly exclude or include specific users (by username), either for all connected activities, or per activity. 

The logic for which activities to display in sessionBody could be simplified. Also note that it currently only deals with individual activity structures (not socialAttributes). This will be improved in a future PR, but since this PR touches on so many files, I think it's better to push it now (if it works correctly etc).

Also note that I added an object students to the globalStructure with names and student IDs, and this meant I had to update a bunch of tests for flow to be happy...  

Please test well :)


frog/imports/ui/StudentView/SessionBody.jsx
```diff
@@ -17,17 +21,62 @@ const getInitialState = (activities, d = 1) => {
       };
 };
 
-const SessionBody = ({ session }: { session: Object }) => {
+const checkActivity = (activityId, operators, connections) => {
+  const connectedNodes = connections
+    .filter(x => x.target.id === activityId)
+    .map(x => x.source.id);
+  const controlOp = operators.find(x => connectedNodes.includes(x._id));
```
lfaucon: This `find` method returns only one control operator if several exist. The code works well, but it means that if several control operators control one activity, only one of the operator will be used. I believed there should never be two control operators for the same activity, but the graph checks green. => Create issue 

screenshot from 2017-09-04 08-58-06

PR #464, opened 2017-09-06
Default config

### LINKED ISSUE #459: Config issue with default fields
It's impossible to delete the 3 in desired group size in the config of op-create-groups. Probably because this is set as a default value, which pops up instead of empty.

lfaucon:
Just look at the diff of the last commit to see the fix. The rest of the changes are only moving code around and splitting into new files.

Closes #459


lfaucon:
Yes, but I was waiting for your review before doing so.

houshuang:
Looks good to me. ANd I approve of your code cleanup effort as well! :) I don't know if there is a way of abstracting the logic to avoid having to copy it.

PR #471, opened 2017-09-07
First quiz

lfaucon:
Modified ac-quiz to accept latex text inputs, which will be useful for the first quiz of STAT101.

## Comments
- LatexWidget is a modified version of RadioWidget ([from here](https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/widgets/RadioWidget.js))
- Add one eslint exception triggered on the code of RJSF
- Add one node module to flow ignore


ac/ac-quiz/src/LatexWidget.jsx
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import React from 'react';
+import Latex from 'react-latex';
+
+export default (props: Object) => {
+  const { options, value, required, disabled, readonly, onChange } = props;
+  // Generating a unique field name to identify this set of radio buttons
+  const name = Math.random().toString();
```
houshuang: why don't you use uuid()?:)

houshuang:
This looks great! I made some small changes

- I installed KaTeX locally, I think it's better to always be able to run activities offline etc
- The preview was broken, because you changed the form schema. I used Valerie's form as a new example data, it's a nice example anyway

One thing I don't love is the resulting format of the data, like:

```js
{ form:
  0:
    radio: "Pour une variable discrète, le nombre de valeurs possibles est dénombrable." 
}
```

I would prefer something simpler like

```js
{ form: { 0: 1, 1: 2, 3: 1 } 
```

ac/ac-quiz/src/LatexWidget.jsx
lfaucon: That the code I copied from the RJSF RadioWidget. I fixed a few eslint errors on it, but didn't look at it much more. I guess the safest is to leave it as it is.

PR #478, opened 2017-09-11
Fixing/improving ac-ck-board

houshuang:
This PR rewrites some of the logic of ac-ck-board

- It adds the option of displaying a background image
- Either a background image or the four quadrants are sized to fit the entire screen
- All coordinates are translated back and forth between a fictional 1000x1000 grid stored in the database, and the actual screen size, taking the mid-point of a box, ensuring that a box is always located at the same location in relation to a background when scaling/distorting
- Font and box size is scaled up/down in accordance with screen size
- Also fixes the general logic to make it work

Possible future work:
- Add option to have image always preserve aspect ratio. A bit tricky, would have to either know the dimensions of the image and scale it properly, or use CSS to scale it while preserving ratio, and then listen to the size of the image to know how to scale boxes.


houshuang:
Updated against develop, added remote image. The onTouchTap is a problem with material-ui, which doesn't seem to cause any issue. Eventually we might want to switch away from material-ui or upgrade to 2.0.

Issue #482, opened 2017-09-12
Teacher often gets logged out during development automatic reload


houshuang:

houshuang:
Related to the use of ?parameters in URL. Might be fixed already, will wait for all PRs to be merged, and then investigate.

PR #479, opened 2017-09-12
Full-screen preview

houshuang:
This PR improves preview

- icons for actions, to distinguish from example cases
- adding ability to see activity in full-screen, making the "control panel" a floating object

Future work:
- eventually the Preview components (in GraphEditor and /) need some clean-up/refactoring
- the floating control panel shows below the Mosaic, not sure how to get it on top. Tried with zIndex

PR #483, opened 2017-09-12
Adding multiple users and dashboard to preview

houshuang:
This PR builds on #479 and #480 (and should not be merged before these two), and expands Preview by letting you see multiple users interacting, as well as the dashboard. Currently, the logic for this is kind of simplified, but it works - this could be done more nicely, but we eventually need to clean up the code/UI a bit anyway. How it works:

- You can add multiple activity windows, and each window gets its own user. 
- The users are automatically divided into groups of 2, and given their own instance. The third window automatically shows a dashboard (if one is available)
- Tested with ac-video and ac-imageView dashboards

Known issues:
- the control panel is still stuck behind Mosaic in full-screen view, would be great to get fixed
- sometimes some console warnings when adding windows etc. Doesn't seem to be a way to avoid redrawing all components when adding a window to React mosaic. Doesn't seem to break anything though.

Once we get #482 fixed, this will make it much easier to test both activities and dashboards, the interaction between students in groups, etc. 

Note that if activities have pre-loaded data, no analytics/log events will be sent based on these - only on subsequent user actions. We might eventually add log entries to exampleData, to be sent to mergeLog on load, especially useful to debug dashboards in different states.


frog/imports/ui/TeacherView/Dashboard.jsx
```diff
@@ -17,7 +17,7 @@ const Container = styled.div`
   flex-direction: row;
 `;
 
-class DashboardComp extends Component {
+export class DashboardComp extends Component {
```
lfaucon: I get this error when the DashboardComp mounts
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the _DashboardComp component.
```

lfaucon:
I don't observe the control panel stuck behind Mosaic in full-screen view

houshuang:
I fix the control panel :)

PR #492, opened 2017-09-19
Things from Tallinn demo

houshuang:
Small changes from Tallinn demo:

- op-aggregate and op-aggregate-p2
  These should have tests, and perhaps they should be unified into a single operator, but they work right now, and I would rather merge them than have the code floating around
- ac-brainstorm style, valid config changes
- ac-ck-board merge function updated to new coordinates
- ac-video now writes player state to reactive data, but only reads from reactive data on reload
  somehow ac-video does not work well with countdown currently
- ensure that we never have more than one session called DEBUG (can cause bugs)
- updating yarn dependency


ac/ac-video/src/index.js
```diff
@@ -32,32 +32,47 @@ export const config = {
       title: 'URL of video'
     },
     playing: {
-      type: 'string',
+      type: 'boolean',
       title: 'Should video begin auto-playing?',
-      enum: ['true', 'false']
```
lfaucon: Haha, that old code is funny. Good to see that we learnt some stuff

PR #503, opened 2017-09-24
Login by selecting name on a list

houshuang:
This adds a new button in the session view, which lets teachers edit a list of all students that should have access to the current session (it's a text area, one name per line, it gets sorted/trimmed/etc afterwards). The list is stored as an array studentlist on the session object. 

If students enter the URL to a session, without being logged in, or using the ?login parameter, and that session has a list of students, the list will be shown, and once the student chooses, he/she will be logged in.

This is intended as one more possible way of giving access to login for FROG. It will in particular be used by the OSIP script.

### DIFF
--- a/frog/imports/api/sessions.js
+++ b/frog/imports/api/sessions.js
@@ -28,8 +28,11 @@ Meteor.methods({
         return;
       }
       sessionCancelCountDown(session._id);
-      Meteor.call('flush.session', session._id);
       const newSessionId = Meteor.call('add.session', graphId, { debug: true });
+      Sessions.update(newSessionId, {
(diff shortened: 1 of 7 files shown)


houshuang:
This will be used with an operator that distributes different students to different social groups, however this is where we would really need streaming operators - since it's a problem to have to wait for every student to join before the graph starts... To overcome this, I now run a check in runSession, creating all missing students, and making sure all students are joined to the session. This means the operator will generate social data for everyone when run, even if they join later. This works, but we should think about more flexible ways.

We might also re-introduce the `presence` stuff, to be able to distinguish between students who have joined a session, versus students currently online.

PR #505, opened 2017-09-24
Operator mapping list of student names to social attributes

houshuang:
This operator maps a list of names to the attributes of one or more social groupings. The list is pure text, but this makes it easier to for example paste in a CSV from a spreadsheet etc. All the config data should be thoroughly validated to ensure for example that each student has the right number of attributes.

Also a small change to the config validator, which gives better stack traces if a validation function fails.

It even has a test :)


op/op-social-name/src/index.js
```diff
@@ -0,0 +1,105 @@
...
+  shortDesc: 'Maps students to groups by name',
+  description:
+    'Input a list of student names, and group names, and it matches the students to the groups'
+};
+
+const config = {
+  type: 'object',
+  required: ['groupnames', 'defaultgroup', 'studentmapping'],
+  properties: {
+    groupnames: {
```
lfaucon: I would suggest to change the naming of the config fields to be consistent with the rest of the Engine. I believe:
- `groupnames` should be `groupingKeys`. This one is particularly confusing as groupnames sounds like it would be groupingValues, while it is groupingKeys.
- `defaultgroup` should be `defaultGroupingValues`

houshuang:
@lfaucon Please review again, I have updated the naming, and added test of no registered students, and several tests of the config validator, as well as updating on the latest develop branch. Should be ready for merge.

PR #508, opened 2017-09-25
Add streaming feature

lfaucon:
## Changes

Any activity can be configured to stream to a P3 activity. Then its ActivityRunner gets a props `stream` which is a reactiveFunction on the data of the P3 activity instance. I find it pretty simple and I think it will be easier to implement next week script that way rather than with a fake P3 activity.

Graph for testing: https://github.com/chili-epfl/FROG-graphs/blob/master/STREAMING_EXAMPLE.frog 

## Future Work

Need to review the UI of ac-imgClass to handle the external updating of the image list.

### DIFF
--- a/ac/ac-imgClass/src/imgClassifier.jsx
+++ b/ac/ac-imgClass/src/imgClassifier.jsx
@@ -79,6 +79,9 @@ export default ({ activityData, data, dataFn }: ActivityRunnerT) => {
     categ.forEach((x, i) => Mousetrap.unbind(shortCuts[i]));
   return (
     <div style={{ margin: '1%', height: '100%' }}>
+      <pre>
+        {JSON.stringify(data, null, 2)}
+      </pre>
       <h4>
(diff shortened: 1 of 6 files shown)


houshuang:
Will review this. Are you suggesting we make this a regular feature of FROG, or is this just to enable the UNIL script?

lfaucon:
I think it's a simple and stable way to implement what we want for next week. Also it could be a basis for implementing streaming operators.

The possible next steps for making this feature into streaming operators could be for example:
- add operator layer which could transform the data on the way
- add merging layer for the activity receiving the streaming data
- plug the flow directly from dataFn and not form the extra props `stream` that I implemented
- handle plane differences. For now I only stream to a P3 because it is easier as there is only one instance.

houshuang:
@lfaucon I'm guessing it's easier to close this and create a new one based on the actual code in unil branch, but up to you. Let me know when you want me to review for merging to develop.

PR #541, opened 2017-10-11
Move streamer to a Meteor.call

lfaucon:


lfaucon:
You are correct! I'll fix that by tomorrow

PR #548, opened 2017-10-12
UNIL - only frog-utils and frog

houshuang:


houshuang:
I read through this, and I'm actually pretty happy about merging pretty much all of it - we have to put back all the activity/operator imports of course (I really wish we'd figure out a way of doing that automatically, and not have to change core code to add an activity). Are you happy with trying to merge this to develop? Feel free to have a read-through. 

I know we have to come up with a development alternative for uploading files, otherwise I don't think we're breaking any existing scripts or activities with what we've changed. We also need to check how much there is to do to make it "green".

PR #550, opened 2017-10-13
Merging only frog-utils from UNIL branch

houshuang:
Trying to slowly merge Unil branch into develop, began only with frog-utils, to minimize Eslint/Flow etc errors. This code should be mainly independent, apart from one tiny thing I had to do in ac-imgView/components/webcam, which we'll fix when we merge that activity.

### DIFF
--- a/frog-utils/src/index.js
+++ b/frog-utils/src/index.js
@@ -1,8 +1,8 @@
 // @flow
-import React, { Component } from 'react';
+import React from 'react';
 
 import { compose, withHandlers, withState } from 'recompose';
-import { shuffle, omit } from 'lodash';
+import { shuffle } from 'lodash';
 
 export {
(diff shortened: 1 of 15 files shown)


frog-utils/src/index.js
```diff
@@ -131,3 +135,87 @@ export const wordWrap = (text: string, maxLength: number): string[] => {
...
+    .slice(0, n)
+    .join('');
+
+type ReactivePropsT = {
+  path: string | string[],
+  dataFn: Object,
+  type: 'textarea' | 'textinput'
+};
+
+export class ReactiveText extends Component {
```
lfaucon: Would it be nicer to move this Component to `frog-utils/ActivityComponents`. Along with the type definition

PR #552, opened 2017-10-13
Merge the core

houshuang:
I am basing this on mergeutils now to see the right diff, but mergeutils should be merged first, and then this should be merged into develop. It's a cleaned up version of the base unil changes, excluding activities/operators.

There are some things I still want to revise, like the way file uploading works, but I think this code is pretty much all good, and ready to merge. We need to make sure everything works well on develop etc, and the best way is to start working off develop as soon as possible. This also means that new features (logging etc) can be based directly off develop (and merged into unil if we need to keep a separate branch).


frog/client/main.html
```diff
@@ -1,5 +1,9 @@
 <head>
   <title>FROG</title>
+  <script src="https://cdn.ravenjs.com/3.17.0/raven.min.js" crossorigin="anonymous"></script>
```
houshuang: Do you know of an easy way to make this configurable? It does not yet have access to Meteor.settings at this stage.

frog/imports/api/uploads.js
```diff
@@ -1,9 +1 @@
-// @flow
-
-import { FS } from 'meteor/cfs:base-package';
-
-export const Uploads = new FS.Collection('uploads', {
-  stores: [new FS.Store.FileSystem('uploads')]
-});
-
-Uploads.allow({ insert: () => true });
+export const Uploads = { insert: () => {} };
```
houshuang: Putting this here for now to make Flow happy. Will rethink a bit how to do this to make sure it works both in development and production.

frog/imports/ui/App/index.js
```diff
@@ -66,7 +74,8 @@ class FROGRouter extends Component {
     const hasLogin = query.login;
 
     if (this.state.mode !== 'loggingIn') {
-      if (process.env.NODE_ENV !== 'production') {
+      if (true) {
```
houshuang: Eventually I want to make this more secure again

frog/imports/api/uploads.js
lfaucon: ok

frog/server/main.js
```diff
@@ -19,19 +20,19 @@ import { Sessions } from '../imports/api/sessions.js';
 import { Products } from '../imports/api/products.js';
 import { Objects } from '../imports/api/objects.js';
 
+if (process.env.NODE_ENV !== 'production') {
+  // eslint-disable-next-line global-require
+  require('longjohn');
+}
+
+process.setMaxListeners(100);
```
lfaucon: Did this make any change for us?

frog/server/sharedbCache.js
```diff
@@ -0,0 +1 @@
+export const Cache = {};
```
lfaucon: Love that file :) Should it have flow annotation?

lfaucon:
@houshuang I made a small change on how we generate random slugs. You can give it a look, then merge it

PR #567, opened 2017-10-17
op-add-category

houshuang:
Operator to add categories based on instances. From osip demo. Not ready.


houshuang:
@lfaucon Before I fix this, I'd like your input on the approach. Basically the idea was that I wanted to combine input from multiple groups (which also requires product operators accepting multiple inputs, but that's a separate issue). I needed to do two things - first was to add a category/label to the ideas generated by a single activity (first students were asked to brainstorm problems, then solutions), and then also add a category for which instance (group) generated the different ideas (we had them grouped in natural sciences, humanities etc)... Then I had a version of ck-board which let you turn on and off the keys for various categories (show all problems, show problems and solutions only for physicists etc).

The way I did it was add an option to op-aggregate to add instances as categories, and created a new operator which only adds a specific category to all items... This worked, but seems a bit confusing.

Of course, in the brainstorm activity I could add an option "Add this category to all items", but then I would have to add this option to any other activity as well.

I'm also thinking ahead at this concept of learning components, that we could have some kind of a "wrapper", with a bunch of default fields "created_at, created_by_user, created_by_instance, created_in_activity, categories, votes/score, learning_component_type, payload" - where for example learning_component_type could be 'brainstorm', and payload {title: ..., content: ...}.

We could even imagine adding stuff like versioning (it was created in activity1, then modified in activity2 - let me see original version), and even parent-child (let's say you had an op-distribute which takes an incoming event and sends it to three different student groups, who are able to edit it, it would basically create three copies, with parent_id pointing to the same original item, so that you could easily group the three different edited answers together etc)...

Lot's of things to think about here - but if we could have a bunch of basic learning components (videos, photos, textarea, drawing, circuit board), and we could create, list, sort, vote, organize graphically (ck-board), link between (concept map etc), ... AND create "compound learning components" like a template (ie. an observation is an image plus two fields, "my hypothesis is", and "my reasoning is"), and this then could be passed around in exactly the same way, and always know how to render itself, make itself editable, sortable, taggable etc... I think we'd have something incredibly powerful. 

And I also think this is worth discussing in depth, because it will have important consequences for how we deal with concurrent activities, and live operators. (One of the big constraints for me with live operators etc is the idea that we would have to re-run the operator every time someone typed a letter... but if we operate at the level of learning components, and just re-run the operator everytime someone adds a new learning component, that becomes much more doable... ANother issue is the case of data consistency - if we have two activities open at the same time, say ck-brainstorm and ck-board, and both allow editing, what happens when an activity is created in ck-brainstorm, sent to ck-board, and then edited? Does the change get sent back to ck-board? does it stay in ck-brainstorm? What happens when the item gets edited in ck-brainstorm after having been edited in ck-board?)

(I have also thought about whether it would be possible to simply have two activities share the same reactive data structure, if their internal data structures matched completely... still not sure if that is a good idea)

PR #570, opened 2017-10-18
Making upload work in development mode

houshuang:
This adds an API which accepts any files, stores them in /tmp/ and returns them when requested. This means that ac-image etc works without any other dependencies. The upload/access method is exactly the same for production, the only difference is that the URL is different (pointing to a Minio server). 

It also modifies the automatically reloading images to reload maximum 4 times, and to wait longer for each time (first 1 second, then 2, then 3 etc).

You might not be able to test this in ac-image, without also accepting the previewStream PR.


lfaucon:
I think you made a mistake in `frog/server/minio.js` which I fixed. If you are ok with it, you can merge

PR #579, opened 2017-10-20
Op mix categories

lfaucon:
Only works with `all` incoming data and grouped target for now

Example graph to test it
[TEST_OP_DIS_CAT.frog.zip](https://github.com/chili-epfl/FROG/files/1401327/TEST_OP_DIS_CAT.frog.zip)


houshuang:
Thank you for zipping the graph file, to save 1.2 kb :)

op/op-distribute-category/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  shortDesc: 'Distribute categorized products objects to instances',
+  description: 'Distribute categorized products objects to instances.'
+};
+
+const config = {
+  type: 'object',
+  properties: {}
+};
+
+const pickRandom = arr => arr[Math.floor(Math.random()) * arr.length];
```
houshuang: You can use sample from lodash
http://devdocs.io/lodash~4/index#sample

op/op-distribute-category/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+const checkObject = obj =>
```
houshuang: Eventually we'll have strong type checking so we don't need this :)

lfaucon:
@houshuang Github does not support yet `.frog` extensions. Next time I'll share the graph file as a pdf if you prefer :p

PR #580, opened 2017-10-20
Op-filter

lfaucon:
example graph to test:

[TEST_OP_FILTER.frog.zip](https://github.com/chili-epfl/FROG/files/1401670/TEST_OP_FILTER.frog.zip)


op/op-filter/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+const config = {
+  type: 'object',
+  properties: {
+    field: {
+      type: 'string',
+      title: 'Filtering field'
+    },
+    value: {
+      type: 'string',
+      title: 'Filtering value (leave empty for testing with booleans)'
```
houshuang: Maybe specify whether this means matching values will be included or excluded?

op/op-filter/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+    remove: {
+      type: 'boolean',
+      title: 'Remove matching objects?'
```
houshuang: Ah, just saw this. Never mind :)

PR #582, opened 2017-10-23
MergeLog sharedb connection

lfaucon:
Remember the bug about subscription to sharedb that required restarting the server after the start of the activity for streams? I found the same thing happening for logs.

Also removes unnecessary config fields in ac-image


houshuang:
What is the procedure to reproduce the error that this PR solves? Does #584 also work? I think its cleaner to not duplicate, plus I don't think we need subscribe here, fetch should be enough. If you can confirm, we can merge these two PRs. Thanks.

PR #594, opened 2017-10-26
Export data

houshuang:
I'm planning to work more on this, but it should already be mergeable, if you want to look at it. 

I want to add export of a full graph (doing this for each activity, and packaging it up), as well as some whole graph tables, like user mapping in a CSV file.

### DIFF
--- a/frog/imports/ui/TeacherView/ShowInfo.js
+++ b/frog/imports/ui/TeacherView/ShowInfo.js
@@ -4,25 +4,42 @@ import { Inspector } from 'react-inspector';
 import { createContainer } from 'meteor/react-meteor-data';
 import { A } from 'frog-utils';
 import Modal from 'react-modal';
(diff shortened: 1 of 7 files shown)


frog/imports/ui/TeacherView/ShowInfo.js
```diff
@@ -2,12 +2,40 @@
...
 import { connect } from '../GraphEditor/store';
 import { Objects } from '../../api/objects';
 import { Activities, Operators } from '../../api/activities';
+import { activityTypesObj } from '../../activityTypes';
 import { Products } from '../../api/products';
 
+const downloadExport = (item, object, product) => {
+  const aT = activityTypesObj[item.activityType];
+  const zip = new JSZip();
+  const img = zip.folder(item._id);
```
lfaucon: `img` stands for image? does that make any sense here?

PR #602, opened 2017-10-27
Improving Preview and Dashboards

houshuang:
This PR should fix 

- problem with preview not reloading with multiple windows
- problem with dashboard not loading data
- updating ac-video dashboard to new log format

Please test both in Preview and in regular mode (session).

### DIFF
--- a/frog/imports/ui/TeacherView/Dashboard.jsx
+++ b/frog/imports/ui/TeacherView/Dashboard.jsx
@@ -3,6 +3,7 @@
 import React, { Component } from 'react';
 import { Meteor } from 'meteor/meteor';
 import { createContainer } from 'meteor/react-meteor-data';
+import Spinner from 'react-spinner';
 
 import { withState } from 'recompose';
 import { Nav, NavItem } from 'react-bootstrap';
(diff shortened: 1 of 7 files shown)


lfaucon:
Not looking at the code yet, just testing. 

## Problems

### 1
From preview/ac-image, open 3 students and the dashboard. Switching from one example to the next does not refresh the dashboardView, for this we need to remove/add the dashboard

### 2
Uploads are for some reason not being counted

### 3
When using the preview from an activity in the graph and filling up the dashboard, then going to see the preview of that same activity type, the dashboard data has not been reinitialized.

frog/imports/ui/TeacherView/Dashboard.jsx
```diff
@@ -20,67 +21,53 @@ const Container = styled.div`
...
+  unmounted: boolean;
+  conn: any;
 
   constructor(props: Object) {
     super(props);
     this.state = { data: null };
   }
 
-  componentDidMount() {
-    this.mounted = true;
```
lfaucon: Oh no! Why would you do that ? Isn't it not less easy to read `this.mounted` and `!this.mounted` than `!this.unmounted` and `this.unmounted`

PR #604, opened 2017-10-27
Check for groupingKey in op-distribute

houshuang:
Absence of grouping key, unless individual: true, leads to crash.


houshuang:
Ping

lfaucon (approved):

PR #613, opened 2017-10-31
Fix text

lfaucon:


ac/ac-induction/src/TestPart/Test.jsx
```diff
@@ -66,13 +66,8 @@ export default ({
...
       <ExContainer style={{ padding: '20px' }}>
         <TestCorrectionDiv style={{ justifyContent: 'space-evenly' }}>
-          <h3>This image corresponds to an example of the concept </h3>
-          <Switch
-            tmpList={tmpList}
-            feedback={feedback}
-            data={data}
-            dataFn={dataFn}
-          />
+          <h3>This example is a {title}</h3>
```
romainAA, line 69: I think it was the same before, but it really looks like an affirmation instead of a sentence linked to the toggle button. This might confuse the student

ac/ac-induction/src/TestPart/ResponsePanel.jsx
```diff
@@ -83,8 +79,8 @@ const FalsePanel = ({
     <Main>
       <h4>
         {(even
-          ? "Select properties that exclude him of being an example of the concept '"
-          : "Select what would be missing to be an example of the concept '") +
+          ? "Select one property that excludes it from being a '"
+          : "Select one property that is missing to be a correct example of '") +
```
romainAA: one property may not be enough to make it a correct example

ac/ac-induction/src/index.js
```diff
@@ -29,27 +29,27 @@ const meta = {
...
-          'they all have the same color',
-          'they all have the different shapes',
-          'they all have the different colors',
-          'exactly 2 have the same shape',
-          'exactly 2 have the same color'
+          'At least one element is not filled',
+          'At most one element is filled',
+          'The three elements have the same shape',
+          'The three elements have the same filling',
+          'The three elements have the different shapes',
```
romainAA: Still the 'the different' problem

romainAA (changes requested):

ac/ac-induction/src/TestPart/Test.jsx
lfaucon: I don't think it is a problem. I believe people are used to the type of question: Affirmation + True/False ? But we should keep that in mind for our pre-experiment

romainAA (approved):

PR #614, opened 2017-10-31
Induc test

lfaucon:
@romainAA this PR comes after #613 

For the onClick of the Switch component, you were copying a whole list, then modifying it locally, then sending it back through dataFn to the reactive data. It is more efficient to simply change the few fields that you want to update. Look at how I changed it. 

Please, make sure that everything is working at least as well as before


romainAA (dismissed):
Everything I tried worked well, but we should merge it after the modif and merge of inducText I guess.

PR #615, opened 2017-10-31
Improving log and session export

houshuang:
This ended up making quite a few changes. The main goal is to improve the export of individual activities, and of an entire graph. 

There is some new code to format time strings (I decided to just copy this into frog-utils, instead of adding a dependency - libraries seemed much bigger). 

I refactored a bit the stuff in next.activity to calculate which activity is next, and added a function which calculates the "sequence number" for all activities in a graph - this is used in naming for the log export.

I have a bunch of code to clean up the result of exportData functions. It's possible I'm doing this not in the most efficient way, but I wanted it to be as easy as possible for functions to provide an exportData function, then I clean empty columns (with lot's of options, we might have lot's of empty columns, like votes for ac-image), and I also either replace instanceIds with userids etc if p1, otherwise I generate a separate file with usernames/ids mapped to instanceIds.

The exact API for exportData might change - if I find that I do the same stuff for every activity, I might try to simplify it a bit. 

Also there might be changes depending on people's feedback on the export format. But I think in general this is going in a pretty good direction.


frog/imports/api/sessions.js
```diff
@@ -140,8 +140,14 @@ export const updateOpenActivities = (
 ) => {
   if (Meteor.isServer) {
     Sessions.update(sessionId, { $set: { state: 'WAITINGFORNEXT' } });
+    Sessions.findOne(sessionId).openActivities.forEach(activityId =>
+      Activities.update(activityId, { $set: { actualClosingTime: new Date() } })
+    );
     openActivities.forEach(activityId => {
       Meteor.call('dataflow.run', 'activity', activityId, sessionId);
+      Activities.update(activityId, {
+        $set: { actualStartingTime: new Date() }
```
lfaucon: I think that this would be a problem. If an activity runs in parallel to several other activities, then the `actualStartingTime` will be overwritten and only the last starting time will remain. There must be a way to do an update query that does not overwrite.

houshuang: Good point, my latest push fixes this - should be quite robust. Of course, logic has to be revised if we have flexible time in the future, but for now seems to work.

lfaucon (approved):

PR #621, opened 2017-10-31
React 16

houshuang:
So I tried upgrading React and almost all of our other deps (apart from ESLint, and Flow), and amazingly it seems to work perfectly fine - I'm not getting any warnings etc! However, I think we should test this more thoroughly before merging - try every activity, all the different functionality etc, while watching the console.

I also added an example ErrorBoundary (just copy+pasted, not saying we'll keep this) from Sentry. The showInfo in teacherview is broken (just to test it), so if you go to teacher view and click on an activity/operator, you can see how it looks - the message in console is a bit more informative, and we have a custom display - this could be changed to be per-activity/dashboard etc in the future.

Next step is testing this well and getting it merged, and then upgrading flow (more work).


houshuang:
I tried clicking everywhere (preview, running graph etc), and I got no errors or missing functionality. I recommend merging this now, unless you found anything, and if there are hidden bugs, we'll find them gradually.

lfaucon (approved):

PR #629, opened 2017-11-01
Induc dashboard

romainAA:
Creates a dashboard for the induction activity.
The dashboard display 2 progress bar for each part of the activity:
• one for the progression on the part by all instances
• one for the progression the number of instances that are done with the part


ac/ac-induction/src/Dashboard.jsx
```diff
@@ -0,0 +1,142 @@
...
+                    </div>
+                  )}
+                </div>
+              </div>
+            ) : (
+              <div key={x.toString() + i.toString()} />
+            )
+        )}
+      </div>
+    </div>
```
lfaucon: Your dashboard has two very similar parts, that's what React components are for. Make this into one react component that you use twice with different data and title

ac/ac-induction/src/Dashboard.jsx
```diff
@@ -0,0 +1,142 @@
...
+
+const Viewer = ({ instances, data, config }: Object) => {
+  const parts = [
+    1,
+    config.hasExamples ? config.nbExamples : 0,
+    config.hasTestWithFeedback ? config.nbTestFeedback : 0,
+    config.hasDefinition ? 1 : 0,
+    config.hasTest ? config.nbTest : 0
+  ];
+  const nbInst = Object.keys(instances).length; // need to change to have the instance instead of the users
```
lfaucon: It is changed isn't it, you can remove the comment then.

lfaucon (changes requested):
It looks nice !

lfaucon (approved):

PR #630, opened 2017-11-01
induc improvement on FB

romainAA:
I changed the respected properties that were printed on the feedback modal for the image of the current example


lfaucon (approved):
I'll merge this, but there is still work to do on the UI and on the explanation. Also there are some times when the selected example is the same as the current example. We should identify when that can happen and maybe think about an other type of feedback

PR #631, opened 2017-11-01
Induc disable submit

romainAA:
The submit button is now disabled when no property is selected.
I still have a problem: the setDisable in the truePanel and falsePanel works well to enable the button,
but when you deselect all properties, you need to refresh the page in order to redisable the button.

### DIFF
--- a/ac/ac-induction/src/TestPart/ResponsePanel.jsx
+++ b/ac/ac-induction/src/TestPart/ResponsePanel.jsx
@@ -30,6 +30,7 @@ export default (props: Object) => {
 
 const TruePanel = ({
   title,
+  setDisable,
   properties,
   feedback,
   tmpList,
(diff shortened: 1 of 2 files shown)


ac/ac-induction/src/TestPart/Test.jsx
```diff
@@ -9,70 +10,96 @@ import Correction from './Correction';
...
+              feedback,
+              data,
+              dataFn,
+              setDisable
+            }}
+          />
+          <ExButton
+            className="btn btn-default"
+            onClick={clickHandler}
+            disabled={disable}
```
lfaucon: You are doing quite complicated things here. I think you should just have here something like: 
`disabled={tmpList[data.indexCurrent].selectedProperties.length < 1}` 

I think that would be reactive, unlike your current implementation

ac/ac-induction/src/TestPart/ResponsePanel.jsx
```diff
@@ -107,19 +110,20 @@ const FalsePanel = ({
...
+                      )
+                    : [...tmpList[data.indexCurrent].selectedProperties, x],
+                  [
+                    feedback ? 'listIndexTestWithFeedback' : 'listIndexTest',
+                    data.indexCurrent,
+                    'selectedProperties'
+                  ]
+                );
+                setDisable(
+                  tmpList[data.indexCurrent].selectedProperties.length < 0
```
lfaucon: why `< 0` and not `< 1` ?

lfaucon (changes requested):

romainAA:
Indeed, at first I had a problem: the button was shown disable, but the clickHandler was still working, that's why I put a state to be able to access it. But I just selected the clickHandler as onClick only if tmpList[data.indexCurrent].selectedProperties.length > 0 and It seems to work well and to be easier.

lfaucon (approved):

lfaucon (approved):

PR #636, opened 2017-11-02
Adding ability to duplicate activities in Graph Editor

houshuang:
For the record, this took 22 minutes, including looking at my PR's problems with you :) (Edited by lfaucon)

### DIFF
--- a/frog/imports/ui/GraphEditor/store/activityStore.js
+++ b/frog/imports/ui/GraphEditor/store/activityStore.js
@@ -59,10 +59,7 @@ export default class ActivityStore {
   duplicateActivity = () => {
     if (store.ui.selected instanceof Activity) {
       const x = duplicateActivity(store.ui.selected.id);
-      this.all.push(
-        new Activity(x.plane, x.startTime, x.title, x.length, x._id, x.state)
-      );
-      store.addHistory();
+      this.mongoAdd(x);
     }
   };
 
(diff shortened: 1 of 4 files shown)


frog/imports/ui/GraphEditor/store/activityStore.js
```diff
@@ -55,6 +56,17 @@ export default class ActivityStore {
   }
 
   @action
+  duplicateActivity = () => {
+    if (store.ui.selected instanceof Activity) {
+      const x = duplicateActivity(store.ui.selected.id);
+      this.all.push(
+        new Activity(x.plane, x.startTime, x.title, x.length, x._id, x.state)
+      );
```
lfaucon: With this four lines, you create a new activity in db and a new one in the store. I observe two activities created with a weird behaviour (they follow each other)

lfaucon (changes requested):

frog/imports/ui/GraphEditor/store/activityStore.js
houshuang: Feel free to improve the logic. Initially I created only in the store, but then of course it doesn't get the activity type and the config. Then I only created in the db, but it didn't automatically get updated in the store - probably the store only listens to server-created db events, not locally created. So then I do this. And then the addHistory should make sure that the store is flushed to the db, and that everything is up to date.

Can you be more specific about your weird behaviour (following each other)? Does this mean that if you drag one activity, the other one follows along? This sounds very strange.

lfaucon (approved):

PR #637, opened 2017-11-02
Instance preview db

romainAA:
I added the list of instances to the preview, if you could just check that it is indeed the only place I had to change :)


romainAA:
(It's build on top of the induc dashboard branch so that I could test my dashboard with it, so please merge inducDashboard before)

frog/imports/ui/Preview/Preview.js
```diff
@@ -310,6 +310,7 @@ export const StatelessPreview = withState(
                     config={activityData.config}
                     reload={reload}
                     doc={dashboard}
+                    instances={users.map((u, i) => Math.ceil(i / 2))}
```
lfaucon: I think that is not correct, in that list instances will appear as many times as they have users.

lfaucon (changes requested):

houshuang:
@lfaucon Is this done now? would be good to get it merged

lfaucon (approved):

PR #639, opened 2017-11-06
Induc part 'end'

romainAA:
Even if this part is not necessary, I think it would be nice for the student to have an overview on how he did especially on the test without feedback. But the UI don't look really nice, if you have ideas about that please share them :)

### DIFF
--- a/ac/ac-induction/src/End.jsx
+++ b/ac/ac-induction/src/End.jsx
@@ -1,34 +1,47 @@
 // @flow
 
-import React, {Component} from 'react';
+import React, { Component } from 'react';
 
 export default class End extends Component {
-
-  // constructor(props: Object){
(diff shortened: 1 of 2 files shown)


ac/ac-induction/src/End.jsx
```diff
@@ -1,14 +1,47 @@
...
+          <Result list={this.props.data.listIndexTest} />
+        )}
+      </div>
+    );
+  }
+}
+
+const Result = ({ list }: Object) => (
+  <div style={{ display: 'flex', flexDirection: 'row', margin: '10px' }}>
+    Test with feedback :
```
lfaucon: One is "Test with Feedback", the other one is just "Test"

lfaucon (approved):

PR #649, opened 2017-11-09
Add text in ac-classifier and update selection display

lfaucon:


lfaucon:
@houshuang You can look in the meta example what data type I chose for text learning components. Suggest changes on it, or adapt the other ac/op to it

houshuang (approved):
Looks good to me. I rewrote the ObjectPanel to make the logic a bit more separate from the content types, and I also added the ability to move left and right using arrow keys. Feel free to merge if you're happy with these changes.

PR #654, opened 2017-11-09
Updating example definition in activity template

houshuang:


templates/activity/src/index.js
```diff
@@ -8,7 +8,11 @@ const meta = {
   shortDesc: 'New activity, no description available',
   description: 'New activity, no description available',
   exampleData: [
-    { title: 'Case with no data', config: { title: 'No data' }, data: {} }
+    {
+      title: 'Case with no data',
+      config: { title: 'No data' },
+      activityData: {}
+    }
```
lfaucon, line 15: I don't understand this change. I have been using this field `data` and it works as it should

PR #656, opened 2017-11-10
Changing throw to console.error in dataStructureTools

houshuang:
Not sure what the best way of dealing with this is - but at least this gives us warnings in console, whereas the previous way (throw), when called from within a Promise, only gives us "error undefined", which takes forever to try to hunt down.


houshuang:
Argh, and of course Flow is not happy because I might not return anything (if there is an error), but if I make the function say that it might not return anything, every function relying on it will have to check... Anyway this is something that should never happen - but if it does, during development, we want a proper error message, not undefined...

lfaucon (approved):

PR #657, opened 2017-11-10
Fixing EnhancedSchema (again) and some other related bugs

houshuang:
This PR is another effort to fix/improve EnhancedForm, which still wasn't behaving properly.

What I am doing now is to let it revert to being an uncontrolled form (ie. we don't pass formData into the props on every change). However, we use onChange to store the changes, which are both stored in the database, and used to recalculate the schema (and then the schema is updated only if it has actually changed). Note the difference between this.formData, which does not cause a re-render when updating, and this.state.formData, which does. (Together with this.state.schema). 

The idea is that SidePanel should only re-render the form when there is a new activity/operator, and EnhancedForm should only re-render the form when its a new Activity/Operator, or the schema has changed (due to hiding etc). This should make typing much more stable. We are not worried about synchronizing state with other users live anyway (if we were, we'd have to use Collaborative Form). 

I also fixed a bug in graph copy (api/graphs), where it saw that an operator had an activity selection field (although it was hidden), and tried to look up the activity id. 

I messed around for quite a while trying different things while trying to solve this - it's possible that with my latest change, some of the code is obsolete, however it seems to work well now - any change requires testing. A useful thing is to put a console.count in the render statements of both EnhancedForm and ConfigForm, to make sure they are not re-rendering when you don't want them to.

### DIFF
--- a/frog/imports/ui/GraphEditor/SidePanel/ConfigForm.jsx
+++ b/frog/imports/ui/GraphEditor/SidePanel/ConfigForm.jsx
@@ -27,6 +27,14 @@ export default class ConfigForm extends Component {
     }
   }
 
+  shouldComponentUpdate(nextProps: PropT): boolean {
+    if (this.props.node._id === nextProps.node._id) {
+      return false;
+    } else {
+      return true;
+    }
+  }
+
   render() {
     const {
       node,
@@ -48,7 +56,6 @@ export default class ConfigForm extends Component {
         groupingKey: node.groupingKey
       },
       onChange: data => {
-        this.setState({ formData: data.formData });
         if (node.operatorType) {
           addOperator(node.operatorType, data.formData, node._id);
         } else {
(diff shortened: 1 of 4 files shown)


frog/imports/ui/GraphEditor/SidePanel/ConfigForm.jsx
```diff
@@ -27,6 +27,14 @@ export default class ConfigForm extends Component {
     }
   }
 
+  shouldComponentUpdate(nextProps: PropT): boolean {
+    if (this.props.node._id === nextProps.node._id) {
+      return false;
+    } else {
+      return true;
+    }
```
lfaucon, line 35: I can see you are a beginner programmer.
```
return !(this.props.node._id === nextProps.node._id)
```

lfaucon (approved):

PR #664, opened 2017-11-12
Unil w7

lfaucon:


frog/imports/ui/GraphEditor/store/store.js
```diff
@@ -171,10 +171,12 @@ export default class Store {
           !source ||
           !target
         ) {
-          throw 'Cannot find connection source/target, or source/target is a connection';
+          return null;
```
houshuang: When did this case happen, and could it be prevent somewhere else instead of here?

houshuang:
@lfaucon Do you want to try to get this merged to develop? It looks fine to me - the only thing I am wondering about is the random assignment of students to experiment/feedback, I guess that is hard-coded in the operator right now?

lfaucon:
This should be ready to be merged now. The changes are:
- increase font size of text learning component in ac-classifier
- disallow completing a quiz which does not have answer to every questions
- export data in ac-textarea (yours)
- add guidelines before the prompts in ac-textarea
- fix wrong filter for finding a control operator in `frog/server/main.js`

houshuang (approved):

PR #665, opened 2017-11-13
React-Ace editor for ac-autocode

Keylas:
Use of Skulpt to create a python console.


ac/ac-autocode/src/ActivityRunner.jsx
```diff
@@ -1,30 +1,91 @@
...
+    this.state = {
+      inputCode: 'print "Hello World"',
+      outputFeed: "You'll see the output of your code here"
+    };
+
+    this.handleChange = this.handleChange.bind(this);
+    this.runit = this.runit.bind(this);
+  }
+
+  componentDidMount() {
```
Keylas: Using Skulpt require to execute the two following scripts, which create the object 'Sk'.
The current solution calls the scripts from the Skulpt distribution site and has them included on the page.

ac/ac-autocode/src/ActivityRunner.jsx
```diff
@@ -1,30 +1,91 @@
...
+    const script = document.createElement('script');
```
Keylas: Using Skulpt require to execute the two following scripts, which create the object 'Sk'.
The current solution calls the scripts from the Skulpt distribution site and has them included on the page.
The object will then be referred by window.Sk

ac/ac-autocode/src/ActivityRunner.jsx
```diff
@@ -1,30 +1,91 @@
...
+
+  outfunction(text: string) {
+    window.Sk.runner.state.outputFeed =
+      window.Sk.runner.state.outputFeed + '\n' + text;
+  }
+
+  runit: Function;
+  runit() {
+    if (window.Sk) {
+      window.Sk.runner = this;
```
Keylas, line 66: Outfunction has a different scope because it is called by Sk (so using this.state in outfunction does not work).
This workaround creates an explicit link between Sk and the ActivityRunner.

houshuang:
Why aren't you using Reactive Data to store the editor contents?

PR #684, opened 2017-11-20
induc logger

romainAA:
added log for the correction of the tests and tests with FB

### DIFF
--- a/ac/ac-induction/src/TestPart/Test.jsx
+++ b/ac/ac-induction/src/TestPart/Test.jsx
@@ -18,7 +18,8 @@ export default ({
   feedback,
   dataFn,
   data,
-  logger
+  logger,
+  groupingValue
 }: Object) => {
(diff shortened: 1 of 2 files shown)


ac/ac-induction/src/TestPart/Test.jsx
```diff
@@ -45,9 +46,14 @@ export default ({
       feedback ? 'listIndexTestWithFeedback' : 'listIndexTest'
     );
     if (feedback) {
+      logger({
+        type: 'TestFB:' + groupingValue,
```
lfaucon: The grouping value is already part of the log automatically. It does not make sense to include it in the type. We want to use type in the logs for possible actions: "looking at an example", "submitting an answer", "going to the next phase".

lfaucon (changes requested):

lfaucon (approved):

PR #696, opened 2017-11-22
Randomize quiz

### LINKED ISSUE #692: Randomize options in ac-quiz

houshuang:
Adding option to shuffle both order of questions and order of answer alternatives for each question in ac-quiz. This is to prevent copying by looking at someone else's screen. Internally question IDs and answer alternative IDs are kept consistent, so submitted answers are comparative.

Uses seededShuffle package which will always shuffle an array in the same way for a given student ID, this is important, otherwise the answers would be reorganized for each re-render.

screen shot 2017-11-22 at 06 47 55

Closes #692

### DIFF
--- a/ac/ac-quiz/package.json
+++ b/ac/ac-quiz/package.json
@@ -16,7 +16,8 @@
     "react-dom": "^16.0.0",
     "react-jsonschema-form": "^1.0.0",
     "react-latex": "^1.0.1",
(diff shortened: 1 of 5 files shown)


yandim:
Interesting new option... Of course, its use depends on the pedagogical intention and the specific setting.

lfaucon (dismissed):

PR #705, opened 2017-11-23
Adding quiz answers, operator, formatProduct

### LINKED ISSUE #542: Export function in activity packages
Add export function for reactive data, to be called when merging the reactive data into a product.

### LINKED ISSUE #682: Add option to provide answers in ac-quiz
Add option to provide correct answer when configuring a quiz, and to show this to students when they have submitted (you got so many correct, etc).

houshuang:
This PR adds a bunch of things that all work together - happy to discuss if you don't like how I designed it.

- we've talked about having a way for an activity to format the reactive data before it gets turned into product. I introduce a formatProduct function, which is run once per instance, and receives config and the reactive data from an instance
- I introduce the option of marking which answer alternatives are correct in ac-quiz. Note that multiple answers can be correct - this is not checked. (We should perhaps enforce that at least one alternative is marked correct per question?)
- I use this to take the internal data from quiz, and make it much easier to deal with for other activities, with one array for questions, one for answers (with the full answer text, not 0/1/2 etc), one for [true , false] whether answers are correct or not, and finally a count of correct answers, and a count of total answers
- I introduce a new control operator which splits the class in two based on an incoming quiz answer (this could come from different activities, we can try to standardize the score etc to make these kinds of operators more general). We could have arbitrary number of splits/conditions etc, but to make it easy, I split into high/low performance, and you can provide a number of correct answers, or a percentage of the same, to split, and one activity for high and one for low activity students. 

I had to clean up a few small things around control operators, copying graphs etc to make things work. 

I also had to change the internal data structure of ac-quiz, which means all old graphs will be invalid :) But I don't see any way around this. 

Currently op-performance-select only takes p1 data as input, we could add p2 as well. I don't remember exactly how we wanted to deal with this, I think the idea was that the control operator would take the social structure, and the output would always be mapped to p1 - but in the type definition we have structure: 'individual" etc, so not too sure. 

To test, you can check out the graph European capitals from frog-graphs. 

You can try it out in Preview, including a new example I added with correct answers indicated.

Closes #682 
Closes #542


ac/ac-quiz/src/index.js
```diff
@@ -31,12 +31,38 @@ const exportData = (configData, { payload }) => {
...
   id: 'ac-quiz',
   type: 'react-component',
   meta,
   config,
+  configUI,
   ActivityRunner,
   dashboard,
-  exportData
+  exportData,
+  formatProduct
```
lfaucon, line 67: I don't really like the name you chose, and I think the naming is important here as it concerns the API. Did you have other ideas for names? What do you think of `makeProduct` ?

lfaucon: Overall I think it would be beneficial to spend some time renaming a few things around, starting maybe from what confused Yannis the most.

lfaucon (approved):

ac/ac-quiz/src/index.js
houshuang: I agree - I have thought for a while that we should rethink a bit the entire ActivityPackage API, since we keep adding new stuff to it. I'll merge this for now, but let's find some time to think about this.

PR #706, opened 2017-11-27
Make file upload work when using external IP

houshuang:
Previous development mode had hard-coded localhost, which means that if you run FROG in development but expose a public IP/DNS, it would fail. Made it more flexible.


lfaucon (approved):
I tested it and it works well, you can look at my last commit that uses correctly the upload function in the graph side panel and in ac-uploader.

PR #711, opened 2017-11-30
Upgrading React (16.2), and all other dependencies

houshuang:
Should be green, and have checked with all the preview activities. Also some minor Prettier changes, and more strict Eslint, which I turned off in eslintrc :) At some point we should look at accessibility of FROG.

Example of Fragment usage here: https://github.com/chili-epfl/FROG/pull/711/commits/edff77ba4df05471a9898ab6dffb58af20751442 - a bit awkward still, before Meteor upgrades to Babel 7 etc, and we can use <> </>


lfaucon (dismissed):
You forgot to change the react version in the activity template

PR #712, opened 2017-12-01
Updating form data when switching node

houshuang:
Currently EnhancedForm also takes an optional id prop, which is used to determine whether a new object is displayed, and to decide to redraw the form. This is used by ConfigForm to update formData when a new node is selected.


lfaucon (approved):
Thanks!

PR #729, opened 2017-12-08
Embedding H5P activities in FROG

### LINKED ISSUE #736: Activity connections in graph editor don't update in real-time
If I have an operator, for example op-control-group, which has activity fields, and the operator is not connected to any activities, it will show red. If I then connect to activities, it won't update the editor until I close and open the sidebar or switch operator and switch back again (force a recalculation).

houshuang:
Leaving this here, needs a bit of clean up.

Note that I introduce both the concept of "native" activityTypes (defined within frog/frog), and of dashboard components (not using react-jsonschema-form). Both were surprisingly easy to add, but the exact APIs etc probably need some discussion.

We also need to think a bit about file storage, currently it's using /tmp when not using Minio, but tmp might not survive restart for example, making it not safe for a FROG in production that doesn't want to use an external storage server. (Applies to ac-image, uploadFn, as well).

One interesting future direction could be to try to capture some of the log data into product, so that for example quiz results could be usable by operators etc - the problem is that all the activities are different, and for example video might have multiple quizzes etc.

Note also that this builds heavily on h5p-standalone (https://github.com/tunapanda/h5p-standalone), which I copied directly into frog/public/h5p, because I am also modifying some of the files etc. That's why so many lines added.

https://www.youtube.com/watch?v=_8YhWEkeG48&feature=youtu.be

Also fixes some bugs in the config form / side panel, like 736.

Closes #736


houshuang:
@lfaucon I think this is ready to merge. I reworked a bit the ConfigComponent thing, currently we declare it as part of the main config schema, under component: , which means that we can also specify required fields, validateConfig functions etc, and have this checked at runtime. I had to do a bit of trickery with the ConfigForm to make sure that it is updated when ConfigComponent updates, and doesn't overwrite etc, but it should work well now.

Also added more checks on the upload to make sure the h5p file uploads correctly and is valid.

Assuming this doesn't break anything else, it would be great to get it in, since some students might want to use it etc. I think the only controversial thing I'm doing is ConfigComponent, let me know if you want me to change anything about the API here.

lfaucon:
I get several errors when testing it.

- In preview because there are no example provided. Maybe we want to handle that case without crashing
- When opening the side panel on and ac-h5p
```
Uncaught TypeError: Cannot read property 'map' of undefined
    at RawEditActivity (EditActivity.js:146)
```

frog-utils/src/types.js
```diff
@@ -135,7 +135,8 @@ export type ActivityPackageT = {
     initData: any
   },
   exportData?: (config: Object, product: activityDataT) => string,
-  formatProduct?: (config: Object, item: any) => any
+  formatProduct?: (config: Object, item: any) => any,
+  ConfigComponent?: ReactComponent<any>
```
lfaucon: It makes sense to me that this component has the props `configData` and `setConfigData`. Might be useful to add it to the type definition

houshuang:
Thanks for reviewing this.

I don't think it makes sense to include exampleData for this, since it would mean including a full h5p package as part of FROG source code. However, I added a filter, which only lists activities with exampleData in the Preview view (the upcoming webRtc activity is another one which will not have a nice preview).

I also fixed the map thing, added the type definition, and also fixed a tiny bug where showing the logs in preview within the graph editor wouldn't work.

lfaucon (approved):

PR #734, opened 2017-12-11
Cs411 bkt

Keylas:


ac/ac-autocode/src/config.js
```diff
@@ -93,11 +93,30 @@ export const exampleConfig1 = {
...
+      {
+        description: 'Edge case 1',
+        preCode: 'a=[42]',
+        postCode: 'print b'
+      },
+      {
+        description: 'Edge case 2',
+        preCode: 'a=[]',
+        postCode: 'print b'
+      }
```
lfaucon: Can you use the configUI widget textarea so we can write several lines here for preCode and postCode just like for template and solution?

screen shot 2017-12-12 at 08 51 58

ac/ac-autocode/src/config.js
```diff
@@ -93,11 +93,30 @@ export const exampleConfig1 = {
...
+    templateCode: 'b=a',
+    solution: 'b=sorted(a)',
+    tests: [
+      {
+        description: 'Already sorted',
+        preCode: 'a=[1, 2, 3]',
+        postCode: 'print b'
+      },
+      {
+        description: 'Normal case',
```
lfaucon: Seems to me that you don't use that description anywhere. So if you don't think it should be used, then don't make the configuration more complex with it.

ac/ac-autocode/src/ActivityRunner/TestPanel.jsx
```diff
@@ -54,9 +54,10 @@ const Feedback = ({ input, expected, received, error, stdout, debug }) => (
 const TestList = ({ tests, ...props }) => (
   <ButtonList>
     <Debug key="debug" {...props} />
-    {tests.map((test, index) => (
-      <Test key={test + index} test={test} index={index} {...props} />
-    ))}
+    {tests.map((test, index) => {
+      test.id = index;
+      return <Test key={test.id} test={test} index={index} {...props} />;
```
lfaucon: Nice trick..., but isn't it nicer to just add a tag to ignore the eslint error here? Think about someone else going to read and work on your code in a few months, what would be the easiest to understand?

ac/ac-autocode/src/ActivityRunner/Header.jsx
```diff
@@ -6,5 +6,6 @@ export default ({ config, style }: Object) => (
   <div style={style}>
     <h2>{config.title || 'Coding activity'}</h2>
     <h3>{config.guidelines || 'No guidelines were given for this exercice'}</h3>
+    <h3>{config.specifications || ''}</h3>
```
lfaucon: Maybe `<p>` instead of `<h3>`?

If I understand correctly guidelines would be: "Write an algorithm to sort a list." and specifications would be: "There will be an input `a=[3,1,2]` (for example) and you should write the sorted list in a variable named `b`".  So guidelines are about the goal of the exercises while specifications are about the details of the implementation. Is that correct?

ac/ac-autocode/src/ActivityRunner/TestPanel.jsx
```diff
@@ -0,0 +1,97 @@
...
+  align-items: stretch;
+  margin-left: 10px;
+  margin-right: 10px;
+`;
+
+const Feedback = ({ input, expected, received, error, stdout, debug }) => (
+  <div>
+    {[
+      input && <div key="inputT">INPUT</div>,
+      input && <CodeBox key="inputB">{renderLines(input)}</CodeBox>,
```
lfaucon: The input looks weird on the three examples:
```
a=[10, 2, 15, 8, 7]
print b
```
or 
```
print roman_num(50)
```
I think I'd just like the input shown to the students to be of the form 
```
a=[10, 2, 15, 8, 7]
```
or
```
num=50
```
So it could be only the preCode, then if you are afraid of students changing the variable that you created in the preCode for the postCode, we could just set it again in the post code, or I don't mind having:
```
num=50
```
then
```
print roman_num(50)
```

lfaucon (changes requested):

lfaucon (approved):

PR #740, opened 2017-12-14
Experimenting with Cypress for E2E front-end testing

houshuang:
This is pretty cool, Cypress automatically loads every single example in all activities (not hard-coded), in a real browser, and checks that none of them create errors etc. See video here:

https://youtu.be/CWLT-WA0X8U


lfaucon (dismissed):
It's great but I have a few problems with it:

- Initial setup takes much longer as it downloads and unzips Cypress every time. Should we have a setup for frog only without dev dependencies?

- The initial_setup ends with this error, but I don't see what has changed that would cause it.
```
mkdir: frog/node_modules: File exists
ln: frog/node_modules//@babel: File exists
ln: frog/node_modules//@cypress: File exists
ln: frog/node_modules//@houshuang: File exists
ln: frog/node_modules//@types: File exists
ln: frog/node_modules//Base64: File exists
ln: frog/node_modules//abab: File exists
ln: frog/node_modules//abbrev: File exists
ln: frog/node_modules//ac-autocode: File exists
```

houshuang:
A setup without dev dependencies would be complicated, since it would not be possible to even build FROG without babel etc... How to separate essential vs non-essential dev dependencies since it's all in the same package.json? However we could request that cypress be installed globally, and not include it in package.json. 

Part of me wonders about using something like Docker or Vagrant for development of FROG - since the core image wouldn't change much between versions. But I think for now it would introduce more complexity rather than less.

houshuang:
Suggestion: We remove it from package.json, and have a test script which works a bit like initial_setup with yarn - checks if the correct version of Cypress is installed globally (to save time) and if not, installs the correct version of Cypress locally and runs the tests. That way it should work quickly for core developers (and I can add a global install of Cypress early in the CI docker file as well), and it would still work (but more slowly) for occasional developers...?

lfaucon:
Sounds good to me.

houshuang:
@lfaucon I would like to get this merged to FROG. I have changed the package so that it does not run Cypress tests on CI or install anything extra - but if you have Cypress installed globally, you can easily run the tests - still useful. Hopefully in the future, we can figure out a way to integrate it with CI.

lfaucon (approved):
Very nice to have :)

PR #745, opened 2017-12-22
Ac autocode javascript

Keylas:
Add javascript support for activity acAutocode


.eslintrc.js
```diff
@@ -63,7 +63,8 @@ module.exports = {
     'react/require-default-props': 'off',
     'react/sort-comp': 'off',
     'jsx-a11y/mouse-events-have-key-events': 'off',
-    'global-require': 'off'
+    'global-require': 'off',
+    'no-eval': 'off'
```
Keylas: will remove the global disable of the 'no-eval' rule and add the eslint comment at the single line involved.

lfaucon:
Looks good.
Sometimes when I open the preview the template code does not appear in the editor. Do you also reproduce this? Do you know where it comes from?

lfaucon (approved):

Keylas:
I also have this issue upon opening the preview mode for the first time. Switching example then resetting reactive data give the template code back. My guess was that the editor starts empty and the reactive data is overridden by that before the template code is actually put in the editor.

PR #746, opened 2018-01-05
Material Next Teacher Views

aperritano:
Working version of material-ui for teacher views. 

All views have been converted except the graph and session view due to bug #738.


lfaucon:
@aperritano Thank you for sharing your work. It looks great.

When running it, I had this error:
```
W20180108-10:52:28.725(7)? (STDERR) ReferenceError: acCkStudent is not defined
...
```
Which I fixed by adding ac-ck-student to `frog/package.json` and importing it in `frog/imports/activityTypes.js`.

Also if I don't specify the `/teacher` in the url (while logged in) I get:
```
Uncaught TypeError: Cannot read property 'to' of undefined
    at new TopBar (TopBar.js:53)
...
```
The previous behaviour was to redirect to `/teacher` by default

There are a lot of things broken all around due to the removed css, but it looks like it is going in the right direction. We should discuss how we plan to fix and test everything. I think both Stian and I are looking forward being able to give the FROG UI to professors which we were not able to do for our first run.

houshuang:
@lfaucon Pasting some stuff from an email to Tony here, seems promising:

As far as I know, the two main concerns are the graph editor, and the react-jsonschema-form. For the graph editor, we should be able to figure out which definitions are missing and just add these directly to the code, for the RJF form, it appears in many places, but I can look at the documentation to see how to deal with it. I’m wondering if it’s possible to import bootstrap globally, but only trigger it for stuff with a certain selector - for example if you did `<div class=‘md-2’>` it would have no effect, but if you did `<div class=‘bootstrap’><div class=‘md-2’></div></div>`, then the rule in the inner div would trigger? 

This seems to indicate that I’m not the first one to think about this, and seems quite promising. https://stackoverflow.com/questions/11831346/applying-css-styles-only-to-certain-elements

It would be great if we could side-step the tricky issues for now and make sure that there is no global CSS that interferes with your activities / Material, but at the same time be able to just slap a `<class bootstrap>` around the graph editor and the forms, and have it render perfectly…

houshuang:
Closed in favor of #751

PR #749, opened 2018-01-08
Uses Modal setAppElement

lfaucon:


houshuang:
Can you explain better?

lfaucon:
Since a previous updates, we would receive a warning every time we opened a modal. It was not an error, but it is better to handle it properly. 

See [here](https://github.com/reactjs/react-modal#app-element) for the documentation

houshuang (approved):

PR #753, opened 2018-01-09
Ac autocode dashboard

Keylas:
Use of [Victory](https://formidable.com/open-source/victory/) graphs to provide the teacher with some information about the activity (namely, the current success rate of the students on the tests and their use of the debug functionality)


lfaucon (commented):
@Keylas I tried to reduce the number of lines of code for the dashboard to make it easier to update in the future. You had some sort of optimisations implemented which were questionable as they required you to store in the dashboard data the code of all the students. I removed it, but now there will be a few more calls to `dataFn`.

ac/ac-autocode/src/Dashboard.jsx
```diff
@@ -3,23 +3,152 @@
...
+const Viewer = (props: Object) => {
+  // props = {users, instances, data, config}
+  const chartData = value =>
+    props.config.tests.map((test, index) => ({
+      x: 'test ' + index,
+      y: props.data.tests[index] ? props.data.tests[index][value] : 0
+    }));
+
+  const debugData = () => {
+    const o = Object.keys(props.users).reduce((acc, userId) => {
```
lfaucon: The viewer receives as props the list of users and instances. I think the dashboards should be based on instances. Logs also have the instanceId to be used in the merge function

lfaucon (changes requested):

Keylas:
The reason I was logging students' code was in case it would happen to become useful at some point (say the teacher would want to "probe" some code that have errors to get an idea of why his students are having trouble). This indeed led to weird test cases. The only sort of optimization it allowed was to not take into account multiple successive clicks on the buttons but even then it was not perfect.

Keylas:
- changed userId to insanceId because it does make more sense
remark: in preview, instances goes [0,1] while the log.instanceId are 1 and 2 so the graph is not correct.
- took Stian's #755 issue into account. It actually makes sense to have an activity without tests so functionalities that use it now verify that props.config.tests is not undefined (most problems came from the fact that they assumed it to be [] initially whereas it seems to be undefined). Added a validateConfig to check that a non empty solution is provided when there are tests.

lfaucon (approved):

PR #754, opened 2018-01-10
Ac autocode javascript

Keylas:
Transition from using eval to calling a web worker for javascript.
MakeRunCode is split into one file per language.


lfaucon:
@Keylas could you branch from develop if you have more PRs so I can see only the list of commits that concerns your current changes.

Thank you for your work, this looks good 👍

lfaucon (approved):

PR #769, opened 2018-01-24
Headless FROG

houshuang:
This PR implements an API for exposing activities and configuration to external systems. The external API is [documented](https://github.com/chili-epfl/FROG/wiki/Headless-activity-API) in the wiki.

There have been quite a few changes to core functionality, to remove the assumption that a side panel will always be writing its configuration directly to the database for example. Some of this could be further cleaned up and simplified in the future. 

@lfaucon There is a single Flow error left, which I cannot understand, and which annoys me. Can you have a quick look?


frog/imports/ui/App/index.js
```diff
@@ -205,15 +207,77 @@ const FROGRouter = withRouter(
...
+    InjectData.getData('api', data => {
+      this.setState({ mode: 'ready', api: !!data, data });
+    });
+  };
+
+  render() {
+    if (this.state.mode === 'waiting') {
+      return null;
+    } else if (this.state.api && this.state.data) {
+      if (this.state.data.callType === 'config') {
```
lfaucon: When I visit `.../api/config/ac-autocode` I wish there was the same button to copy the standalone link as in the new side-panel in the graph editor

frog/imports/ui/App/index.js
```diff
@@ -205,15 +207,77 @@ const FROGRouter = withRouter(
...
+            username={data.username || 'Anonymous'}
+            userid={data.userid || '1'}
+            stream={() => {}}
+            reactiveId={data.instance_id}
+            groupingValue={data.instance_id}
+            activityData={{
+              data: data.activity_data,
+              config: data.config || {}
+            }}
+          />
```
lfaucon: I think we would benefit in the future from moving the logic about routing for the API in its own file `App/ApiRouter.js`. This file is already overloaded and we don't really know anymore everything that is happening in it.

lfaucon (approved):

PR #776, opened 2018-01-26
Graph firefox

romainAA:
So this seems to fix the problem but I don't quite understand why you used `e.nativeEvent.offsetX` instead of `e.clientX` in the first place (same for Y).


houshuang (approved):

houshuang:
Great work! Thanks.

PR #797, opened 2018-02-09
Ac stroop

lfaucon:


lfaucon:
@houshuang I updated the templates which were outdated since the recent changes, please check if they if they are good now.

houshuang (approved):

PR #798, opened 2018-02-09
Cleaning up unneeded files

houshuang:
./run_and_watch_all.sh is not needed anymore, and I took the time to remove several other files which are not used anymore as well.


lfaucon (approved):
My favorite kind of PR

PR #801, opened 2018-02-11
Improvements in ac-stroop

lfaucon:
@houshuang You merged my previous PR quite prematurily (CI was red and it had the "not ready" tag). This version should be much better. Let's try it out Tomorrow with Jenny also to discuss the last things we need to implement.


ac/ac-stroop/src/ActivityRunner.jsx
```diff
@@ -42,80 +49,136 @@ const generateExample = (objects, colors) => {
...
-  const objectList = objects.split(',')
+let noAnswerTimeout;
 
-  const nextQuestion = () => {
-    setQuestion(generateExample(objectList, colorList))
-  }
+const Question = props => {
+  const { setQuestion, question, progress, setProgress } = props;
+  const { objectName, colorName, colorFill } = question;
+  const { logger } = props;
```
houshuang: why not on line 98? :)

ac/ac-stroop/src/ActivityRunner.jsx
```diff
@@ -42,80 +49,136 @@ const generateExample = (objects, colors) => {
...
+    logger({ type: 'progress', value: progress + 1 });
+    setProgress(progress + 1);
+    setQuestion('waiting');
+  };
+
+  Mousetrap.bind('y', onClick(true));
+  Mousetrap.bind('n', onClick(false));
+
+  clearTimeout(noAnswerTimeout);
+  noAnswerTimeout = setTimeout(onClick(undefined), 10000);
```
houshuang, line 116: Shouldn't this be setTimeout(() => onClick(undefined))?

houshuang: Ah ok, onClick returns a function. C'est tout bon :)

ac/ac-stroop/src/Dashboard.jsx
```diff
@@ -1,29 +1,35 @@
...
+      answerTime
+    } = log.payload;
+    const qType = isConsistent ? 'consistent' : 'inconsistent';
+    if (isCorrect === answer) {
+      dataFn.numIncr(1, [qType, 'correct', 'count']);
+      dataFn.numIncr(answerTime - startTime, [qType, 'correct', 'time']);
     } else {
-      dataFn.numIncr(1, [stroopType, 'wrong', 'count'])
+      dataFn.numIncr(1, [qType, 'wrong', 'count']);
+      dataFn.numIncr(answerTime - startTime, [qType, 'wrong', 'time']);
```
houshuang: I don't understand this - you only store the total wrong time and total right time for all questions? Is this what the dashboard needs?

houshuang (changes requested):

houshuang:
You're not storing current state in reactive data, which means that on reload you restart the exercise?

ac/ac-stroop/src/Dashboard.jsx
lfaucon: I store the sum and the count of all the time spent on questions for each of the 4 category (inconsitent/consistent x correct/wrong). That allows to compute the averages we need for the dashboard.

jkolsen: In the current dashboards, they are not distinguishing the correct/wrong but just has graphs for the inconsistent/consistent and checking to see if the student correctly marked it as wrong or not.

lfaucon:
Should be good to go now

PR #806, opened 2018-02-12
Reuse leaderboard

lfaucon:
replaces #803 
comes after #804


houshuang (approved):

houshuang:
Looks OK to me. Don't we need all the CSS you removed?

lfaucon:
The CSS was for the monty-table component and messes up with Prastut's dashboard

PR #808, opened 2018-02-13
Ac dash

lfaucon:


lfaucon:
DeepScan error does not make sense

lfaucon:
I fixed the issue coming from the merge. This should work now. Some of the props of the dashboard like the list of students might not work properly however

Issue #772, opened 2018-01-26
Activity library


lfaucon:
- A new tab where teacher can browse and create new activities: 
    - From scratch by choosing the type
    - Copying and editing an existing one
- In the graph editor side panel allow to create from scratch or copy existing
- In the graph editor side panel allow to save current activity to the library

Details here: https://docs.google.com/document/d/1tduj6JEzFRl4wCdSapCpdKnk_HF9UeRGmaOv6oeTkjo/edit

PR #809, opened 2018-02-13
Act lib

romainAA:
Add a library activity, this parent id saved in the library DB, but is not really used
#772


frog/imports/api/activities.js
```diff
@@ -20,7 +20,7 @@ export const addActivity = (
   groupingKey: ?string
 ) => {
   if (id) {
-    const toSet = omitBy({ data, groupingKey }, isNil);
+    const toSet = omitBy({ activityType, data, groupingKey }, isNil);
```
houshuang: Why this?

houshuang (changes requested):
This is a great beginning!

Note that the parentId relationship should go the other way - always point towards an activity in the library (when you export an activity from a graph, the in-graph activity should be updated with the parentId of the activity in the library, and when you create a new in-graph activity from library, it should also have parentId pointing back to library).

Currently preview button is broken.

It would be better to call it activity library, at least in code, since we might have an operator library or other library in the future.

frog/imports/api/activities.js
romainAA: It makes things a bit easier, in the chooseActivity select function, the DB is directly modified through `Activities.update(this.props.activity._id, $set: { activityType: activityType.id }});` since the add activity previously could not change the type of an activity. But I think it makes more sense to use the `addActivity` function with an existing id to modify it, no ?

houshuang:
Can you ping me when you want me to look at this again @romainAA ?

romainAA:
@houshuang should be good

frog/imports/ui/GraphEditor/SidePanel/ModalExport.jsx
```diff
@@ -0,0 +1,94 @@
...
+        <Button
+          className="btn btn-danger"
+          onClick={() => this.props.setModal(false)}
+        >
+          Cancel
+        </Button>
+        <Button
+          className="btn btn-primary"
+          onClick={() => {
+            addActivityToLibrary(
```
houshuang: Almost perfect. Can you move this logic so that you here only call a function called exportActivity(activity) or exportActivity(activity._id), and the logic is all in the api/activity_library?

Also please update the parentId of the activity you are exporting to point to the newly exported activity. Then we're done :) 👍

PR #811, opened 2018-02-13
induction exampleData

romainAA:
fixed a problem with fb modal and added the full set example to the exampleData


ac/ac-induction/src/TestPart/ModalSubmit.jsx
```diff
@@ -12,116 +12,129 @@ import {
...
-  ][data.indexCurrent].correction || {
-    result: 2,
-    show: 0,
-    propertiesIndex: []
-  };
-  return (
-    <Modal isOpen={data.feedbackOpen} contentLabel="Modal">
-      <div style={{ display: 'flex', flexDirection: 'row' }}>
-        <div>
+class ModalSubmit extends Component<Object> {
```
houshuang, line 15: Can you move this to componentDidMount? You should not do any DOM related stuff in constructor (only setstate etc)

houshuang: Actually you don't need a constructor here anyway

ac/ac-induction/src/index.js
```diff
@@ -29,17 +29,17 @@ const meta = {
     {
```
houshuang: can you put the exampleData in a separate file and import it?

romainAA:
@houshuang done

PR #817, opened 2018-02-14
Add dashboards for ac-stroop

lfaucon:


jkolsen:
The timing dashboard is currently only checking the time when an answer is correct. I think it also needs to account for wrong answers.

jkolsen (changes requested):
The graph for time currently only shows time for correctly answered questions. I think it should also display for incorrectly answered questions.

jkolsen (approved):

PR #819, opened 2018-02-14
replaced createContainer

romainAA:
as createContainer was deprecated, I did this small PR to change it to withTracker. I didn't make a lot of testing, but everything seems to work.


houshuang:
@romainAA Lot's of errors.

PR #830, opened 2018-02-15
remove from library

romainAA:
add the possibility to remove an activity from the library with a confirmation modal

### DIFF
--- a/frog/imports/ui/GraphEditor/SidePanel/ModalDelete.jsx
+++ b/frog/imports/ui/GraphEditor/SidePanel/ModalDelete.jsx
@@ -1,16 +1,9 @@
-import React, { Component } from 'react';
+import React from 'react';
 import Modal from 'react-modal';
 import { Button } from 'react-bootstrap';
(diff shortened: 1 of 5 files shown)


frog/imports/ui/GraphEditor/SidePanel/ModalDelete.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from 'react';
+import Modal from 'react-modal';
+import { Button } from 'react-bootstrap';
+
+class DeleteModal extends Component<Object> {
+  componentDidMount() {
```
houshuang: You shouldn't need this after my #829

houshuang:
There is a small CSS issue
screen shot 2018-02-15 at 12 14 31

Otherwise, seems to work very well

romainAA:
@houshuang done.
I also required the title and description fields of the export to be non-empty. I think it makes sense, if not I'll remove it

PR #831, opened 2018-02-15
Disable keys in ac-stroop during waiting

houshuang:
Fixes bug that let students press y/n even while waiting for a new question


lfaucon (approved):
Cool, that's exactly the same fix I made in #833

PR #834, opened 2018-02-15
 Allow display of dashboard of all activities

lfaucon:


houshuang:
If you click on an activity that has not happened yet, it spins, but it doesn't actually crash - seems still fully functional, and once the activity starts, it works, so I guess it's fine. Not sure if it should automatically switch to a new dashboard when you click next.activity from dashboard view - but I guess we'll re-do how we do dashboards anyway so for now it seems good.

Funny how we talked about this forever and it's such a small change :)

PR #836, opened 2018-02-16
Allow leaderboard to receive several score values

lfaucon:


houshuang:
I see you added a new column on the leaderboard, is this combined time + accuracy? Does it need a header?

lfaucon:
There is no combined time + accuracy. The second column is the time taken (when used by ac-stroop, otherwise it is whatever the activity type logs as second score). It is only used when there is an equality on the primary score (number of correct answers for ac-stroop).

PR #840, opened 2018-02-16
Hardcode questions for ac-stroop

lfaucon:
Please make sure that no mistake are included, especially for the isCorrect & isConsistent tags


jkolsen (approved):
Questions look correct.

PR #844, opened 2018-02-19
Stop progress graph once the activity is closed

jkolsen:
Once the activity is closed, no longer update the activity.


frog-utils/src/dashboards/progress.js
```diff
@@ -50,24 +50,27 @@ export const LineChart = ({
 );
 
 const TIMEWINDOW = 5;
+let timingData;
```
lfaucon: - This is used inside the Viewer component, but defined outside. Not a good practice. 

- If the Viewer component is called for the first time after the activity has ended (for example if the page is refreshed) timingData will be undefined.

frog-utils/src/dashboards/progress.js
```diff
@@ -50,24 +50,27 @@ export const LineChart = ({
...
 
-  const numWindow = Math.ceil(
-    (timeNow - activity.actualStartingTime) / 1000 / TIMEWINDOW
-  );
-  const timingData = [[0, 0, 0]];
-  const factor = 100 / Object.keys(instances).length;
-  for (let i = 0, j = -1; i <= numWindow; i += 1) {
-    if (i * TIMEWINDOW === (data['timing'][j + 1] || [0])[0]) {
-      j += 1;
+  if (activity.actualClosingTime === undefined) {
```
lfaucon: I think a good way to do is would be to do:

```

const ViewerWithTime = ({ data, instances, activity, timeNow }) => {
    ...
    return ( ... )
}

const Viewer = (props) =>
    props.activity.actualClosingTime 
        ? <ViewerWithTime {...props} timeNow={activity.actualClosingTime} />
        : <TimedComponent(ViewerWithTime, 5000) {...props} /> 
```

lfaucon (changes requested):

frog-utils/src/dashboards/progress.js
jkolsen: I changed the suggestion for how to include the activityclosingtime

houshuang (approved):

PR #845, opened 2018-02-19
Add way to check for multiple time entries

jkolsen:
Added a catch all when timing has multiple entries for the same time.


lfaucon (approved):
Good :)

PR #847, opened 2018-02-19
bugfix babel crash on syntax error

### LINKED ISSUE #842: watch/watchAll should not crash on syntax error
Before we used to use a node package that just watched for file change, and then ran babel for the entire directory if a file changed. Currently, we are using the --watch function built-in to babel, to automatically rebuild files which are changed. This is more efficient (only one file is rebuilt, when changed).

However, when Babel finds a syntax error (quite often when you are editing), it crashes, and stops rebuilding. We obviously don't want this. 

Someone should look into what's a better way - is it to check if Babel processes die and then restart them? Or use the watch script, but run actions on individual files and not on whole directories?

As an added bonus, it would be great if we could run a single watch script for all of FROG instead of one per directory... I'm wondering if there is a node library which could just watch all of FROG for file changes, and then let us implement the logic (filter out node_modules, only rebuild single files, etc).

prastut:
Fixes #842

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -1,5 +1,5 @@
 const { sync } = require('find-up');
-const { dirname } = require('path');
+const { dirname, join } = require('path');
 const { readdirSync } = require('fs');
 
 let dir;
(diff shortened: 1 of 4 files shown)


lfaucon (changes requested):
Looks like you changed watchAll, but the problem will still occur with watch. Can you update your PR to fix watch too?

prastut:
@lfaucon good catch, please check now?

houshuang (dismissed):
Awesome!

prastut:
@houshuang this might break if the filename that babel encounters contains src. Fixing this.

PR #853, opened 2018-02-20
fix the close preview problem

### LINKED ISSUE #832: Activity library preview closing -> ...
When you do preview from the activity library in graph sidebar, and close the preview, the sidepanel shows the list of activity types, instead of back to the activity library

romainAA:
The only thing left is when you have 2 activities w/o type and you go from one to the other, the variable libraryOpen of the store stays the same since the ChooseActivity doesn't unmount.

close #832


lfaucon (changes requested):
I still find the same bug as before

lfaucon (dismissed):

lfaucon (approved):

Issue #733, opened 2017-12-11
Undo in graph editor broken


houshuang:
In the current develop, undo seems completely broken, activities loose their names, activity types etc. Probably something with how we store the undo data, or merge it back (would be easier if graph was a single entity).

PR #856, opened 2018-02-21
Fix undo

romainAA:


romainAA:
#733

lfaucon (changes requested):
The names don't disappear anymore! That's great, testing it, I lose all the activityTypes and configuration when I click undo. There is certainly more fixing and cleaning that we can do here

PR #864, opened 2018-02-27
Default configuration

### LINKED ISSUE #764: [GraphEditor] Default config params not applied if AC is not edited
How to reproduce: Add an ac-webrtc activity to a graph, without having sidebar open or making any edits. Graph is green, but when running, it will fail, because default config was not written to database. 

In this case, of course ac-webrtc could also be more robust, but why should we need to specify default settings in two places? We should look at where default settings are merged, and ensure that they are set even if an activity config is never edited.

### LINKED ISSUE #858: Pre-fill export field with activity name
When exporting an activity from sidepanel, pre-fill the title field with the title of the activity.

romainAA:
Closes #764 
Closes #858


romainAA:
The last push is here because I forgot to switch branch before I pushed, but it has nothing to do with it: it fixes #858

lfaucon (dismissed):
This is quite good, but it won't work if the config is nested
```
config = {
    type: 'object'
    properties: {
        myArrayField: {
            type: 'array'
            items: {
                type: 'string'
                default: 'hello world'
            }
        }
    }
}
```

What we should look for I think is a way to force a trigger of onSubmit on the form. or programmatically get the formData

houshuang:
@lfaucon Fixed this using jsonschema defaults. Please review and merge.

lfaucon (approved):

PR #871, opened 2018-03-05
Dashboard preview timeslider

houshuang:
This is a first implementation of a time slider in the log preview. It works, even though it has some small quirks, but I think it will still be useful. In the future, I will use simple JS objects instead of ShareDB in-memory collection (with a shim over generateReactiveFn) which will speed up a lot, and also avoid reloading the dashboard component (so we will not loose internal state, like switching tabs), make it easier to slide back and forth (cache every nth item etc). Another idea is ability to autoplay, for example based on the timestamps.

Another added functionality is clicking the list icon next to the dashboard preview, which would usually show the reactive data, will now show the reactive data of the dashboard preview. (Note that this can be slow if you have hundreds of items in the structure).


lfaucon (approved):
To be optimized when we have time.

PR #881, opened 2018-03-07
Added missing features to watch

### LINKED ISSUE #874: Improve watch
Watch is currently limited to only change in files. 

- [x] Transpile all files on start. 
- [x] Pipe new files when added.

prastut:
Closes #874


houshuang:
I removed the ->src in the logger, I think this looks cleaner.

This seems to work well, however it takes a ton of CPU when it runs initially, spinning up lot's of node instances to recompile all the files. Sometimes this is useful, but not always - would it be possible to control this with an option? or have watchAll/buildAndWatchAll ? 

But we of course want to start watching newly added files - not sure if it's possible to separate this? (If I run watchAll right after initial_setup, I don't need to recompile everything, but want to watch new files)?

prastut:
Yup I can seperate this, will do it.

prastut:
@houshuang now you have: 
* `watchAll` which does not build all files, but watches `add|change` for new files. 
* `buildAndWatchAll` which builds all files and then does what `watchAll` does.

Issue #890, opened 2018-03-14
Mongo keys cannot have dots


lfaucon:
Mongo keys, even deep into a Mongo document structure, cannot have dots. This means we have to be very careful when using user-defined strings as keys.

Specific bug: The activityData from ac-show-social, when a name has dots.

More general: consider storing black-box data (config, product etc) as JSON.stringified. Downside is that it is harder to debug when looking in DB etc.

(this is Stian)

houshuang:
@lfaucon We know the concrete issue here. The rule of thumb should be to never use student generated content as a key in an Object. Do we also need to for example wrap product in a JSON.stringify before writing to Mongo to ensure that an activity cannot mess up for a whole graph? We would have to make sure that we unwrap it everywhere we use it, etc...

lfaucon:
Partially fixed by #891

PR #891, opened 2018-03-18
Changes required for running an arguegraph session

lfaucon:
## Changes
- ac-display-social:
    - Display the names of students sharing the same group. This allowed to discover the bug #890, we should do update the core to provide this automatically
- ac-quiz: 
    - Add one text area for justification with a dashboard that shows the justification of each instance.
    - Coordinates are added to the reactive data when clicking submit. 
    - Add button to go back to the quiz after submitting it.
    - Add noise to Scatter dashboard so that points do not overlap
- op-argue:
    - now takes as input the following:
```
activityData: {
  structure: 'all',
  payload: {
    all: { data: { instances, distanceMatrix  } }
  }
}
```
where for example
```
distanceMatrix = [
￼     [0, 0, 0, 5, 3],
￼     [0, 0, 0, 3, 5],
￼     [0, 0, 0, 3, 3],
￼     [5, 3, 3, 0, 0],
￼     [3, 5, 3, 0, 0]
￼  ]
```
- op-xy-distance:
    - takes as input data with coordinates and computes the matrix of distances

## Other changes from @houshuang 
- Add logout functionality & login with names and list of students


houshuang (approved):

houshuang:
@lfaucon I made some small changes, can you check them, and then we can merge? I didn't check your operators closely, I'm assuming you tested these thoroughly. I considered passing the names of the students in an instance to the activity runner, but this became complicated because of how we do the subscriptions etc, so I limited myself to fixing the bug in the way you do ac-social, and we'll keep a bigger change for the future.

lfaucon:
@houshuang I merged this PR now because I need it for working on the dashboards without creating conflicts. I'll do the optional logout after that.

PR #893, opened 2018-03-21
Change Progress Calculation

jkolsen:
The progress calculation is changed to just students that have started the activity and students are now marked as starting the activity when they hit the start button (or other specified starting point at the beginning of an activity).

### DIFF
--- a/ac/ac-stroop/src/ActivityRunner.jsx
+++ b/ac/ac-stroop/src/ActivityRunner.jsx
@@ -364,14 +364,22 @@ const Question = props => {
 };
 
 const Main = withState('question', 'setQuestion', null)(props => {
-  const { activityData, question, setQuestion, data, dataFn } = props;
+  const { activityData, question, setQuestion, data, dataFn, logger } = props;
   const { maxQuestions, delay } = activityData.config;
   const lang = data.language;
   const { name } = props.userInfo;
   if (!lang) {
     return <Form onSubmit={l => dataFn.objInsert(l, 'language')} name={name} />;
   } else if (question === null) {
-    const start = () => setQuestion('waiting');
+    const start = () => {
+      setQuestion('waiting');
+      logger([
+        {
+          type: 'progress',
+          value: data.progress / activityData.config.questions.length
+        }
+      ]);
+    };
     const { guidelines } = activityData.config[lang];
     return <Guidelines start={start} guidelines={guidelines} lang={lang} />;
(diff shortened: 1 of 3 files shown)


ac/ac-timedQuiz/src/ActivityRunner.jsx
```diff
@@ -164,13 +164,21 @@ const Question = props => {
...
     shuffledQ = shuffledQuestions(props);
-    const start = () => setQuestion('waiting');
+    const start = () => {
+      setQuestion('waiting');
+      logger([
+        {
+          type: 'progress',
+          value: data.progress / activityData.config.questions.length
+        }
+      ]);
```
lfaucon, line 180: You created a new 'progress' log that pops up for each question, making it two logs per question. As we have already too many logs, I'd rather use an other solution. As your only goal here is to have at least one log per instance, why not use the `activityDidMount` log common to all instances and write `progress = 0` when this log is received?

lfaucon (changes requested):

ac/ac-timedQuiz/src/ActivityRunner.jsx
jkolsen: I don't want to put it on the activity mounting because that can happen before the student actually starts the activity since they can sit on the start screen for a long time. I wanted to get an actual accurate measurement of the time they took on the task, which is why I put it on the start button.

lfaucon (approved):

ac/ac-timedQuiz/src/ActivityRunner.jsx
lfaucon: OK! I misunderstood the function you updated, I think it's good like this now.

ac/ac-stroop/src/ActivityRunner.jsx
```diff
@@ -364,14 +364,22 @@ const Question = props => {
...
   if (!lang) {
     return <Form onSubmit={l => dataFn.objInsert(l, 'language')} name={name} />;
   } else if (question === null) {
-    const start = () => setQuestion('waiting');
+    const start = () => {
+      setQuestion('waiting');
+      logger([
+        {
+          type: 'progress',
+          value: data.progress / activityData.config.questions.length
```
lfaucon, line 379: @jkolsen This line creates a bug. ac-stroop does not have a configured list of questions

PR #894, opened 2018-03-22
Multi dash

lfaucon:
## API change
Instead of exporting `{Viewer, mergeLog, initData}`, activities now export `{[name: string]: {Viewer, mergeLog, initData}}`

## Things not yet working:
- [x] api
- [x] dashboards in preview
- [x] preview with example logs 
- [x] updating ac-dash to be configurable with the name of the dashboard to display


lfaucon:
@houshuang I'll wait for your PR modifying the preview before updating the preview of dashboards

houshuang:
The pull-down menu looks weird, looks like the dashboard comes "in-between":

screen shot 2018-03-22 at 15 52 31

In general, I think I would prefer something like what we had before, with the various activities in a vertical list, and then perhaps the various dashboards in tabs, exactly like what it looked like (except now the tabs would be generated by the system, and not be baked into a dashboard)... It would make it much faster to switch, instead of first clicking for the pull down, and then selecting... Should also automatically open the first available one, etc.

lfaucon:
@houshuang This is ready for review. More changes than I anticipated, but all the code changes should be easy to understand I believe.

PR #896, opened 2018-03-22
Improving Preview

### LINKED ISSUE #605: Make preview list of activities look nicer
Using the component from graph editor (list/search/description etc)

houshuang:
This PR uses the standard activity selector from the side panel on the preview page. To access the traditional preview, with stable URLs, click the preview button. If you "select" an activity type, you get a config panel, and the moment the config is valid, it will preview on the right. You can click on any of the examples to pre-fill the preview with the config, etc.

Currently there is no way of getting back to the list of the activity types if you have a config that is invalid (the X button is only shown when activity gets rendered at least once). You can just switch tabs for now to get back. 

It needs better integration with the activity library etc going forwards. 

Closes #605

### DIFF
--- a/frog/imports/ui/App/APICall.js
+++ b/frog/imports/ui/App/APICall.js
@@ -32,6 +32,7 @@ export default ({ data }) => {
       <ApiForm
         activityType={data.activityType}
         config={data.config}
+        hidePreview
         hideValidator={data.hideValidator}
       />
     );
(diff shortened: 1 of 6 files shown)


houshuang:
Fixed that it was not updating multiple activities.

lfaucon (changes requested):
I changed a bit the UI so that we can't scroll away from the top bar and that the config and activity scroll independently. 

The mergeFunction that fills reactive data from config is not called anymore. You can test it with ac-image (gallery)

lfaucon (approved):

lfaucon:
I see that this still works when we open the preview from the "eye" button in the activityType list, so even though it needs better UI and a lot of code refactoring, I'll merge it.

PR #919, opened 2018-03-31
graph dropdown fix, now shows available graphs for session creation in session view.

aperritano:


houshuang:
I think this is based on a misunderstanding of how the graph/session relationship works currently - graphs that have a sessionId have been generated from an original graph, and cannot be used as the basis to start a new session. 

However, I am aware that the way we're doing it right now is a bit confusing, and @prastut is currently working up some prototypes that will be quite radically different - maybe moving the graph/session creation/management out of this screen, enabling editing of a running graph directly from within this screen etc. Prastut will share prototypes online so you can comment as well.

PR #920, opened 2018-03-31
Experimenting with Learning Items

houshuang:
Experimenting with the API and implementation for learning items. Currently does not work in Preview (we use a lot of ShareDB functionality), and is focused on ac-brainstorm. Early days.

Documented here: https://docs.google.com/document/d/1paNHdBbDh-jzcsGXeIE31t2ivolZ2DHFtfV0HwyzQR0/edit


houshuang:
Hi @lfaucon, now that our big mile stone has passed, I'd love to start discussing this. Before I do any more coding, I'd like to discuss the APIs etc. Don't worry about the merge conflicts etc, but if you check out this branch and git clean, it will work. Note that it only works in a session right now, not in Preview (because of ShareDB, should be fairly easy to fix).

So when you have time (not super urgent, but I'm eager to get started :)), have a look at the code, both how I've changed the different activities to support LIs, and how the core code in StudentRunner (should be moved) works, how the learning item types are implemented, how we should define external learning item type packages (currently all the LIs are internal, development was much faster like that).

One key decision I took is that a reference to a learning item is only an ID, and that the ActivityRunner which renders a learning item cannot know anything about which learning item it is, the metadata etc. Instead it calls <LearningItem id={} /> which uses ReactiveHOC internally (which works impressively well! but should be tested with hundreds of LIs on the screen etc). The key is that you can supply a render component, which then gets all the metadata and other props - you can see how this is working in Brainstorm for example.

One limitation of this is when you want to sort learning items by some metadata (like votes), this is currently not possible... One possibility would be to have let LearningItem take an array of IDs, and then render each of them using the render component... there might be other ways too (and maybe other limitations with this approach).

We should also think more about what metadata we store about different LIs... Since each activity has it's own reactive data, there's no problem to have a structure saying [{liId: 'a344', votes: 3}, {liId: '3434a', votes: 4}] etc. That would resolve the problem I mentioned above... Basically the question is whether the metadata should follow the LIs around or not. Currently I think I store the metadata inside the LIs in all the activities, but this might not make sense...

Anyway, once you have time, please have a look at the code, the Google Doc that I linked above, and see what you think. Once we make some decisions about API, I can go in and clean the code, and see what the minimal change I can make to the current version of develop, and make a PR. Since this is fully backwards compatible, we could start with a PR adding LIs to ac-gallery or something like that, which wouldn't break anything existing.

ac/ac-ck-board/src/index.js
```diff
@@ -8,17 +8,12 @@ import { meta, config, configUI } from './meta';
...
-      box.id = uuid();
-    }
-    if (box.title && box.content) {
-      dataFn.listAppend({
-        ...box,
-        x: Math.random() * 800,
-        y: -(Math.random() * 800)
-      });
-    }
+  (object.data || []).forEach(box => {
```
lfaucon, line 11: Can still configure list of starting boxes, but they won't be used.

We could discuss if starting data of an activity should always come from an operator or it could be part of the configuration somehow.

ac/ac-ck-board/src/board.js
```diff
@@ -28,45 +28,54 @@ const BoardPure = ({
...
+      <div>
+        <LearningItem
+          type="create"
+          dataFn={dataFn}
+          onCreate={e =>
+            dataFn.listAppend({
+              x: Math.random() * 800,
+              y: -(Math.random() * 800),
+              li: e
+            })
```
lfaucon, line 74: You write the coordinates `x,y` here in the reactive data of the activity. However in ac-brainstorm you made `score` part of the metadata of the LI. What are your reasons for choosing to do so? What do you think should be our general approach on that point?

ac/ac-classifier/src/Classifier.jsx
```diff
@@ -124,33 +87,44 @@ class Runner extends Component<Object, RunnerStateT> {
...
+          }}
+        />
       </Main>
     );
   }
 }
 
-export default withState('objectKey', 'setObjectKey', null)(Runner);
+export default compose(
+  withState('ourDataFn', 'setDataFn', undefined),
```
lfaucon, line 128: Should rename to `learningItemDataFn` or `liDataFn`, otherwise I will call all my components `MyComponent`

ac/ac-classifier/src/Classifier.jsx
```diff
@@ -124,33 +87,44 @@ class Runner extends Component<Object, RunnerStateT> {
...
+  withState('objectKey', 'setObjectKey', null)
+)(Runner);
```
lfaucon, line 130: Weird code here, why using `withState` if the component is written as a class?

ac/ac-classifier/src/Classifier.jsx
lfaucon: I don't really understand what is happening here actually.

Comparing that code with the code also found in ac-brainstorm. The `dataFn` props that we have when we use the `render` props of the `LearningItem` component, makes modifications on the metadata. Here you have implemented this in the thumbnail list to access the `dataFn` of teh LI we are working on, but couldn't you have accessed it where you display the LI in mode history?

ac/ac-classifier/src/components/ObjectPanel.jsx
```diff
@@ -1,63 +0,0 @@
...
-      <Container>
-        <Comp obj={obj} />
-      </Container>
-    );
-  } else {
-    // eslint-disable-next-line no-console
-    console.warn('No renderer for this object type');
-    return null;
-  }
-};
```
lfaucon, line 63: That's nice to see that functionality that we implemented before will now come for free with LIs :)

lfaucon (commented):
EDIT: didn't read your comment above before reviewing the code
______

Here is the first round of code review for the LI feature. I find it pretty neat how it works now. The UI needs work to make everything pretty in all situations.

The code looks nice, but it adds up one thing to learn to work on FROG.

## LI
- There is the compound LI "Idea with attachments". Do we need a developer to creae new compound LIs? I expected that this would be doable  
- In ac-chat with li-file, clicking the file downloads it three times
- I cannot compound "Idea with attachments" with another idea or with "Idea with attachments", but there is the option to select that type.
- li-table viewer is editable.
- In ac-brainstrom li-table has zoom and edit buttons which appear to do nothing

## Bugs
- Sorting of the list by number does not work anymore in ac-brainstorm
- You made changes to ac-ck-board. It is not working well for me. I think this bug had been fixed:
```
Warning: React does not recognize the `zDepth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `zdepth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (created by Paper)
    in Paper (created by WithStyles(Paper))
    in WithStyles(Paper) (created by ObservationContainer)
    ...
```
- In ac-brainstorm the edit icon should become a "save" icon when in editting mode

frog-utils/src/generateReactiveFn.js
```diff
@@ -36,6 +39,20 @@ class Doc {
...
 
+  createLearningItem(liType, item, meta) {
+    const id = uuid();
+    const itempointer = this.doc.connection.get('li', id);
+    itempointer.create({
+      liType,
+      payload: item,
+      createdAt: new Date(),
+      ...meta,
+      ...this.meta
```
lfaucon, line 50: Wouldn't it be better to have ?
```
{
      liType,
      payload: item,
      createdAt: new Date(),
      meta: { 
          ...meta,
          ...this.meta
      }
}
```

So that ac-pkg can safely write things in meta without overwriting other fields?

frog/imports/internalLearningItems/li-idea/index.js
```diff
@@ -0,0 +1,105 @@
...
+      title: 'Idea'
+    },
+    content: {
+      type: 'string',
+      title: 'Text'
+    }
+  }
+};
+
+const creator = withState('formdata', 'setFormdata', {})(
```
lfaucon, line 56: For `li-idea` is the Create component really different from the Editor component ? You use totally different code.

frog/imports/internalLearningItems/li-idea/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+export default {
+  viewThumb: viewIdea,
+  editable: true,
+  create: creator,
+  zoomable: false,
+  edit: editIdea,
+  name: 'Idea',
+  id: 'li-idea',
+  dataStructure: { title: '', content: '' }
```
lfaucon, line 104: I'm not happy with the naming of the different components. I'd rather write them with an uppercase. For example `{ Viewer, Creator, ThumbViewer, Editor }` or `{ ViewComponent, ThumbComponent, CreateComponent, EditComponent }`

Do we need that booleans editable and zoomable ? Maybe not as editable is true if there is a EditComponent and zoomable is true if there is a ThumbComponent (according to ViewComponent being the minimum required default)

frog-utils/src/generateReactiveFn.js
lfaucon: Also as ac-pkg will have access to the dataFn, I thought at first that they were only able to modify the `meta` field. With the full dataFn, we can imagine that an ac-pkg can implement a custom EditorComponent for a LI or even change the `liType`?

PR #935, opened 2018-04-05
CSS Cleanup

prastut:
This enables scroll on the Sessions page as a permanent fix. The height should be automatic to not allow scroll which I will take it up once I redo Sessions page.  

@lfaucon withStyles renders CSS based on the component name to make it local scoped. When we do this: 

```
styles = {
  root: {
    margin: 0
  }
}

const X = withStyles(styles)(({classes} => <div className={classes.root} />))
```

the CSS rendered on the client is `<div class="Component-root-1 /> `

whereas if you do: 

```
styles = {
  root: {
    margin: 0
  }
}

const X = ({classes} => <div className={classes.root} />)
const StyledX = withStyles(styles)(X) 
```
the CSS rendered on the client is `<div class="X-root-1 /> `

this makes CSS local scoped otherwise all components with direct withStyles will have a class of Component and then they will clash. ¯\_(ツ)_/¯

### DIFF
--- a/frog/client/stylesheets/bootstrap-isolated.css
+++ b/frog/client/stylesheets/bootstrap-isolated.css
@@ -3,14 +3,6 @@
  * Copyright 2011-2016 Twitter, Inc.
  * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
  */
-html {
-  font-family: sans-serif;
-  -ms-text-size-adjust: 100%;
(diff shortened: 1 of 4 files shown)


lfaucon:
@prastut good to know.

I think there are many other places where I did directly:
```
const X = withStyles(styles)({classes} => <div className={classes.root} />)
```

lfaucon (approved):

PR #938, opened 2018-04-05
Changing how session restart works

### LINKED ISSUE #839: Restarting now-broken graph crashes
If you edit a running graph to make it invalid, and try to restart graph, it will crash (because it sends the message "invalid graph", but the restart function isn't listening - should check for this and show message if cannot restart).

houshuang:
This makes two small but important changes to how session restart works. First, it restarts the session from the session graph, not the underlying graph. This means that if you go to the graph editor and makes some small changes, and then restart, these changes will survive. You can always start a new session if you want to "reset" from the initial graph. I think this makes more sense with how we actually work.

Also, before restarting, it checks whether the graph is broken, and does not let you restart if it is. 

Closes #839


lfaucon (changes requested):
screenshot from 2018-04-05 11-11-21

The naming has become crazy, can you update it accordingly to your changes?

houshuang:
Good point. I fixed the naming structure, now sessions always have a # first, and then they have a number at the end. I also fixed a bug where the "computed" property wasn't reset on activities.

frog/imports/api/sessions.js
```diff
@@ -185,13 +185,18 @@ const addSessionFn = (graphId: string, slug: string): string => {
 
     const sessionId = uuid();
     const graph = Graphs.findOne(graphId);
-    const count = Graphs.find({
-      name: { $regex: '#' + graph.name + '*' }
-    }).count();
-    const sessionName = '#' + graph.name + ' ' + (count + 1);
-
+    const match = graph.name.match(/(.+)\(([^(]+)\)$/);
```
lfaucon: Should be `/(.+)\((\d+)\)$/`

If I name a graph "hello (world)" the session associated becomes "#hello (NaN)"

lfaucon (changes requested):

houshuang:
@lfaucon Thanks, good idea. Fixed.

lfaucon (approved):

PR #941, opened 2018-04-05
Create ac ranking

jkolsen:


houshuang (dismissed):
If we can stop storing titles in the objects, I think this can be merged - then we can work on the dual + multiple view (we should support three in case of uneven numbers)

jkolsen:
I have fixed the storage issue. I would say this is ready to merge and I will start a new branch for the P2 plane changes.

PR #942, opened 2018-04-05
Display name on mouseOver in coordinates dashboard

lfaucon:
@houshuang What do you think of VictoryChart ? It was pretty easy for me to do what I wanted for this PR


houshuang:
How do I test this? It would be good to have some example data for dashboard preview. I tried different versions of the quiz preview, but none of them seem configured to generate the appropriate data?

lfaucon:
You can use ac-quiz example called "With Justification". And also the quiz should be submitted to send the coordinates

houshuang (approved):

PR #947, opened 2018-04-06
Remote graphs

romainAA:
should be done and everything works for me


frog/imports/api/remoteActivities.js
```diff
@@ -0,0 +1,36 @@
+import { uuid } from 'frog-utils';
+import { Activities } from '/imports/api/activities';
+
+export const removeActivity = (id: string) =>
+  fetch('http://icchilisrv4.epfl.ch:5000/activities?uuid=eq.'.concat(id), {
```
houshuang: you know you can also use 'string' + id ?

houshuang: can you make the http URL a separate const, something like

const RemoteServer = Meteor.settings.public.remoteServer || 'http://...'

and then reuse that in all the functions?

frog/imports/api/remoteGraphs.js
```diff
@@ -0,0 +1,43 @@
...
+import { Graphs } from '/imports/api/graphs';
+import { graphToString, doImportGraph } from '../ui/GraphEditor/utils/export';
+
+export const removeGraph = (id: string) =>
+  fetch('http://icchilisrv4.epfl.ch:5000/graphs?uuid=eq.'.concat(id), {
+    method: 'DELETE'
+  });
+
+export const collectGraphs = () =>
+  fetch('http://icchilisrv4.epfl.ch:5000/graphs').then(e => e.json());
```
houshuang: Can you only fetch the title and description, to save transfer?

frog/imports/api/remoteGraphs.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const sendGraph = (state: Object, props: Object) => {
+  const newId = uuid();
+  const grph = {
```
houshuang: do you really need to save one letter here? :)

frog/imports/api/remoteGraphs.js
```diff
@@ -0,0 +1,43 @@
...
+    title: state.title,
+    description: state.description,
+    tags: '{'.concat(state.tags.join(',')).concat('}'),
```
houshuang: what's with all the concat? :)

you can also use template strings: `{${state.tags.join(',')}}` - whatever is more clear :)

frog/imports/api/remoteGraphs.js
```diff
@@ -0,0 +1,43 @@
...
+    parent_id: Graphs.findOne(props.graphId).parentId,
+    uuid: newId,
+    graph: graphToString(props.graphId)
+  };
+  fetch('http://icchilisrv4.epfl.ch:5000/graphs', {
+    method: 'POST',
+    headers: {
+      'Content-type': 'application/json'
+    },
+    body: JSON.stringify(grph)
```
houshuang: are you sure you need to stringify the body? Did you try without? If you need to, that's fine.

frog/imports/ui/GraphEditor/EditorContainer.js
```diff
@@ -52,7 +59,26 @@ const EditorPanel = () => (
   </div>
 );
 
-class Editor extends Component {
+type StateT = {
+  exportOpen: Boolean,
+  importOpen: Boolean,
```
houshuang, line 66: You could have put these in store.ui, but it's OK like this too.

frog/imports/ui/GraphEditor/EditorContainer.js
```diff
@@ -74,16 +100,56 @@ class Editor extends Component {
         />
       );
     }
+    const setDelete = val => this.setState({ deleteOpen: val });
+    const setIdRemove = val => this.setState({ idRemove: val });
+    const setImportList = (val, fun?) =>
```
houshuang: Why do you want to manage the importList from here? Is it so that you don't have to refetch every time the user opens the library panel? I guess that makes sense. Make sure there is a reload in that case though - and maybe automatically refetch if more than x minutes have passed.

frog/imports/ui/GraphEditor/RemoteControllers/ModalExport.jsx
```diff
@@ -28,8 +30,13 @@ export default class ExportModal extends Component<Object, StateT> {
     };
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({ title: nextProps.activity.title });
+  componentWillReceiveProps(nextProps: Object) {
+    const name = nextProps.activity
```
houshuang, line 34: Don't quite follow - when would it receive new props?

frog/imports/ui/GraphEditor/RemoteControllers/RemoteLibrary.jsx
```diff
@@ -0,0 +1,102 @@
...
+          x.activity_type.toLowerCase().includes(searchStr)) ||
+        x.title.toLowerCase().includes(searchStr) ||
+        x.description.toLowerCase().includes(searchStr) ||
+        x.tags.find(y => y.toLowerCase().includes(searchStr)) !== undefined
+    )
+    .sort((x: Object, y: Object) => (x.title < y.title ? -1 : 1));
+
+class Library extends Component<Object> {
+  componentWillMount() {
+    this.props.setImportList([]);
```
houshuang: Here it seems like you reset the import list every time you open the library panel - then why do you need to pass it in from the root component?

houshuang (dismissed):
In general, really cool. Seems to work well.

I'd like the graph import modal to have a fixed size and width. Now it for example jumps around when I filter (because the list becomes shorter). We assume that there will be many graphs, with good descriptions etc, so it will take up most of the screen - not be tiny.

I don't quite understand how you are using the list of activities that you are passing down from the GraphContainer etc... It is probably a good idea to keep a central cache of activities and graphs, you can keep these in the state, or in the store. Then, whenever you open the library, you can keep track of when it was last updated, and if more than 10 minutes have passed, you can refetch. 

Make sure to have a reload icon for both the two library views. 

Also make sure only to fetch description and name, not the entire graph object.

PS: I meant "Request changes", not "Approve" :)

romainAA:
@houshuang The list is mostly used for the search tool. I need to keep it as a state to be able to filter it in real time.

frog/imports/ui/GraphEditor/RemoteControllers/ModalExport.jsx
romainAA: when for example you delete the graph you were in. otherwise it crashes. For the activities it's when the name of the activity selected is changed on the graph.

lfaucon:
@romainAA Is this still not ready to merge? What else do you have left? Let's try to get this reviewed and merged so we test it enough before next run of FROG

houshuang:
I got this error when I was exporting a graph:

screen shot 2018-04-13 at 22 52 44

You should update the cache whenever an activity or graph is exported - it's confusing to export a graph, and then go to the import list, and not find it there without reloading.

Also it does not work in the Preview window.

romainAA:
The preview window is not implemented yet. I still had a crash on Thursday when I left.

romainAA:
@houshuang everything works (I think). I 'll do another PR to insert it in the preview mode

frog/imports/api/remoteActivities.js
```diff
@@ -0,0 +1,48 @@
...
+    method: 'POST',
+    headers: {
+      'Content-type': 'application/json'
+    },
+    body: JSON.stringify(act)
+  });
+  Activities.update(props.activity._id, {
+    $set: { parentId: newId }
+  });
+  props.setModal(false);
```
houshuang: I wouldn't do setModal in the API, you could include a generic callback function instead. The api shouldn't know anything about the UI.

frog/imports/ui/GraphEditor/EditorContainer.js
```diff
@@ -54,7 +61,42 @@ const EditorPanel = () => (
...
+  importGraphList: Array<any>,
+  locallyChanged: Boolean,
+  idRemove: string
+};
+
+class Editor extends Component<Object, StateT> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      lastRefreshAct: new Date().getTime(),
```
houshuang: why put this here - you are updating after fetching below anyway?

frog/imports/ui/GraphEditor/RemoteControllers/LibraryListComponent.jsx
```diff
@@ -0,0 +1,92 @@
...
+    <div className="list-group-item">
+      <div style={{ marginLeft: '35px', minHeight: '48px' }}>
+        <div style={{ display: 'flex', flexDirection: 'row' }}>
+          <h5 style={{ fontWeight: 'bold' }}>
+            <Highlight text={object.title} searchStr={searchStr} />
+          </h5>
+          {object.activity_type && (
+            <h5 style={{ fontStyle: 'italic' }}>
+              <Highlight
+                text={' ('.concat(object.activity_type).concat(')')}
```
houshuang: or use string templates, like this:  
```
`(${object.activity_type})`
```

frog/imports/ui/GraphEditor/RemoteControllers/ModalExport.jsx
```diff
@@ -28,10 +30,18 @@ export default class ExportModal extends Component<Object, StateT> {
...
+      : Graphs.findOne({ _id: nextProps.graphId }).name;
+    this.setState({
+      title: name
+    });
   }
 
+  // lastRefreshGraph={this.state.lastRefreshGraph}
+  // importGraphList={this.state.importGraphList}
+  // collectGraphs().then(this.props.setImportGraphList);
   render() {
```
houshuang: Remove commented out code

frog/imports/ui/GraphEditor/RemoteControllers/ModalImport.jsx
```diff
@@ -0,0 +1,54 @@
...
+import Dialog, {
+  DialogActions,
+  DialogContent,
+  DialogTitle
+} from 'material-ui/Dialog';
+import Button from 'material-ui/Button';
+
+import Library from './RemoteLibrary';
+
+export default ({
```
houshuang: might be easier to just pass props to <Library {...props} />

frog/imports/ui/GraphEditor/RemoteControllers/RemoteLibrary.jsx
```diff
@@ -0,0 +1,142 @@
...
+import { collectActivities, importAct } from '/imports/api/remoteActivities';
+import LibraryListComponent from './LibraryListComponent';
+
+const filterWithStr = (list: Array<any>, searchStr: string) =>
+  list &&
+  list
+    .filter(
+      x =>
+        (x.activity_type &&
+          x.activity_type.toLowerCase().includes(searchStr.toLowerCase())) ||
```
houshuang: you could do const lowSearch = searchStr.toLowerCase() above, to avoid repetition

frog/imports/ui/GraphEditor/RemoteControllers/RemoteLibrary.jsx
```diff
@@ -0,0 +1,142 @@
...
+
+  componentWillMount() {
+    if (
+      this.props.libraryType === 'activity' &&
+      new Date().getTime() - this.props.lastRefreshAct > 600000
+    )
+      collectActivities().then(this.props.setImportActivityList);
+    else if (
+      this.props.libraryType === 'graph' &&
+      new Date().getTime() - this.props.lastRefreshGraph > 600000
```
houshuang: You can actually juse use new Date(). Try in a node terminal:

```js
a = new Date()
b = new Date()
b - a
```

frog/imports/ui/GraphEditor/SidePanel/ActivityPanel/index.js
```diff
@@ -7,14 +7,36 @@ import ChooseActivity from './ChooseActivity';
...
-  ({ activity }) => {
-    if (!activity) {
-      return null;
-    }
-    if (activity.activityType) {
-      return <EditActivity activity={activity} />;
-    } else {
-      return <ChooseActivity activity={activity} />;
-    }
+  ({
```
houshuang:
```js
export default withTracker(({ id }) => ({ activity: Activities.findOne(id) }))({activity, ...rest})

<ChooseActivity {...rest} />
```

houshuang (changes requested):
Looking pretty good. Some verbosity which I pointed out. There might be other stuff too, I don't have energy to review the whole way you've structured code, but it seems to work OK.

houshuang (approved):

PR #949, opened 2018-04-06
Show mouseover names for activities and operators in graph editor

houshuang:
This revolutionary PR displays the name of operators and activities when you put your mouse over it. Especially useful for operators, but also for activities with too long names to be displayed. We can set the delay before the tooltip is displayed in EditorContainer.


lfaucon (commented):
The tip is not showing for operators for me.

houshuang:
Fixed it. Turns out I had placed the component that actually shows the tooltip in a bad place, so it only worked when the sidebar was open (which also has a tooltip component :))

lfaucon (changes requested):
Now it is only working for me when the sidebar is open on an activity

houshuang:
I thought it was working before, but you're right. Fixed it again - it works for me now with or without side panel open.

lfaucon (approved):

PR #955, opened 2018-04-09
mergeData/Dashboard 2.0 in progress

houshuang:


ac/ac-chat/src/Dashboard.jsx
```diff
@@ -3,42 +3,40 @@
...
-    </div>
-  );
+  const data = [...Object.keys(state).map(w => ({ text: w, value: state[w] }))];
+  return { iMax, data };
 };
 
-const mergeLog = (data: any, dataFn: Object, log: LogDBT) => {
-  const tmp = log.value && String(log.value);
-
+const mergeLog = (state: Object, log: Object) => {
```
lfaucon: Why not using LogDBT anymore here ?

frog/imports/ui/Preview/dashboardInPreviewAPI.js
```diff
@@ -93,49 +84,160 @@ export const createLogger = (
...
+              : null
+          }
+        />
+      ) : (
+        <Viewer
+          state={this.state.state}
+          activity={this.props.activity}
+          instances={uniq(this.props.instances)}
+          users={this.props.users}
+        />
```
lfaucon, line 176: The props `config` is missing here, which makes the dashboard of ac-autocode crash (you can easily reproduce it). Later we added the props `activity` which makes the props `config` redundent, so we could remove it. It is in the type definition of dashboardViewerPropsT. We should get that flow fixed and working asap.

frog/server/dashboardSubscription.js
```diff
@@ -0,0 +1,63 @@
...
+import { DashboardData } from '../imports/api/activities';
+import { DashboardStates } from '../imports/api/cache';
+import { regenerateState } from '../imports/api/mergeLogData';
+
+const interval = {};
+const subscriptions = {};
+const oldState = {};
+const oldInput = {};
+
+const update = (that, dashId, func) => {
```
lfaucon: The naming is not self explanatory here. If I understand correctly what this is for I would suggest 

update => updateDataForDisplay
func => prepareDataForDisplayFn

frog/server/dashboardSubscription.js
```diff
@@ -0,0 +1,63 @@
...
+  if (!isEqual(oldInput[dashId], DashboardStates[dashId])) {
+    const newState = func
+      ? func(cloneDeep(DashboardStates[dashId]))
```
lfaucon: Is this cloneDeep simply here to avoid that func modifies the state ?

frog/server/main.js
```diff
@@ -3,6 +3,7 @@
 /* eslint-disable prefer-arrow-callback */
 /* eslint-disable func-names */
 
+global.window = {};
```
lfaucon: Is this to avoid crashes of client packages imported on the server?

ac/ac-autocode/src/index.js
```diff
@@ -42,7 +42,7 @@ export default ({
   configUI,
   validateConfig,
   ActivityRunner,
-  dashboard,
+  dashboards,
```
lfaucon, line 45: Good to have fixed that grammar mistake. It was bothering me too. Doing a quick text search I find many places in the code where this "s" has not been added. For example in real-life mode I find no dashboard because the test `!!aT.dashboard` returns false

lfaucon: Also, I don't think you've changed the type definition

ac/ac-brainstorm/src/index.js
```diff
@@ -53,9 +53,9 @@ export default ({
   id: 'ac-brainstorm',
   type: 'react-component',
   ActivityRunner,
-  Dashboard,
```
lfaucon, line 56: Nice to have updated this 3 versions old dashboard 👍

ac/ac-brainstorm/src/Dashboard.jsx
```diff
@@ -3,48 +3,46 @@
...
-        />
-      )}
-    </div>
-  );
+const mergeLog = (state: Object, log: Object) => {
+  state[log.type] = state[log.type] || {};
+  state[log.type][log.instanceId] = (state[log.type][log.instanceId] || 0) + 1;
 };
+
+export default { 'Event types': { initData: {}, mergeLog, Viewer } };
```
lfaucon, line 60: I feel that the code is less messy with the new API. I like how this dashboard here looks very reusable: counting eventType per instance

lfaucon (dismissed):

frog/server/dashboardSubscription.js
houshuang: yes

frog/server/main.js
houshuang: yeah but it actually doesn't help so I removed it

frog/imports/ui/Preview/dashboardInPreviewAPI.js
lfaucon: fixed

ac/ac-autocode/src/index.js
lfaucon: fixed

ac/ac-quiz/src/Dashboard/CountDashboard.jsx
```diff
@@ -3,18 +3,27 @@
...
+        vAxis="Possible answers"
+        hAxis="Number of answers"
+        categories={q.answers.map(x => x.choice)}
+        data={answerCounts[qIndex]}
+      />
+    ))}
+  </React.Fragment>
+);
+
+const prepareDisplay = (state: any, activity: ActivityDBT) => {
```
lfaucon: I think you have got the renaming of this function only half done. For example I observe that the count dashboard for ac-quiz works in preview perfectly, but not in a real session.

lfaucon (changes requested):

lfaucon:
Deepscan can be so mysterious. I would not be surprised nor sad if it ends up flying away from our repo some day :)

lfaucon (approved):

PR #961, opened 2018-04-10
Add coordinate dashboard to ac-ranking

lfaucon:


houshuang (changes requested):
Looks good, but something is broken - if I type in the justification box, it doesn't increase progress, and the submit button stays disabled.

jkolsen (changes requested):
The submit button is broken.

lfaucon:
The bug was due to usage of `Object.keys(activityData.data).length` being used, which I only half removed. This was pretty unsafe given that we don't know what operator will be linked to this activity and we have no control on what this object will be

houshuang:
Looks good to me, Jenny.

houshuang (dismissed):

jkolsen:
Yup, sure. Once Stian's PR has been merged in, we can merge this in.

jkolsen (approved):

PR #966, opened 2018-04-10
Log logIn to FROG

### LINKED ISSUE #849: Log data from students starting the graph
The logs currently log when students successfully start an activity but there is no information about when they enter the session as a whole. When the students enter the session, this activity should be logged.

houshuang:
This sends a new log message whenever someone logs into FROG, even if no activity is open. This means that in the majority of the cases (when an activity is open), we'll send two messages in short order, but I guess it's not a big problem. I also unify a bit how we handle the "teacher" user. 

Closes #849


houshuang:
Ping

lfaucon (approved):

PR #973, opened 2018-04-10
Use localStorage instead of url for saving preview's state

lfaucon:
Also fix a several errors in the examples config of ac-quiz.


houshuang:
Changed to using sessionStorage, which survives reloads, but can be different in different tabs. Also factored out the style code in Preview into a separate file.

PR #976, opened 2018-04-10
Lot's of small improvements to Preview

### LINKED ISSUE #917: Fix Cypress tests
The Cypress tests are broken because the CSS looks different after the redesign, we need to update the selectors.

### LINKED ISSUE #962: Add margins between config panel and preview
The config panel seems very squashed :)

screen shot 2018-04-10 at 10 18 55

houshuang:
- I fixed Cypress tests!
- Whole name/description of activity is now clickable in chooseActivity (also removed preview icon)
- Added a "back" button and title of activity on top of config form. Much more ergonomic for me to go back to chooseActivity like this if I'm scrolling through several activities quickly.
- Removed details about which activity from the actual Preview pane
- Added default empty config, which is the first selected, and which you can always go back to. Also went through activities and removed a bunch of empty examples.
- Added a bit of rightMargin to the config form so it's not so squished up against the preview

Closes #917 
Closes #962

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
     'import/prefer-default-export': 'off',
     'jsx-a11y/no-noninteractive-element-interactions': 'off',
     'jsx-a11y/anchor-is-valid': 'off',
+    'jsx-a11y/no-static-element-interactions': 'off',
     'jsx-a11y/anchor-has-content': 'off',
     'jsx-a11y/click-events-have-key-events': 'off',
     'jsx-a11y/href-no-hash': 'off',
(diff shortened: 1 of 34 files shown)


lfaucon (changes requested):
I see some unexpected behavior:
- Do you think "Empty" should be renamed "Default" ?
- Open preview on ac-quiz, then click the already selected Empty example, the config changes from an empty list of question to a list with one empty question (see images)
screenshot from 2018-04-11 11-07-43
screenshot from 2018-04-11 11-07-58
- Open preview on ac-quiz, then select "Statistic Quiz", then select "Empty config": The rte box of the first question  of the empty config does not update and shows the text of the first question of the Statistic quiz
screenshot from 2018-04-11 11-11-20

PR #978, opened 2018-04-11
Better handle of userId in the preview

lfaucon:


jkolsen (approved):
It fixed the issue in preview

PR #979, opened 2018-04-11
Collaborative Ranking

jkolsen:
Made collaborative ranking and refactor code. Still need to get the user names.


houshuang:
Looks great. One small thing, can we always keep the name of the people at the same place, no matter how many items they have submitted? We also need much clearer instructions that they cannot submit until they've ordered their lists identically, but I guess this is just the prompt. Maybe there could be a little "alert" if you click submit before you're ready though. People often skip instructions.

screen shot 2018-04-11 at 14 14 24

houshuang (changes requested):
Layout

jkolsen:
Yes, I will fix the name.

The instructions are just the prompt so that is just filler and will not be what we give them but I will look into having an alert.

ac/ac-ranking/src/ActivityRunner/Answer.jsx
```diff
@@ -0,0 +1,93 @@
...
+    dataFn.objInsert(newAnswers, ['answers', userInfo.id]);
+  }
+};
+
+export default (props: Object) => {
+  const { answer, memAnswers, userID, uiID } = props;
+  return (
+    <ListGroupItem>
+      <font size={4}>
+        {userID.toString() === uiID.toString() && (
```
houshuang: do you need this? userId should be a string always. With Louis' change, it will be a string in preview too

houshuang (dismissed):

jkolsen:
The changes are complete.

ac/ac-ranking/src/ActivityRunner/Answer.jsx
jkolsen: UserID was a string but uiID was a number, I had put it in because it seemed like it could not be guaranteed.

lfaucon (dismissed):

lfaucon (approved):

PR #980, opened 2018-04-11
Train Activity

prastut:


jkolsen (changes requested):
- [x] Could you address the errors found in the tests that failed?

ac/ac-train/src/index.js
```diff
@@ -0,0 +1,58 @@
...
+      },
+      data: {}
+    }
+  ]
+};
+
+const config = {
+  type: 'object',
+  required: ['timeOfEachInstance'],
+  properties: {
```
jkolsen: Could we specify here if it should be in seconds or milliseconds?

jkolsen (dismissed):
Comments from walking through interface:

Buy and Help Button: 
- [x] Could we put the buy button and maybe one for the help? It is not clear that these are clickable. 
- [x] Could we put the help button by the buy button since they are both interactive and the timer is 

Rest: 
- [x] The graphical interface map was not working.
- [x] One the map, could we make the dot for Bern smaller?
- [x] Add in correctness feedback. 
- [x] Instead of "Activity 1"...etc. could we say "Interface 1"?
- [x] When the timer is double digits, it is slightly cut off by the help button.
not?
- [x] For the graphical, the map gets very small making it hard to use while the text stays the same size. It may be better to also shrink the text. (optional)

prastut:
@jkolsen @lfaucon  which 🥇 ?

screen shot 2018-04-12 at 2 43 44 pm
screen shot 2018-04-12 at 2 47 03 pm
screen shot 2018-04-12 at 2 51 46 pm

jkolsen:
I vote for the first or third.

ac/ac-train/src/index.js
prastut: ^ @lfaucon how would I do this?

prastut:
#985 is essential to get some of the errors fixed that flow throws. #986 is essential because it involves some breaking changes with MUI.

jkolsen:
It looks like the progress dashboard still needs to be connected to the train activity.

jkolsen:
1. The interface currently only gives one type of ticket purchase (Geneva to lausanne) 
2. All interfaces are marking my input as wrong even when it is correct
3. The dashboards are putting all of the information under command whether I am filling out form or command etc

jkolsen:
The drag and drop activity just loaded as a hello world and crashed the dashboards

prastut:
@jkolsen apologies for the mess. I had to rewrite many files to get the dashboards working and forgot to remove the testing code. 

Give me some time to whip the code in shape.

prastut:
@jkolsen 

 > It looks like the progress dashboard still needs to be connected to the train activity.  
   
Fixed

> 1. The interface currently only gives one type of ticket purchase (Geneva to lausanne) 

Fixed

> 2. All interfaces are marking my input as wrong even when it is correct

Fixed. Needs to be checked. 

> 3. The dashboards are putting all of the information under command whether I am filling out form or command etc

Fixed.

prastut:
Documenting remaining thing: 

- [x] Problems with `time` reactive data structure. To check, for the very first interface -> first iteration, don't do anything, let the timer run out -> dashboards crash. Possibly will get fixed with the new logger. 

- [x] What if the student fills something out and doesn't press buy, should we make the answer undefined or use whatever state the answer is it in. Currently it's being undefined and so the answer is marked as false. 

- [x] Dragdrop crashes on preview because of conflict with React Mosaic react-dnd. https://github.com/react-dnd/react-dnd/issues/186

ac/ac-train/src/Dashboards/MeanPerInterface.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const interfaces = Object.keys(dash);
+
+  if (interfaces.length > 0) {
+    const allInterfaces = ['dragdrop', 'form', 'graphical', 'command'];
+
+    const coordinates = interfaces.map(int => {
+      if (whichDash === 'help') {
+        let countSum = 0;
+
```
jkolsen, line 37: When asking for help before any data has been put in the dashboards (help on the first instance of the first interface) the dashboards crash with error: cannot read property 0 of undefined.

ac/ac-train/src/ActivityRunner.jsx
```diff
@@ -0,0 +1,107 @@
...
+    dataFn.objInsert(false, 'guidelines');
+  };
+
+  beginActivity = () => {
+    const { dataFn } = this.props;
+    dataFn.numIncr(1, 'step');
+  };
+
+  componentWillMount() {
+    this.interfaces = ['start', ...shuffle(['graphical', 'command', 'form'])];
```
jkolsen: Drag and drop still is not included as an interface

jkolsen:
@prastut 

The graphical interface is still marking everything as not correct.

jkolsen:
@prastut 

-  What if the student fills something out and doesn't press buy, should we make the answer undefined or use whatever state the answer is it in. Currently it's being undefined and so the answer is marked as false.

To the question above. Let's leave it as undefined. If they did not press buy then it is timing out (as real interfaces can) and they did not successfully buy. To purchase a ticket, they must press buy.

prastut:
@jkolsen 

> When asking for help before any data has been put in the dashboards (help on the first instance of the first interface) the dashboards crash with error: cannot read property 0 of undefined.

Fixed using a bandage. When we have the new logger, these problems of undefined will go away. 

> Drag and drop still is not included as an interface

Fixed. But will not work in preview. Works normally. 

> The graphical interface is still marking everything as not correct.

Fixed. 

> If they did not press buy then it is timing out (as real interfaces can) and they did not successfully buy. To purchase a ticket, they must press buy.

👍

ac/ac-train/src/Dashboards/MeanPerInterface.jsx
```diff
@@ -0,0 +1,135 @@
...
+                }}
+                data={coordinates}
+                x="interface"
+                y="avg"
+                labelComponent={<VictoryTooltip />}
+              />
+            </VictoryChart>
+          </Grid>
+        </Grid>
+      </Paper>
```
jkolsen, line 118: The legend for the time and error per interface plots have the legend in the middle of the plot.
screen shot 2018-04-18 at 11 35 44 am

jkolsen:
On the intro screen, currently the interface type is not appearing. 
screen shot 2018-04-18 at 11 35 49 am

prastut:
Damn, I forgot to add the interfaces back. Just add graphical, form,
command to this.interfaces.

ac/ac-train/src/Interface.jsx
```diff
@@ -0,0 +1,191 @@
...
+    case 'command':
+      return <Command {...props} />;
+    case 'form':
+      return <Form {...props} />;
+    case 'dragdrop':
+      return <DragDrop {...props} />;
+    case 'graphical':
+      return <Graphical {...props} />;
+    default:
+      return <h1>Hello World</h1>;
```
jkolsen, line 22: This is getting returned as one of the interfaces
screen shot 2018-04-18 at 11 35 49 am

prastut:
https://github.com/chili-epfl/FROG/blob/6e834d8dc91647dfa50b801d43b1d28440a66e56/ac/ac-train/src/ActivityRunner.jsx#L66



You can add all the interfaces back to show all interfaces here.

jkolsen:
When I load the train activity (at least in preview), I get this error:

Stateless function components cannot be given refs. Attempts to access this ref will fail.
Check the render method of `DragDropContext(Content)`.

ac/ac-train/src/Guidelines.jsx
```diff
@@ -0,0 +1,158 @@
...
+}) => (
+  <Grid container justify="center">
+    <Grid container className={step > 0 ? classes.instance : ''}>
+      <Grid item sm={12}>
+        {activity !== 'start' && (
+          <Typography variant="display2" gutterBottom>
+            Interface {step}
+          </Typography>
+        )}
+      </Grid>
```
jkolsen: We do not need to label the interface number that they are on, so this can probably be deleted.

jkolsen:
For the first four graphs, there is always one extra point at the end of the graph that has not occurred. Could this point be deleted?
screen shot 2018-04-18 at 12 51 55 pm

ac/ac-train/src/Guidelines.jsx
prastut: Removing the number looks weird: 

image

lfaucon (approved):

lfaucon:
@jkolsen @prastut Great work on this activity. The dashboard code was so well written that it took me 2 minutes to update it to the new API. I haven't looked at any other parts of the code. I could make a few comments to improve the UI of each interface, but I think the best it to show the interface as it is now to Pierre tomorrow.

Issue #907, opened 2018-03-29
ac-image with categories


houshuang:
Doesn't look good after redesign

![screen shot 2018-03-29 at 12 23 07](https://user-images.githubusercontent.com/61575/38084135-02b60b14-334c-11e8-948a-1557e5a747da.png)

prastut:
Closed via #991

PR #991, opened 2018-04-14
Temp fix for ac-image

prastut:
The whole activity could do with a redesign. I was thinking to do it, but then shelved it since each activity can do with a redesign. It would be better to take them up separately or better on importance. Like which activities are being used in the next activity or are being most can go first, then rest of them follow.


prastut:
Closes #907

houshuang:
Thanks. Yes, agree - I think we should wait until we start implementing the Learning Items as well, which might force a bit of a rethink of both the interfaces, and which activity types that contain items that we really need - which can be merged, etc.

PR #995, opened 2018-04-16
Upgrading Yarn to 1.6

houshuang:


lfaucon (approved):
@houshuang If you don't have yarn@1.6 installed globally when running initial setup, it modifies the package-lock.json file. I committed it. If you are ok with the change, you can merge it.

PR #1007, opened 2018-04-19
Feature improve train graphical

prastut:
Making it more intuitive sticking to the realms of Material UI. Let's see if everyone finds it intuitive if not will have to step outside of it.


prastut:
Closed as merged in #1008

PR #1021, opened 2018-04-20
Trying to fix ReactiveHOC

### LINKED ISSUE #993: Forever spinning
Sometimes even a very simple graph (ac-create-group, and p2 chat) leads to never-ending spinner in student view. Might occur especially right after initial setup. We should fix this specific bug, but also make the login process more robust - with a timeout for the different places where it's waiting for Spinner, automatic reload, explicit error message (also sent to Sentry) etc.

houshuang:
I had several issues where activities were stuck with the Spinner from ReactiveHOC, because somehow dataFn was created, but data was unavailable - even though data had become available in the meantime. I tried solving this by using an interval, which gets canceled once you receive data, or if you leave the component. I also added a bunch of logging - we could keep some of it in production until we're sure we've fixed this problem, so we can diagnose it next time it happens.

However, now I had a new problem! Sometimes an activity would be "sticky" - even though I advanced in the teacher view, it wouldn't change. When checking the React tree, I saw that the <Runner> actually had the right (new!) activity id, but ReactiveHOC was still showing the old activity... I guess I messed up somewhere and made it so that ReactiveHOC doesn't always update on new props... As a quick fix, I added the two key statements in Runner, to try to force ReactiveHOC to always render from scratch from a new activity type. I think this worked, however someone's browser might have crashed (I asked them to reload, and then it worked - but I don't see anything specific about that on Sentry)...

Anyway, not saying this is the best way, but let's fix it together and make sure it says fixed - it's really fundamental!

Closes #993


houshuang:
@lfaucon Here's another example of another problem this PR also tries to fix, being "stuck" on an activity. If you look at the React tree, Runner is trying to load ac-dash, but ReactiveHOC is still showing chat...

I'm going to merge this now because I'm tired of getting these timeouts. It might not be perfect, but I think it's better than what we have now, and I'd rather do a lot of testing with this than without it - feel free to review and make a new PR to improve this!

screen shot 2018-04-25 at 18 55 27

PR #1023, opened 2018-04-22
Letting you deep-link to different components in Preview exampleData

### LINKED ISSUE #1022: Easier preview of multi-stage activities
Trying to develop a multi-stage activity, like the Train Activity, is frustrating - when you are working on a specific interface, it might take you a lot of time to get back to it whenever you make a change, and it reloads.

Perhaps we could have an extra argument in the exampleData, which gets picked up by the renderer, and let's us directly jump to specific parts or states of the application in Preview.

houshuang:
I had this idea that it would be nice to be able to work specifically on one component, without having to go through the whole random ordering every time you reload. I filed an issue, thinking it would be a lot of work. Then I realized that we can just pass a parameter from the exampleData, which does not exist in the config (ie. could never happen in a graph), and check for this in the ActivityRunner. Turns out it works perfectly! I added an optional type for exampleData, and render these special examples in italic, so users realize that these are not "valid configurations", but to test out something specific. I find it incredibly useful!

screen shot 2018-04-22 at 14 30 26

Closes #1022

### DIFF
--- a/ac/ac-train/src/ActivityRunner.jsx
+++ b/ac/ac-train/src/ActivityRunner.jsx
@@ -51,7 +51,8 @@ const ActivityEnded = () => (
 
 type PropsT = {
   dataFn: any,
-  data: Object
+  data: Object,
+  activityData: { config: Object }
 };
 
 class Main extends React.Component<PropsT> {
@@ -76,7 +77,14 @@ class Main extends React.Component<PropsT> {
 
   render() {
     const { step, guidelines } = this.props.data;
-
+    if (this.props.activityData.config.interface) {
+      return (
+        <Interface
+          whichInterface={this.props.activityData.config.interface}
+          {...this.props}
+        />
+      );
+    }
     if (step < 5 && guidelines) {
       return (
         <SpecificGuideline
--- a/ac/ac-train/src/index.js
+++ b/ac/ac-train/src/index.js
@@ -6,6 +6,7 @@ import dashboards from './Dashboards';
 
 const interfaceExample = int => ({
   title: `${int} interface`,
+  type: 'deeplink',
   config: {
     timeOfEachIteration: 600000,
     iterationPerInterface: 5,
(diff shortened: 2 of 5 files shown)


prastut:
This is brilliant! I was going to discuss this with you, I ran into a lot of these issues that if I wanted to hack on a specific component, I had to remove all interfaces-> add the specific one -> improve/commit -> put back all interfaces (which I kind of forget to do 50% of the times). 

Specifically, I was thinking to start a discussion whether we can branch out of all these components, for eg: 
* Having an activity which takes a sequence of mini-activities. 
* Each mini-activity will have components like activity-content and submit functionality, edit as you want. You can choose from a library of activity-content components like dual game, stroop game,  symmetry game, and other interfaces from the train and other activities. Since this individual components library is going to grow, if we do end up modularizing these, then there is going to be a ton of ac-X packages. Will need to discuss more on that. 
* Activity helper components that generalize over activities like: 
  * Guidelines before the activity. 
  * Help. 
  * Timer. 
  * Transitioning to the next activity, whether you want to show incorrect/correct. 

Right now the train activity components are written in a way that they can be easily put somewhere else but it would be difficult to tie these together or mix-match these components without an overlying structure. 

I remember a class in my middle school where teachers used to give us these simple activities (not at all as complex as what we are doing) but to get used to a computer interface. I can imagine a teacher who wants to try out just to do something cool in his/her class and wants to put together the dual game -> quick quiz -> dropdown form component.

houshuang:
Yeah, these compound activities are kind of awkward right now - Louis' induction activity is an example of an activity where it's quite likely that a teacher would want to do it differently. For this, the most important thing is flexible time, which we will work on, but we don't even have a spec/RFC for it yet (how will the UI look, what will the engine do)... Anyone is invited to start drafting this! 

In cases where there is a data dependency (like in Louis' case), we might also need more flexible/live operators. This we are also working on - learning items, and even some of the work with refactoring the dashboards, will come in handy here.

prastut (approved):

prastut:
Didn't think about the data dependency factor, I agree we need a spec document. I want to take it up as soon as I finish the redesign task.

lfaucon:
I hope `improved-train-on-feedback` will get merged along with your new feature @houshuang

PR #1025, opened 2018-04-22
Styling EnhancedForm with Material Design

houshuang:
Hi Prastut,

as you know, we use react-jsonschema-form very extensively in our app, for the SidePanel, config, etc. We have heavily customized it, and it would be non-trivial to switch to something else. RJF does not support Material Design, however they suggest various CSS solutions to show an MD view using Bootstrap classes. For example [MDB](https://mdbootstrap.com/).

I tried adding the MDB CSS to Meteor, using SASS to "isolate" the components to only respond to being under a `<div className='bootstrap'><div className='md-form'>...</div></div>`. I then apply these two classes to EnhancedForm. It kind of works, but is horribly broken - something about alignment/spacing etc. It's possible that is to do with the global Material Design, or something else. 

Could you have a quick look and see if it's easily fixable? If it is, it would probably make sidebar and other components look more integrated. If it's not easy to fix, or if after fixing it, some of our complex forms still don't look/work well, we can forget about it.

Examples:

screen shot 2018-04-22 at 17 08 52
screen shot 2018-04-22 at 17 08 46
screen shot 2018-04-22 at 17 08 40
screen shot 2018-04-22 at 17 08 26


prastut:
@houshuang looking over the code: 

* There are 2 bootstrap classes working on the component:
   * https://github.com/chili-epfl/FROG/blob/mdb/frog/imports/ui/GraphEditor/SidePanel/index.js#L43
   * https://github.com/chili-epfl/FROG/blob/mdb/frog-utils/src/EnhancedForm.js#L80
   * The second one has no effect since the first one encapsulates it. 



* The repeated md-form bubbles the CSS up. 
image
   * As seen from the image there are 2 CSS attaching to the form and then conflicting with each other, the div.md-form .md-form and the global bootstrap. 

I tried to remove global bootstrap but the isolation has problems: 

Original mdb.css 
```
.md-form input[type=text] {
  transition: all 0.3s;
  outline: none;
  box-shadow: none;
  border: none;
  border-bottom: 1px solid #ced4da;
  border-radius: 0;
  box-sizing: content-box;
  background-color: transparent;
}
```

Modified mdb.css

```
div.md-form .md-form input[type=text] {
  transition: all 0.3s;
  outline: none;
  box-shadow: none;
  border: none;
  border-bottom: 1px solid #ced4da;
  border-radius: 0;
  box-sizing: content-box;
  background-color: transparent;
}
```

According to CSS scoping rules, this is what the modified CSS would attach to: 

image


There will never be an `.md-form` class inside the `react-json-schema-form` because it renders bootstrap based definitions. The reason you see some changes is because of the double md-form, the scoping rules match, but then they create more problems. 

I removed the double scoping and the global boostrap and I get something on the screen, but since most of the form container styles depend on bootstrap so 💀 . The only way this could work is to write `!important` in mdl-form classes. 

It would be easier to write our custom styles (or fit the mdb ones properly), since `react-json-schema-form` provides a custom CSS class names. There are only so many elements inside a form so I don't think it would be hard, just time taking.

houshuang:
Closing this since @prastut is a genius.

PR #1028, opened 2018-04-22
Add templates to Yarn workspaces

### LINKED ISSUE #1011: Include templates in Yarn monorepo
To avoid versions getting out of sync

houshuang:
This adds the templates (ac/op) to the Yarn workspaces, meaning they will be included in `yarn upgrade-interactive`, and not be left behind. The files do not get compiled (they're not included in `watchAll` etc), and since all the dependencies are pooled at the root level, this does not result in more files being installed.

This immediately revealed that our templates were behind with regards to nps :)

Closes #1011


lfaucon (approved):
Thanks!

PR #1042, opened 2018-04-24
Removed hardcoding from train dashboards:

jkolsen:
Small fixes to remove hardcoding from dashboards.


prastut (commented):
@jkolsen woudn't it be better to use `iterationPerInterface` which is get's in every dashboard file as props `activity.data.iterationPerInterface`

That makes reading code easier to get.

ac/ac-train/src/Dashboards/MeanPerTryForEachInterface.jsx
```diff
@@ -51,7 +51,9 @@ const MeanPerTryForEachInterface = props => {
     }));
 
     const domain =
-      whichDash === 'error' ? { x: [0, 4], y: [0, 1] } : { x: [0, 4] };
+      whichDash === 'error'
+        ? { x: [1, count[interfaces[0]].length], y: [0, 1] }
```
prastut: @jkolsen we can avoid this by using `iterationPerInterface`

prastut (changes requested):

ac/ac-train/src/Dashboards/MeanPerTryForEachInterface.jsx
jkolsen: fixed!

prastut:
Thank you @jkolsen!

PR #1043, opened 2018-04-24
Adding copy to clipboard to ShowInfo

houshuang:
I cleaned up ShowInfo a bit and added an icon to copy each data structure to the clipboard. In my specific case, I was trying to generate a specific reactive data structure, and then copy this to a test case to test an operator.

It also shows both the underlying reactive data and the result of formatProduct (if that exists).


houshuang:
I think the issue we had with activityData etc is orthogonal to this, so I'd recommend accepting this PR, and I'll look at the other issue separately.

lfaucon:
Looks like another of your PR included these changes and has been merged. Reopen if not

PR #1044, opened 2018-04-24
Init data problem in Preview

houshuang:
Louis, do you want to take a shot on this, you might understand the logic flow better than me.

I fixed the out-of-sync ShowInfo problem (line 95 in  frog/imports/ui/Preview/Content.jsx) - I think we were also passing the wrong activityData to activities, but because most activities don't use activityData (they just use reactiveData and config), this was not discovered.

I've left the console.logs so it's easier to debug.

Note that in general we should be very careful about calling stateful functions (like initActivityData) from stateless functions like StatelessPreview, because we cannot control how often they get re-rendered, especially in this case, where it's subscribing to a million different props... Probably, this should be a class, which sets up initActivityData on componentDidMount or something, and then it can rely on Control/Content etc to explicitly reload activityData on user actions (it can still be "stateless" in the sense that it relies on external props and setters for all of it's data, but we would be able to move that function out of the render() function). 

screen shot 2018-04-24 at 10 48 08

### DIFF
--- a/frog/imports/ui/Preview/ConfigPanel.jsx
+++ b/frog/imports/ui/Preview/ConfigPanel.jsx
@@ -25,6 +25,7 @@ class ConfigPanel extends React.Component<*, *> {
     if (e.errors && e.errors.length === 0) {
       const aT = activityTypesObj[e.activityType];
       this.props.setConfig(e.config);
+      console.log('init from config');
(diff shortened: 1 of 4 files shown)


houshuang:
@lfaucon Can you remind me what the issue we had was (repro)?

lfaucon:
Closing, as fixed by #1128

PR #1047, opened 2018-04-24
Does not run mergeLog on the client

lfaucon:
In session mode, mergeLog would run on the student's browsers because of Meteor's optimistic UI feature.

### DIFF
--- a/frog/imports/api/mergeLogData.js
+++ b/frog/imports/api/mergeLogData.js
@@ -122,6 +122,10 @@ const archiveDashboardState = activityId => {
 };
 
 Meteor.methods({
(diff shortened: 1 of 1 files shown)


houshuang (approved):

houshuang:
Good catch. I believe this was originally in /server, which meant it would never get run on the client, but when we moved it to imports/api, because we needed some functions for Preview... So many unintended consequences... :)

How did you notice - did it cause crash?

jkolsen:
I was trying to test something and was too naive to know to look at the server part and was trying to debug in the student browser.

Issue #1017, opened 2018-04-20
Windows 10, IE 11 broken


houshuang:
I think this happened during the workshop, Window 10, IE 11.

<img width="979" alt="screen shot 2018-04-20 at 16 14 49" src="https://user-images.githubusercontent.com/61575/39056080-007537f0-44b6-11e8-85c3-4fad75fcee1f.png">

PR #1049, opened 2018-04-24
Deal with browsers

romainAA:
#1017 
Should work fine, I checked also the file uploading tool and the image viewer


frog/server/api.js
```diff
@@ -19,6 +19,20 @@ import { dashDocId } from '../imports/api/logs';
 Picker.middleware(bodyParser.urlencoded({ extended: false }));
 Picker.middleware(bodyParser.json());
 
+Picker.filter(req => {
+  const userAgent = req.headers['user-agent'] || '';
+  const IEStrings = ['Windows', 'WOW64', 'Trident'];
+  const isIE =
+    IEStrings.some(x => userAgent.includes(x)) || userAgent.includes('MSIE');
```
lfaucon, line 26: `isIE` is true if `userAgents.includes('Windows')`

PR #1050, opened 2018-04-24
Optimize prediction algorithm

Jiaxigu:
Features:
- Access the starting time of each student
- Always show actual progress till the latest student activity, instead of refreshing actual progress every X seconds (X - update interval for predictions)
- Restructure the algorithm and reduce 40% of running time (Train dataset: 1.1 ms per run of `prepareDataForDisplay`)

Fixed issues:
- Vertical line indicating current time now refreshes at the same frequency with curves
- Mock/test sessions now register the starting time of students (so prediction curves won't drop in some cases)
- Actual and predictive curves now have the same student counts (!)


jkolsen (changes requested):
After the first key press (second item in progress list), the estimated progress goes to 0 rather than 100. It works after this.

lfaucon:
## Observations

- Works well in dashboard preview
- Works well in real session mode
- Does not work in preview (see screenshot below)
screen shot 2018-04-25 at 22 14 55

## Suggested change

@Jiaxigu Please review the modification I made with my last commit. I don't know all the details of your algorithm, if that might break something. For example it breaks the dashboard preview as the actualStarting time is far in the past. The goal of that change is that we display the correct "now time" and the line flattens if no logs are received from students (see screenshot below). 
screen shot 2018-04-25 at 22 08 35

## Future improvements

- On the above screenshot you can see that when no logs are received your prediction still believes that the students are making progress and does not account that no data has been received. It then predicts that the completion and progress will jump in the next second, which is not true.

- Fix the preview and dashboard preview to better handle time in dashboards.

Jiaxigu:
@lfaucon It's a nice change - IIRC the `prepareDataForDisplay` only runs when new student activity comes in. We need to change it so the function runs with a certain frequency.

For the first point of future improvements - I agree with you. It has been concerning for a while. However, it's part of the limitations of any Linear Regression algorithms - it assumes students will work generally at a consistent rate.

I would like to defend LR because it's really efficient and versatile for FROG. The other algorithms work well under certain class scenarios - Gaussian Process only shines in normal distributed activities, for example. But LR is good for almost all cases.

My short-term plan is to 1) dynamically apply a penalty mechanism to trained weights, or 2) add [currentTime, progress] to the  `state` of a student to implicitly penalize prediction if no progress has been made for a long while. I will try to get this ready for May 1st. The long-term plan is to ensemble LR with other techniques to address this problem mathematically.

For the second point - I have an idea in mind to upgrade the data structure so we keep all the curves and `sessionStatus` and update them every time we run `prepareDataForDisplay`, instead of initializing everything with a discretization on the whole time span of the session. Then we could get more smooth and more precise curves on the dashboard.

PR #1051, opened 2018-04-25
Argue Operator with constraint

lfaucon:
Still needs: 
- Large scale testing
- Teacher should be left out of social structure


lfaucon:
Teacher excluded thanks to @houshuang contribution

lfaucon:
@houshuang we should merge this for now.

PR #1056, opened 2018-04-25
Improve cli tools

prastut:
Documentation: 

## Setup
* `npm start setup` - starts the git clean and initial setup process. 
* `npm start setup.clean` - does what update_package.sh did. 
* `npm start server` - often found myself browsing to the directory, with this I can start it from anyhwere
* `npm start setup server` - often after build, the next step is to start meteor. You can go get water/coffee in the meantime. 

## Build and watch
* `npm start build` and `npm start build.all` - self explanatory. 
* `npm start watch` - now there is no differentiation between `watch` and `watch.all`, instead there is just a general watch: 
   * `build` vs `build.all` has significant performance difference whereas the new watch is cheap. 
   * have found that I built for a specific folder but I like to have a general watch. 

How these play together: 
* `npm start build.all watch` - build all files and watch. 
* `npm start build watch` - cd inside any activity and I would do this. 

## Create Package 
`npm run create-package <activity|operator> <short-name> <title>`

`nps` has a problem with passing arguments, so I had to switch to npm scripts. 

## Rest of the toolset

* `npm start flow` and `npm start flow.quiet` - flow
* `npm start jest` and `npm start jest.watch` -  jest
* `npm start eslint` and `npm start eslint.fix` - eslint
* `npm start test` - combines all the above approriately.


houshuang:
This is cool. Maybe we should incorporate stuff like createPackage as well? One thing I'd really like to get working is to be able to yarn add a file - this breaks because yarn doesn't like that we've symlinked node_modules to frog (which we have to do because Meteor doesn't look in parent directories, like other node tools)... I had an idea for a script that would temporarily remove all the symlinked files, do the yarn add, and the re-add the symlinked files. Don't think I got it to work, but I didn't try very hard.

I wish there was a tool that would cache all the .gitignored files for each branch, and let you easily switch between different commits - it's annoying if I am working on two PRs at the same time, that have slightly different dependencies, and I have to initial_setup each time... Of course it's possible to check out different branches to different directories, but would be cool to have a tool that did this automatically for you :))

Anyway that's future plans - I'll try to review this when I get time. Thanks for the effort.

houshuang:
@prastut You don't want us to merge yet?

prastut:
@houshuang now it is ready to merge. 

>This is cool. Maybe we should incorporate stuff like createPackage as well? 

Had to do it npm scripts since nps has a problem with parsing arguments. There is a lot of info in nps GitHub issues and nothing works and nobody is replying. Avoiding to go down this rabbit hole. 

> One thing I'd really like to get working is to be able to yarn add a file - this breaks because yarn doesn't like that we've symlinked node_modules to frog (which we have to do because Meteor doesn't look in parent directories, like other node tools)... I had an idea for a script that would temporarily remove all the symlinked files, do the yarn add, and the re-add the symlinked files. Don't think I got it to work, but I didn't try very hard.
I wish there was a tool that would cache all the .gitignored files for each branch, and let you easily switch between different commits - it's annoying if I am working on two PRs at the same time, that have slightly different dependencies, and I have to initial_setup each time... Of course it's possible to check out different branches to different directories, but would be cool to have a tool that did this automatically for you :))

^ This is really good feedback. We can open an issue, I can take this up or [Hacktober fest](https://hacktoberfest.digitalocean.com/) is coming soon where we could get small wins from the community.

houshuang (approved):

lfaucon:
@prastut worth it to keep a documentation of the list of available commands for new contributors?

houshuang:
@prastut @lfaucon If you do npm start, it does print an autogenerated help, but because of the way we're autogenerating commands, it looks hopeless

screen shot 2018-04-27 at 11 31 15

We could easily add an npm start help with a quick summary of the tasks.

PR #1072, opened 2018-04-26
Fixed correctness issue

jkolsen:
Changing matching Geneva name so that it matches correctly.


houshuang:
Great find!

If you run
```js
./node_modules/.bin/jest -u
```

from root, you can update the snapshots and get the tests to green.

jkolsen:
Done!

houshuang:
I love how the PR diff captures so well exactly what was changed in the snapshot.

houshuang (approved):

PR #1077, opened 2018-04-27
Force reloading of side panel

houshuang:
Jenny had problems with sometimes text from one activity being "stuck" when switching to another activity of the same type in side panel. Keying the component to the activity._id should force a complete reload (ie. constructor being run etc) whenever we're switching activity types.

Eventually we might use this technique to clean up a lot of the reload uuids we're passing down, but for now, this seems like a quick fix that should clear up her problem (awaiting confirmation).


jkolsen:
Looks like it solved my issue.

jkolsen (commented):
Looks like it solved my issue. I will let Louis look at the code and approve it.

lfaucon (approved):

PR #1082, opened 2018-04-27
Cleaning up Orchestration View

houshuang:
This PR goes even further in trying to make the orchestration view as clean as possible for the teacher.

- Only leaving the "next activity" button. I make sure the session is started as soon as it is added.
- Moving settings and restart to the drop-down menu (we often restart, but teachers should almost never have to)
  - One further improvement could be to have the "next activity" button initially display as a green play button, and then switch to a blue "next" button. I'll leave that as an exercise for the reader.
- Always having the Dashboard/Graph toggle on top, the menus on the bottom, and the info about next activities (mainly for Dashboard view).
- Controversial: I removed the time progression line in the graph. I have noticed that nobody ever look at it, and I think it is confusing. We can rethink it and put it back in the future, maybe connected to progress etc.

There's probably lots of small CSS alignment issues etc, but for now, I think this is a big improvement already.


prastut (dismissed):
Have a look at #1088

frog/imports/ui/TeacherView/OrchestrationCtrlButtons.jsx
```diff
@@ -14,16 +14,32 @@ const OrchestrationCtrlButtons = ({ session, classes }) => {
...
+          <Grid item>
+            <ControlButton
+              btnModel={buttonsModel.next}
+              classes={classes}
+              style={{ transform: 'scale(1.5)' }}
+            />
+          </Grid>
+          <Grid item xs={5}>
+            <ul style={{ listStyleType: 'none' }}>
+              {(session.nextActivities || []).map(x => <li key={x}>{x}</li>)}
```
lfaucon, line 37:
```
Warning: Encountered two children with the same key, `Chat (p3)`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in ul (created by OrchestrationCtrlButtons)
    in div (created by Grid)
    in Grid (created by GridWrapper)
```

frog/imports/ui/GraphEditor/components/ProgressLines.jsx
```diff
@@ -13,13 +13,12 @@ export default connect(
...
     store: {
       ui: { scale },
-      session: { timeInGraph, timeInClass }
+      session: { timeInGraph }
     }
   }: StoreProp & { scaled: Boolean }) => {
     const s = scaled ? scale : 4;
     return (
       <g>
-        <ProgressLine color="red" x={timeToPx(timeInClass, s)} />
```
lfaucon, line 22: Why getting rid of this?

lfaucon (approved):
Looking forward for Prastut to re-do this correctly

frog/imports/ui/GraphEditor/components/ProgressLines.jsx
houshuang: I explained in my PR comment

PR #1088, opened 2018-04-27
improve-cleanerAndCleaner

prastut:
Changes: 
* Moved dashboard button down with the next activity.
* Projector View is now linked to to the Current Session ${slug} button 

Didn't spend too much time on bugfixing because the entire CSS of the page has to be fixed, had to go again with hacky CSS

### DIFF
--- a/frog/imports/ui/TeacherView/OrchestrationCtrlButtons.jsx
+++ b/frog/imports/ui/TeacherView/OrchestrationCtrlButtons.jsx
@@ -5,41 +5,28 @@ import Grid from 'material-ui/Grid';
 import { withStyles } from 'material-ui/styles';
 import {
   ControlButton,
-  OrchestrationButtonsModel
+  OrchestrationButtonsModel,
+  DashToggle
 } from './utils/buttonUtils.js';
 
 import styles from './styles';
 
-const OrchestrationCtrlButtons = ({ session, classes }) => {
+const OrchestrationCtrlButtons = ({ session, classes, visible, toggle }) => {
   const buttonsModel = OrchestrationButtonsModel(session, classes);
 
   return (
-    <Grid
-      container
-      spacing={16}
-      alignItems="flex-end"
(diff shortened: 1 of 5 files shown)


houshuang:
I don't agree with this approach. You've put the most dangerous action for a teacher (next activity affects all students, and is irreversible right now) right next to the most innocuous one. The two buttons are also conceptually totally different - one affects the state of the session, the other only the display...

prastut:
@houshuang my take on the design decisions:

* These 2 buttons come under orchestration for the teacher and observing how Pierre has been using FROG these buttons have been the most used. He tends to use these buttons *while* orchestrating the flow of the graph. 
* I borrow this concept from real life that when you see people orchestrating concerts they do have the power to switch to the next piece of music through a flick of the hand but also the have the notes in front of them to keep track of what has to happen and based on that make inferences. 

For eg: it's now easier for Pierre to switch to the progress dashboard and see if he wants to proceed to the next activity. 

> You've put the most dangerous action for a teacher (next activity affects all students, and is irreversible right now) right next to the most innocuous one. 

That is tech and in the hands of the developer. Do we trade usability for tech? Dangerous is upto us, Pierre doesn't know the underlying mechanisms behind how we build the tech, he only thinks about how smoothly he can control the class. 

> The two buttons are also conceptually totally different - one affects the state of the session, the other only the display..

Again what they do individually have little bearing on how they should be grouped. For the same reason menu items under `File` dropdown are grouped together even though they are conceptually different in what they do. 

I borrow these concepts from [Fitt's Law ](https://lawsofux.com/fittss-law.html).

houshuang:
I don't agree at all unfortunately. I will not merge this now, but I'm looking forward to lot's of discussion (and prototype testing etc) of the _real_ redesign, there I'm very open to different ways of doing it depending on the feedback we get from users etc.

prastut:
No problems :). I will close this for now.

PR #1090, opened 2018-04-28
Warn if teacher presses next.activity twice within 10 seconds

houshuang:
This is to avoid accidental double-clicking (or even fear of such, when you for example first have to focus the page and then click). If the teacher clicks next activity twice within 10 seconds, a confirmation will appear, and it's easy to cancel.


lfaucon (approved):
Great!

PR #1102, opened 2018-04-30
Remove buggy countdown

lfaucon:


houshuang:
It's in my later branch

PR #1110, opened 2018-05-03
External operators

houshuang:
[Screencast](https://www.youtube.com/watch?v=jI8A4Mu_qrQ&feature=youtu.be)

Louis, please have a look at the code and the ideas. I'll try to write up the specs on the wiki tomorrow. I'm not a regular Python user, so I'm open to discussing how we store the operator within FROG, how the Python API should be structured etc. When it comes to the integration between FROG and external operators in general, I think the current way works quite nicely, but happy to hear suggestions.

I think there's a small issue where an external operator might end up many times in the operator list in graph editor, I will look at that.

So don't worry so much about my code yet, but let's agree on the specs and the API etc, then I can make sure the code is really clean.

PS: The actual operator code in Python is very simplistic, since I haven't written Python for a long time, and don't know SciPy etc very well. Currently the number of students are hard-coded, and they have to be equally distributed, but the KMeans doesn't necessarily do that. So this should be improved - ie. it should allow any number of students, and the group size should be balanced (and probably there should be a config whether to have optimal group size, or a fixed group size). I'll leave that to someone who actually knows what they're doing with Python and machine learning :)


houshuang:
@lfaucon Please review this and let's discuss how I've implemented it. Happy to make changes, but would be good to get it in.

lfaucon (approved):

lfaucon:
@houshuang Do you approve the changes to the lock files? I think that should be the case on develop as we've moved to yarn@1.7.0

PR #1113, opened 2018-05-05
Improve package-scripts

houshuang:
`watch` was not working in sub-directories because of the echo, fixed that. Also, initial_setup was compiling index.jsx to index.jsx.flow, whereas the watch file would compile it to index.js.flow. Made this consistent.

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -31,7 +31,7 @@ if (findDir) {
 }
 
 const fromRoot = cmd =>
-  `cd ${dir}/ && PATH=${dir}/node_modules/.bin:$PATH} ${cmd}`;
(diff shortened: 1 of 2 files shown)


lfaucon:
What exactly is the issue with watch? For me it works when called from a sub-directory and making changes in other sub-directories

houshuang:
For me it makes a difference, maybe because I'm using fish shell. So if it works the same for you, but improves it for me, I guess it's good to merge?

screen shot 2018-05-07 at 09 46 46

lfaucon (approved):

prastut:
@houshuang [the log message](https://github.com/chili-epfl/FROG/pull/1113/files#diff-94ebe1c8ceb209223a3ed0e443db9d8bR70) gets displayed on running every command.

PR #1114, opened 2018-05-05
Enabling optional chaining 

houshuang:
Finally we can use optional chaining with support from ESlint and Flow:

`dashboard?[example]?.exampleLogs?[0]`

(Had to upgrade Flow to 0.70, and had to make some small changes to ac-train to make Flow run without errors)


lfaucon (approved):
Great !

PR #1116, opened 2018-05-06
Initial Implementation of the Presentation Activity

dinomujki:
Not ready to merge. To test, copy the 'pdf.js' file from the 'saved_pdf_module_file' folder to 'node_modules/pdfjs-dist/build/'. Switching PDF files causes issues in rendering for other users due to multiple successive updates to the reactive data.


houshuang:
Will start looking at your code to give you some feedback. Make sure to also address the issues, to make it "green". If you run `npm start test` you should get the same feedback on your local computer.

houshuang:
If I start a session with Projector view (teacher), and a student (in Safari): the student never sees the PDF, even if he clicks reload. The teacher initially does not see the PDF, however when he clicks load, he sees it. But if the student now clicks reload, le bordel quoi! :)

screen shot 2018-05-07 at 10 29 06

houshuang:
It also seems like you are rendering every page of the PDF, I thought the idea would be to only render one page at a time?

I think the goal to get this initial code merged (it doesn't have to be perfect), should be:

- [x] It should load the PDF right away (not wait for you to click a button)
- [ ] It should work both in Projector (teacher) view and student view
- [x] It should not do crazy things like the screenshot
- [x] You should specify a URL in the config of the activity
- [x] Showing the debug buttons should be an option in the config (default: off). This still let's us use it for debugging, but also let's us actually try what the activity will look like
- [x] It should only render a single PDF page at a time
- [x] The buttons at the top should be fixed (not scroll with the page)

If we can get this done, and the code is clean, we can merge - and then keep working on it. I'm not too worried about the aesthetics of the buttons etc for now, it can be fixed later, but it should be functional. 

The undo/redo looks like it works great!

houshuang:
For fun I tried with an actual slide deck that I used at a workshop a few weeks ago - where I did use FROG. You can try it as well: https://www.dropbox.com/s/oar8p88vef2ok6r/HSLU%20Seminar%20Weggis%2019.04.2018%20.pdf?dl=0

Warning, it's 75 MB, and 250 slides! :) (It was a four hour workshop, and I ended up not covering everything).

Of course, one issue here is that each student would have to download 75 MB, but this is a bit extreme. In my case, I would be happy to be the only one showing the slides, but showing them in FROG would mean a lot less context-switching when I was running small scripts in the middle...

Anyway I just tried and they loaded perfectly (it was from localhost, so of course fast). 

I see there is still a problem with CORS, I can look at fixing that.

houshuang:
Seems like CORS is a real issue, and that there are two fixes relevant for us:

- for testing, just focus on locally hosted files (either the public/sample.pdf or any kind of PDF uploaded in the graph editor like what I showed you)

- in the future, we can easily have the Meteor server proxy requests (ie. you would make a request to `http://localhost:3000?getPDF=http://dropbox.com/hello`, and the server would download the file and serve it from localhost, thus avoiding any problems.

houshuang:
Update! Things are working much better now. I was testing on my PR, but I don't think that changed much - the key difference is that I used localhost:3000 for both clients, traditionally I used dev1 for students, and I guess that led to issues (either because of CORS, or because you are fetching dev1 instead of dev1:3000 :))

So basically, with both on localhost, a file served out of clientFiles etc, things work so much better!!!

Files load when I start the activity, I can switch pages, annotate, undo, student can switch pages independently or revert to teacher etc. 

If I press load several times on the student, I do end up with a messy state:

screen shot 2018-05-07 at 11 44 20


It seems to me that this would be due to not properly cleaning up the previous state or DOM or whatever when doing a load? One way to do this would be to pass the URL as a key to the rendering React component, because if the key changes, it will force the component to rerender, and this might be enough to clean up (that won't help if the URL is the same, but it should prevent reloading from the same URL anyway). 

Also, it's unfortunate that it needs to re-render on the student when teacher is adding annotations, not sure if there is a way around this (the teacher doesn't rerender when annotating...).

Anyway very good progress!

Also one important thing is to somehow scale the PDF viewer to the window - since we're only viewing one page at a time, I think it should always be shown so that it fits the window, with no scrolling. Or at least width-wise, maybe if it's a letter sized document you can scroll down, but for slides you don't want any scrolling... Don't know if you can do this with CSS alone or you need to update the arguments to the PDF viewer? There is a way to listen to the size of a CSS element, and then update the props.

houshuang:
One more issue, after a while I get lot's of the text from the PDF drawn below the PDF. I'm not sure if this is because I reloaded the PDF or because it was redrawn when I made annotation/changed page etc.

[FROG.pdf](https://github.com/chili-epfl/FROG/files/1979456/FROG.pdf)

dinomujki:
Ok, I'll try to tackle things one-by-one:
 - The reason you were getting the text on the bottom was because the pdf_viewer.css was missing after copying the module. Will be fixed in my next push.
 - The code is rendering one page at a time since this is explicitly requested by the way I am using the module
 - However, it is really unfortunate that the page has to be re-rendered every time there is an annotation, but I could not think of another way (nor do I think there is one) due to the way the library is implemented and the fact that we are using reactive data. I think this should be relatively seamless for light-weight pdf pages.
 - The reason it worked better when you used localhost is because that is how I was testing it and because of the way I defined the initial 'pdf_source'.
 - I will definitely work on styling it more and possibly creating a 'full-screen' mode, although I am not sure how feasible this will be when annotations need to be reactively rendered.
 - Where/How do I setup this projector view and do you have any tips on testing it?
 - Lastly, every time the page gets rendered badly or looks messed up is when there are too many 'renderPage' requests at the same time. If you look at the module itself it is its code that suffers from this. I believe we can try to tackle this through some manner of preventing re-rendering until the previous render has finished, was definitely planning to look into this since this seems to be the major bug left.
 - Oh, and can you point me to an activity that has a nicely setup 'config' file that I can emulate for ac-prez activity

PR #1121, opened 2018-05-08
Fix progress dashboard in Preview

houshuang:


lfaucon:
Logging `activity.actualClosingTime` show the actual closing time changing over time in Preview of dashboard with example logs. Even if this makes things work in our situation, it is clearly not how is should be...

screen shot 2018-05-08 at 10 24 51

PR #1128, opened 2018-05-08
Fix initData on changing example in Preview

### LINKED ISSUE #1041: Preview, switch examples, mergeData not run
Currently in Preview, when you switch between different examples, it looks like mergeData is not run - for example with Brainstorm, the items do not show up until you explicitly choose "Reload". I think this was working before.

houshuang:
Closes #1041

### DIFF
--- a/frog/imports/ui/Preview/ConfigPanel.jsx
+++ b/frog/imports/ui/Preview/ConfigPanel.jsx
@@ -30,7 +30,13 @@ class ConfigPanel extends React.Component<*, *> {
     if (e.errors && e.errors.length === 0) {
       const aT = activityTypesObj[e.activityType];
       this.props.setConfig(e.config);
-      initActivityDocuments(this.props.instances, aT, -1, e.config, true);
+      initActivityDocuments(
+        this.props.instances,
+        aT,
+        this.props.example,
+        e.config,
+        true
+      );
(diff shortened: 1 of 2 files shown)


lfaucon (dismissed):
The activity reactive data is not initialised anymore when we change social structure because it does not trigger a config change. We should test more than the thing we fix, unfortunately.

I'd suggest making less things happen in this `onConfigChange` function. Inside the render function of the preview component was also not a good place. So what about finding all the user inputs that should trigger initActivityDocuments ? 

- changing the config (because the mergeFun can be different depending on the config)
- changing the social structure
- ...somewhere else ?

houshuang:
Hey,
I can't wait to rewrite this thing to use MobX. Seriously, it will solve many problems (not all!)...

the problem is we still have to trigger on config change, because if you manually edit the config, it needs to update. 

So I put back the update initData in Preview. Basically the only problem was that the config update was for some reason set to always set example to -1 (I guess maybe in the beginning we could only edit when there was no example data). When I removed that, everything works. Yes, it is true that when you make a change it will fire initData like four times, and eventually we should fix that, but for now this works.

lfaucon (approved):

lfaucon:
Maybe the solution is to move call the initdata function from the button clicks on the Control component, that should reduce the number of repeating calls

PR #1134, opened 2018-05-08
 Fixes the offset issues while dragging operators or connections

lfaucon:
Replaces #1084 

I moved around an onclick function for creating operators, because I wanted operators to follow the mouse cursor when being created, so the onClick was not registered on the background anymore. 

Please test usecases I haven't thought of


houshuang (approved):

houshuang:
This deserves a kinder egg!

PR #1136, opened 2018-05-10
Learning items 2.0

houshuang:
The spec is documented [here](https://github.com/chili-epfl/FROG/wiki/Learning-Items)

[Video demo](https://www.youtube.com/watch?v=Zt6mkEW0XbM&feature=youtu.be)

Currently, we have several liTypes (I took out morse and table, since morse was just for fun, and table we want to replace with a more functional one). Eventually, we will replace compoundIdea with a more flexible mechanism for templates, but I'm leaving it in for now, since it's important for testing to see compound ideas (it currently works to add a compound idea as an attachment to a compound idea, and so on and so on :)). 

I have only implemented it in one activity for now, ac-brainstorm. It will be quite easy to add other activities (specifically ac-chat, ac-ck-board, ac-image, and ac-classify), since I have the code, just need to update it and clean it up etc. 

It's working perfectly in both preview and session. Note that you might get some warnings in Preview - one is about updating state while rendering, I believe this is because initData is called multiple times right after each other - we are planning to fix this in the future anyway (but it doesn't cause any bugs), the other is too many eventEmitters, especially if you have a bunch of learning items and then add a lot of students. I don't think this is very serious, but we should keep an eye on it and do more testing etc - there might be better ways of doing it, but it has been very robust for me even with sixteen students all having four learning items open in the same browser window.

I updated ac-brainstorm's reactiveData structure according to what we discussed - although I decided to still go with the keyed object approach, since it seems like it makes a lot of dataFn operations more intuitive. We also have a nice function that automatically inserts new learning items into the activity's reactive data, with the meta data you ask for (ie. the meta data is not inserted into the LI, but the LI is created, and then a reference to the LI + your meta data, is inserted into the activity's reactiveData). 

I also decided to bundle LearningItem with dataFn. This is not just to avoid having another prop, but it was actually important to get it work with Preview that LearningItem knows about which `connection` is currently being used (in memory or remote sharedb), and it also enables the autoInsert I mentioned above.

Looking forward to getting this merged!

Next steps:
- converting more activities
- cleaning up layout etc (would love to get a make-over for the li-file / li-image)
- LI dashboard (I have the code) - also in preview, will be nice
- operators that understand the logic we discussed


frog/imports/ui/StudentView/Runner.jsx
```diff
@@ -108,23 +110,54 @@ type PropsT = {
...
+      createdByUser: string,
+      createdByInstance?: Object,
+      sessionId: string
+    } = {
+      createdInActivity: activityId,
+      createdByUser: userid,
+      sessionId
+    };
+    if (groupingKey) {
+      meta.createdByInstance = { [groupingKey]: groupingValue };
```
lfaucon, line 148: Is that the value we want here? I think that having `createdByInstance: groupingValue` which will be `userId` or `all` in P1 and P3 is much easier to use in general. Then we can also carry along `meta.groupingKey` which will be undefined for P1 and P3 and tell us what was the grouping based on in P2

houshuang: Fair enough, if you think this is better. I was trying to reduce a bit the data, since we'll have a ton of these LIs floating around, but I guess it doesn't matter... The thing is that we'll also want to know the individual person creating them, so in this case we'll be duplicating that data in instanceId and userId for p1, but I guess it's OK.

lfaucon (dismissed):

PR #1138, opened 2018-05-11
Converting Chat to support Learning Items

houshuang:
Also changing base data structure to keyed object to conform with our standard interoperable data structure. 

screen shot 2018-05-11 at 09 41 03


houshuang:
@lfaucon This has been updated with the latest code, it should be quite quick to review and get merged?

frog/imports/ui/Preview/Preview.js
```diff
@@ -59,6 +59,7 @@ const styles = {
 
 export const backend = new ShareDB();
 export const connection = backend.connect();
+window.connection = connection;
```
lfaucon, line 62: Do you want to push that code?

ac/ac-chat/src/Chat.jsx
```diff
@@ -106,20 +110,31 @@ class ChatController extends React.Component<StyledPropsT> {
       <div className={classes.root}>
         <h4 className={classes.header}>{activityData.config.title}</h4>
         <div className={classes.content} ref={node => (this.node = node)}>
-          {data.map(chatmsg => (
-            <Chatmsg msg={chatmsg} key={chatmsg.id} classes={classes} />
+          {values(data).map(chatmsg => (
```
lfaucon: The ordering works well, but it might be that if we change the function to generate uuid this will break. I'd suggest adding a field in these LI/messages objects to be sorted on. Timestamp might not work if computed by clients with unsynced clocks, but I think `order = 1 + Object.keys(data).length` can do, if two messages are sent at the same time, they will have the same order and it will not mess the sorting of messages anyway.

lfaucon (dismissed):

PR #1139, opened 2018-05-11
Fix operator panel

### LINKED ISSUE #1137: Operator list in side panel looks off
There is something slightly out of whack with the operator list in the side panel - doesn't look as nice as the activity list

prastut:
Why do we have separate components for Choose Activity and Choose Operator? 

The render function is visually the same for both these components (there is  a lot of code like the top panel that is repeated): 
* Choose Activity: https://github.com/chili-epfl/FROG/blob/fix-operator-panel/frog/imports/ui/GraphEditor/SidePanel/ActivityPanel/ChooseActivity.js#L99-L142
* Choose Operator: https://github.com/chili-epfl/FROG/blob/fix-operator-panel/frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/ChooseOperator.jsx#L146-L171

Closes #1137

### DIFF
--- a/frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/ChooseOperator.jsx
+++ b/frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/ChooseOperator.jsx
@@ -5,11 +5,9 @@ import React, { Component } from 'react';
 import Divider from 'material-ui/Divider';
 import Grid from 'material-ui/Grid';
 import Typography from 'material-ui/Typography';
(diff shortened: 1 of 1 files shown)


houshuang:
I thought about this, but there are also many small differences, and when we wrote it, it was easier to copy than to create a generic component with twenty different props and settings. I am planning to refactor all of this stuff (as I've mentioned) and I might also look at making the code more DRY.

houshuang (approved):

houshuang:
👍

PR #1140, opened 2018-05-11
Does not run initActivityDocuments in preview render function

lfaucon:
I think I tested everything

### DIFF
--- a/frog/imports/ui/Preview/ConfigPanel.jsx
+++ b/frog/imports/ui/Preview/ConfigPanel.jsx
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 import { uuid } from 'frog-utils';
+import { isEqual } from 'lodash';
 
 import Grid from 'material-ui/Grid';
 import IconButton from 'material-ui/IconButton';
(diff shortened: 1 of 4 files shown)


houshuang:
I made a few small changed: https://github.com/chili-epfl/FROG/pull/1140/commits/e457b2cfc324ebca9bfb6abb0cfb06432a795146

With your version, it was still always triggering init twice, once from control and once from config. With the new changes (I've tested it, but feel free to test more):

- if social plane/number of windows change, it's triggered from control and the shouldComponentUpdate prevents config from also triggering it
- if the example changes, the config will also change (even if the exampleData is identical, it hasn't yet gotten the "valid" key from config validator), so config will render and trigger, thus control doesn't need to trigger.

One funny thing is if you have an example open in p1, and then go to p3, switch example and go to p1, you will still see data from the previous example. This is because switching planes calls initData with refresh=false, which means it ensures reactive, but does not overwrite. This is good if you for example have one student, and want to add another one without loosing data, so I don't really want to change it. We can always refresh.

Merge if you're happy with my changes.

houshuang (approved):

PR #1142, opened 2018-05-11
Fix full screen preview

### LINKED ISSUE #1126: Preview full-screen mode single activity hides top
If you have a single activity open in preview, and you go to full-screen mode, the top of the activity is overlaid by the top bar.

Compare below.

There is also something weird about the container for the activity (the modal-body or whatever) - it's at 85% of the screen height for some reason - even more clear when you do full-screen. If I do height=100% in my component (for example ac-textarea) I get a scroll-bar, not sure how I should get this to fill the area but not overflow? And don't know why we need the 30 or so pixels at the bottom of the screen?

screen shot 2018-05-08 at 18 50 08
screen shot 2018-05-08 at 18 50 04

prastut:
Closes #1126

### DIFF
--- a/frog/imports/ui/Preview/Content.jsx
+++ b/frog/imports/ui/Preview/Content.jsx
@@ -4,6 +4,8 @@ import * as React from 'react';
 
 import { compose, toClass } from 'recompose';
 
+import { withStyles } from 'material-ui/styles';
+
 import {
   MosaicWithoutDragDropContext,
   MosaicWindow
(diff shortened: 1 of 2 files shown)


prastut:
This is also hacky in a way that the dismiss option to close the full screen view is inside the Controls. The best fix would have been to rely on [MUI's full screen dialog](https://material-ui-next.com/demos/dialogs/#full-screen-dialogs) as I don't have to think about the z-index property then.  

Trying this out.

houshuang (approved):

PR #1152, opened 2018-05-15
Ability to reset type of activities and operators

houshuang:
Add small icon in side panel for activities and operators that allows to reset type - useful to avoid having to wire up connections again etc. Asks for confirmation first.

### DIFF
--- a/frog/imports/api/activities.js
+++ b/frog/imports/api/activities.js
@@ -15,7 +15,7 @@ export const Connections = new Mongo.Collection('connections');
 export const DashboardData = new Mongo.Collection('dashboard_data');
 
 export const addActivity = (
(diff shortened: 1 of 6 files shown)


lfaucon (commented):
The activity box right after removing the type is not displaying correctly that the type is missing (see below, Chat has no type). The visual updates to be correct after I do another change. 

screen shot 2018-05-16 at 16 32 18

The icon choice seems that it is a delete button for the activity, not for the type only. I don't have a better idea for that icon, maybe moving it next to the selected type in the future.

houshuang:
Your issue (updating visual) should be fixed now.

lfaucon (approved):

PR #1170, opened 2018-05-22
Add fork

romainAA:
So, everything should work, I'll do another PR to change the library interface.
If you would like any modification, say the word
Here are the improvement given by this PR:
• being able to modify an existing activity or graph in the remote DB from one of its child
• being able to fork a remote activity to save it as a new one from one of its child
• full functionalities in Preview window
• show metadatas in Preview only if it's a child of a remote activity


houshuang:
Next time you do a PR, could you briefly describe what is new and how to test it? There are so many different things going on here, it's easy to loose overview.

Quick feedback:
- trying to preview an activity that you've imported in the graph editor crashes
- it seems impossible to add a tag in the metadata in the Preview sidepanel
- perhaps SAVE should only appear once you've modified the config (maybe not so easy to track?)
- SAVE should give some feedback somehow (for example by disappearing, if you did what I suggest above)
- when you fork an activity, you should start editing that new activity, ie. what you see in preview should reflect that new activity, if you save again, it should save over the new activity, etc.
- when you fork an activity, the fields should be pre-filled with info from the activity you forked from

An activity I exported to the cloud by using Save as new from Preview looks weird in the list

screen shot 2018-05-22 at 17 30 01

In general, we should stop requiring description for now.

houshuang (dismissed):
See my comment

romainAA:
For the SAVE button, it is indeed not easy to track changes, mostly because onConfigChange of the ConfigPanel is called I think more often than it should. So I cheated a bit my way, it now kind of works except if you refresh your window, then the changes stay, but the save button disappear. You need to modify something in the config to make it appear again

PR #1171, opened 2018-05-25
Enabling immutable learning items

houshuang:
When createLearningItem gets the option immutable, the id it returns will actually contain the whole learning item. These kinds of learning item types should not enable edit. Currently only enabled for file and image uploads. It's best to use ac-brainstorm to test, because you can see zoom, edit, history slider etc. I've tested with all LIs, including compound LI, where the ideaCompound is a shareDB doc, but the attachments inside are immutable. Seems to work well everywhere.


frog-utils/src/types.js
```diff
@@ -275,6 +275,7 @@ type LIRenderPropsT = {|
 |};
 
 export type LIRenderT = React.ComponentType<LIRenderPropsT>;
+type ImmutableLIT = { id: string, payload: Object };
```
lfaucon: Should be named something else that payload as it is not the payload of the LI. This will be confusing. I'd suggest `liDocument` or `liObject` and maybe adding a type definition for what this field expect (liType, payload, ...).
```
{
  "id": "cjhlndsqo00063i6jw0h924z2",
  "payload": { 
    "liType": "li-image", 
    "payload": { 
      "url": "/file?name=cjhlndsb100053i6ji1pzpdj1", 
      "thumburl": "/file?name=cjhlndsb100053i6ji1pzpdj1thumb" 
    },
    "createdAt": "2018-05-25T07:31:22.176Z" 
  } 
}
```

PR #1175, opened 2018-05-29
Change Meteor bundling

houshuang:


lfaucon (dismissed):
Make it green and go ahead :)

PR #1176, opened 2018-05-29
Library design

romainAA:
Changed the design of the library so that it uses Material-UI.
A also fixed one or 2 bugs like not being able to scroll the library.

Also I used the same List Component for the library and for the choose activity by modifying few things. Ideally we would also want to improve/reuse the topBar of the chooseActivity/activityLibrary for the graphLibrary, but if we could already merge this it would be nice since I won't have time to work on it for a while


prastut:
I see some things that could be fixed:  

* We are slowly moving away from inline styles and use the withStyles HOC, so if you can remove inline styles, that would be best. 
* Using `Grid container` + `Grid item` combination instead of fixing width [here](https://github.com/chili-epfl/FROG/pull/1176/files#diff-454823ebb250d62d4e371c8705256f49R88).
* Remove `position: absolute` and then position it correctly using the Grid container. 

A golden rule of thumb is that `position: absolute` should only be used in emergency cases. It removes the element from the document flow.

PR #1185, opened 2018-06-05
Further Additions to ac-prez activity (ScratchPad, Fonts)

dinomujki:


dinomujki:
I added the files 'for future records' as well since if you reinstall the dependency the changes are gone completely and I did not setup a file comparison for them, so cannot really be sure which exact things I changed as I was debugging it for quite a while.

houshuang:
Hi, looks like fantastic work! I just installed your PR without worrying about the custom library etc, and everything seemed to work really well, except the rectangle not changing color or shape. (What did you have to do with the text, it seems like it already works well?)

I then tried cloning the github repo for pdf-annotate, apply your changes, and publish it. I notice you edited the Babel-processed files, but I don't think that should make any big difference in the end. You can see my version of the package, with your changes applied, here: https://github.com/houshuang/pdf-annotate.js

I then published it to NPM, changed the reference in package.json and tried running ac-prez again, but this time it crashes when I open it... Not quite sure what happened, if I made a mistake when publishing the library etc.

screen shot 2018-06-06 at 10 46 54

PR #1187, opened 2018-06-06
Does not allow progress to decrease

### LINKED ISSUE #1155: Decreasing progress dashboard
Sometime when students go back in one activity, or when activity packages are not well implemented the progress dashboard receives logs of progress lower than previously received. This leads to weird situations. I'd suggest that the `mergeLog` function records progress using `max(log.payload.progress, previousProgress)`

lfaucon:
Closes #1155


houshuang:
If you go to the CS211 week 2 example, and replay it, it crashes pretty quickly:

```js
Uncaught TypeError: Cannot read property 'length' of undefined
    at mergeLog (modules.js:104007)
    at logs.forEach.log (app.js:15401)
    at Array.forEach (<anonymous>)
    at ShowDashExample.displaySubset (app.js:15401)
    at ShowDashExample.loadBySeconds (app.js:15355)
    at window.setTimeout (app.js:15365)
    at d (raven.min.js:2)
```

It's strange that actually running through the entire data works - since the final graph is displayed fine, but not when we replay step by step. It's possible that we're doing something slightly different in the replay than what would happen in a real situation.

houshuang (dismissed):
Don't crash Preview (or fix Preview)

houshuang:
It does something weird right at the end, it jumps from this to this

screen shot 2018-06-06 at 13 58 00

screen shot 2018-06-06 at 13 57 52

I'm wondering if it's something to do with actualClosingTime jumping ahead or something.

PS: We really should have numbers on the x-axis, it makes it very hard to see what's happening otherwise, since the domain of the x-axis keeps changing. THe ideal would be minutes:seconds, but if that's hard, at least seconds.

lfaucon:
I don't understand why it does that. I also don't understand why Flow error poped up.

> PS: We really should have numbers on the x-axis, it makes it very hard to see what's happening otherwise, since the domain of the x-axis keeps changing. The ideal would be minutes:seconds, but if that's hard, at least seconds.

There are number on the x-axis, scroll down to see them.

houshuang:
@lfaucon Closing this for now, let me know if you want to make improvements to this code in the future.

lfaucon:
@houshuang Fixed the bug above which was not due to the progress dashboard, but to the Preview functionality. Not very confident of my fix, the code is very messy and difficult to dig in.

PR #1199, opened 2018-07-03
Only hypothesis

houshuang:
This PR
- adds automatic import of LI types
- adds the possibility of activities/ops to bring their own LIs
- adds the possibility to preview an operator that fetches external data and generates LIs, the preview: true field has to be set. currently uses the ac-gallery code to preview, not so great to have core FROG code locked to a specific activity package, but silly to duplicate this code
- updates the Hypothesis operator to generate LIs

Test by previewing the Hypothesis operator in the graph editor.

I have RSS and Twitter operators too, but I have some issues with them, they will come hopefully soon.


frog/imports/activityTypes.js
```diff
@@ -36,3 +44,19 @@ export const activityRunners = entries(activityRunnersRawInternal).reduce(
...
 );
+
+const internalLIs = importAll.sync('./internalLearningItems/*/index.js');
+
+const packageLIs = [...activityTypes, ...operatorTypes].reduce(
+  (acc, x) => acc.concat(x.LearningItems || []),
+  []
+);
+
+export const learningItemTypesObj: {
```
lfaucon, line 55: We have the two files `opetratorTypes.js` and `activityTypes.js`. Don't you want to export this object from a new file `learningItemTypes.js`. Just to help with the logic of the code.

frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/EditOperator.jsx
```diff
@@ -45,6 +47,18 @@ const TopPanel = ({
...
+          <IconButton
+            icon="glyphicon glyphicon-eye-open"
+            tooltip="Preview"
+            onClick={() => {
+              ui.setShowPreview({
+                operatorTypeId: operatorType.id,
+                config: operator.data
+              });
+            }}
+          />
```
lfaucon: This button should be disabled if the operator is not valid. Right now it crashes on Hypothesis

lfaucon (dismissed):

frog/imports/activityTypes.js
houshuang: Somehow I can't because it is depending on both ActivityTypes and OperatorTypes - something like a circular dependency? I tried before... Maybe it would work now, I could try again.

frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/EditOperator.jsx
houshuang: Good point. Why does it crash though - is it if you don't type anything in any of the fields?

frog/imports/activityTypes.js
houshuang: I just tried again, and verified that it does not work - activityTypes is imported into learningItemTypes before it is defined, and crashes. I'm leaving this alone for now.

frog/imports/ui/GraphEditor/SidePanel/OperatorPanel/EditOperator.jsx
houshuang: Fixed

lfaucon: Yes, when I didn't type anything in the fields

lfaucon (approved):

PR #1200, opened 2018-07-03
Improving ac-ck-board

houshuang:
Adding support for Learning Items, fixing the layout for all three modes, enabling optional creation of new LIs inside the activity.

The CSS of the boxes, especially for things like images, could still be improved.


ac/ac-ck-board/src/ActivityRunner.js
```diff
@@ -26,27 +27,30 @@ const BoardPure = ({
...
+    <div
+      style={{
+        width: width + ' px',
+        height: height + ' px',
+        position: 'relative'
+      }}
+    >
+      {config.allowCreate && (
+        <div
+          style={{ zIndex: 9, position: 'absolute', right: '0px', top: '0px' }}
```
lfaucon, line 43: Prastut would be saddened by this amount of inline style, but I'm fine with it given the mess that this activity used to be.

lfaucon (approved):

PR #1201, opened 2018-07-03
Improving Gallery

houshuang:
This PR improves ac-gallery, by making the layout more flexible, using a library called Masonry, and also trying to make all the config options etc fully implemented (voting, adding any LI, adding specific LI, categories etc). 

The CSS of the detail (zoom view) should be improved, especially in terms of commenting.


ac/ac-gallery/src/ActivityRunner.js
```diff
@@ -9,14 +9,7 @@ import ThumbList from './components/ThumbList';
...
 
-const Main = styled.div`
-  display: flex;
-  width: 100%;
-  height: 100%;
-  flex-direction: column;
-  align-items: center;
-  overflow: hidden;
-`;
+const Main = styled.div``;
```
lfaucon: Should be removed

lfaucon (dismissed):
screenshot from 2018-07-09 13-40-01

It looks like everything is in one fixed-length column, so I don't know what Masonry is doing here. 

screenshot from 2018-07-09 13-37-43

This is the look of the commenting component. Deserves some work.

As we decided to use MUI, would this component fill the role of Masonry? https://material-ui.com/demos/grid-list/

houshuang:
I agree that the commenting function needs work, but I don't have time to work on it. 

Here's an example of the Masonry component "at work":
screen shot 2018-07-09 at 14 44 09

I tried the MUI thing, but it doesn't allow for different heights (as you can see above). I spent a lot of time getting this to work and look nice with a wide variety of different Learning Item types.

PR #1207, opened 2018-07-09
Fixing streaming for Learning Items

houshuang:
This updates streaming to work with Learning Items. 

- on autoInsert, it will also automatically try to stream the LI
- streaming now uses mergeData instead of automatically inserting (means not requiring path etc). 

Currently all activities using autoInsert (including ac-gallery) as well as ac-single-li, know how to stream. The mergeData function in all the LI-compatible activities should also work.

An example of this (this is functionality that Tony and Jim need, so I'd like to get it into develop as soon as possible). An example is the cloud graph Streaming demo 9.07.2018.

In the future, I'd like to merge operators and streaming so that you can connect any two activities, whether they overlap in time or not (currently the receiving activity has to be open when the first stream action happens). Related to the RFC on direct connections.


frog/server/stream.js
```diff
@@ -1,40 +1,42 @@
...
+      }
+      const toSend = {
+        data: { '1': { ...value, id: '1' } },
+        config: {}
+      };
       if (SharedbCache[docId]) {
-        const [doc, dataFn] = SharedbCache[docId];
-        safelyInsertObject(doc, dataFn, path, value, instanceId);
+        const [dataFn] = SharedbCache[docId];
+        mergeFunction(toSend, dataFn);
```
lfaucon, line 27: While testing with Chat, if the mergeFunction crashes, the whole app crashes and the server needs to be restarted. I suggest surrounding this with a try/catch.

frog/server/stream.js
```diff
@@ -1,40 +1,42 @@
...
+  stream: (activity, instanceId, value) => {
     if (activity.streamTarget) {
+      const target = Activities.findOne(activity.streamTarget);
       const docId = activity.streamTarget + '/all';
+      const mergeFunction = activityTypesObj[target.activityType].mergeFunction;
+      if (!mergeFunction) {
+        return;
+      }
+      const toSend = {
+        data: { '1': { ...value, id: '1' } },
```
lfaucon, line 22: For example the merge Function of ac-chat expects an array here, but I think we did agree on using keyed objects instead of array. If that's still what we want, let's look at why ac-chat merge works on array and let's update as much as we can to make activities stream compatible. 

Why is the id here `1` instead of the id of the original LI ?

lfaucon:
@houshuang I have not been able to test it. What activity type work for this? Would be great to quickly update many to make them work

lfaucon (approved):

Issue #1129, opened 2018-05-08
Red error lamp lost in Preview config


houshuang:
The green/red indicator of errors is gone after the redesign, this is important to let users know why their config is invalid.

prastut:
@houshuang need some help for this: 

The lamp post is actually present but due to old CSS, get's lost. This is where it is currently: 
![screen shot 2018-05-11 at 10 23 39 am](https://user-images.githubusercontent.com/10279686/39914546-99e1b850-5505-11e8-9806-0b2c0f950cac.png)


I then removed the CSS (it landed right at the bottom of the config panel)
![screen shot 2018-05-11 at 10 23 57 am](https://user-images.githubusercontent.com/10279686/39914564-aab5d3d2-5505-11e8-8491-6992c7aa6d11.png)


I found that the <Valid /> component is actually inside the Apiform file: https://github.com/chili-epfl/FROG/blob/develop/frog/imports/ui/GraphEditor/SidePanel/ApiForm.js#L249-L255

The placement of this component should go inside the top-panel of the configPanel: https://github.com/chili-epfl/FROG/blob/develop/frog/imports/ui/Preview/ConfigPanel.jsx#L90-L117

If you could give me some clarity on the dependency of this <Valid/> and since it's placed inside Apiform and I see it being imported in many places, without proper clarity refactoring will make more bugs.

PR #1210, opened 2018-07-10
Red Lamp preview

romainAA:
#1129


romainAA:
I ran npm start eslint.fix on my computer, but apparently the online testing doesn't liked it

PR #1212, opened 2018-07-16
Versioning

romainAA:
Everything seems to work with the imports, there is a graph in the FROG-graphs git to test it
https://github.com/chili-epfl/FROG-graphs/blob/master/VersioningTest.frog


ac/ac-gallery/src/upgradeFunctions.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+
+export default {
+  '0': (obj: Object) => ({ ...obj, oldGuidelines: obj.veryOldGuidelines }),
+  '1': (obj: Object) => ({ ...obj, guidelines: obj.oldGuidelines })
+};
```
houshuang: In this case, the upgrade function will still leave oldGuidelines and guidelines in place... would be better if it renamed... You could do this like so:

```js
({oldGuidelines, ...rest}) => ({...rest, guidelines: oldGuidelines})
```

ac/ac-quiz/src/upgradeFunctions.js
```diff
@@ -0,0 +1,70 @@
...
+      question: x.question,
+      answers: Object.values(x.answers).map(y => ({
+        choice: String(y)
+      }))
+    })
+  );
+  return newObj;
+};
+
+export default { '0': upgr0, '1': upgr1 };
```
houshuang, line 70: Are these just used for testing? Would these ever be used in practice?

frog/imports/api/activities.js
```diff
@@ -2,49 +2,146 @@
...
+            activity.configVersion === undefined ? 1 : activity.configVersion,
+            activityTypesObj[activity.activityType].configVersion
+          )(activity.data)
+        : activity.data,
+      configVersion: activityTypesObj[activity.activityType].configVersion
+    };
+    Activities.insert(newAct);
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.warn(e);
```
houshuang: You don't need eslint no-console for console.warn

frog/imports/api/activities.js
```diff
@@ -2,49 +2,146 @@
...
+      data: activityTypesObj[activity.activityType].upgradeFunctions
+        ? chainUpgrades(
+            activityTypesObj[activity.activityType].upgradeFunctions,
+            activity.configVersion === undefined ? 1 : activity.configVersion,
+            activityTypesObj[activity.activityType].configVersion
+          )(activity.data)
+        : activity.data,
+      configVersion: activityTypesObj[activity.activityType].configVersion
+    };
+    Activities.insert(newAct);
```
houshuang: When would this be called and actually need upgrading? Is it when importing? Wouldn't any activity that were in memory already have been upgraded?

frog/imports/api/activities.js
```diff
@@ -2,49 +2,146 @@
...
+        {
+          ...update,
+          data: activityTypesObj[update.activityType].upgradeFunctions
+            ? chainUpgrades(
+                activityTypesObj[update.activityType].upgradeFunctions,
+                update.configVersion === undefined ? 1 : update.configVersion,
+                activityTypesObj[update.activityType].configVersion
+              )(update.config)
+            : update.config,
+          configVersion: activityTypesObj[update.activityType].configVersion
```
houshuang: maybe you can try to split this code out (which is repeated from above) into a single extractUpdatedConfig(activity), which would return an updated config object - activity has activityType, which contains all the necessary information to do this?

frog/imports/api/activities.js
```diff
@@ -2,49 +2,146 @@
...
+      })
+    : insertActivityMongo({
+        _id: uuid(),
+        parentId,
+        configVersion,
+        activityType,
+        data,
+        groupingKey,
+        createdAt: new Date()
+      });
```
houshuang: Even though I don't always follow this myself, in general we try to distinguish between functions returning a value, and functions doing something, using ? : for the first, and if else for the second.

So age = man ? 15 : 20

but if(age>15) { delete(man) } else { delete(woman) }

frog/imports/api/activities.js
```diff
@@ -60,58 +157,20 @@ export const setStreamTarget = (activityId: string, streamTarget: string) => {
 };
 
 export const duplicateActivity = (actId: string) => {
+  // should update the old activity ?
```
houshuang: what do you mean?

frog/imports/api/graphs.js
```diff
@@ -159,5 +219,17 @@ Meteor.methods({
...
+  },
+  'graph.flush.all': graphId => {
+    Graphs.remove(graphId);
+    Activities.remove({ graphId });
+    Operators.remove({ graphId });
+    Connections.remove({ graphId });
+    Sessions.remove({ graphId });
+  },
+  'graph.flush.activity': activityId => {
+    Operators.remove({ from: activityId });
```
houshuang, line 226: Shouldn't this be Connections.remove?

frog/imports/api/operators.js
```diff
@@ -0,0 +1,92 @@
...
+        x.operatorType && operatorTypesObj[x.operatorType].upgradeFunctions
+          ? {
+              ...x,
+              data: chainUpgrades(
+                operatorTypesObj[x.operatorType].upgradeFunctions,
+                x.configVersion || 1,
+                operatorTypesObj[x.operatorType].configVersion
+              )(x.data),
+              configVersion: operatorTypesObj[x.operatorType].configVersion
+            }
```
houshuang: Same comment as in activities.js - see if you can separate this code into a function getUpgradedGraphConfig or something to avoid duplication

frog/imports/ui/GraphEditor/__test__/versionUpgrades.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+// import {chainUpgrades} from 'frog-utils'
+// import {GraphIdUpgrades} from './versionUpgrades'
+
+test('graph versioning test 1', () => {
+  expect().toEqual();
+});
```
houshuang: Remove this file?

frog/imports/ui/GraphEditor/versionUpgrades.js
```diff
@@ -0,0 +1,31 @@
...
+  '1': ({ graphId }: { graphId: string }) => {
+    Activities.find({ graphId })
+      .fetch()
+      .filter(x => x.activityType === 'ac-image')
+      .forEach(x =>
+        Activities.update(x._id, { $set: { activityType: 'ac-gallery' } })
+      );
+    Graphs.update(graphId, { $set: { graphVersion: 1 } });
+  }
+};
```
houshuang, line 20: Let's remove this - I would like the upgrade functions pure, they should not have direct access to database etc. For now, they'll only work on the object when direct import, in the future, the object function can work on the database too (because the graph will be stored as an object).

houshuang (changes requested):
Hi @romainAA , I just reviewed the code and left some comments. I still need to actually do a bunch of testing, since this PR affects everything, so it has to be solid. You can begin addressing some of my comments in the meantime.

ac/ac-quiz/src/upgradeFunctions.js
romainAA: Well, version 01 never existed, I just made it up for testing chains, and version 02 existed, but is probably pretty old and I don't think we will need to count them, hence the current version is still one and undefined will be replaced by one as default for configversion

frog/imports/api/activities.js
romainAA: For now it is called in :
- addActivity(): called when I pull an activity from the remote library where it is necessary to update the config. But this function is also called in 2-3 other places where it might not be relevant so I could create a separate method for remoteActivities instead of using addActivity
- duplicateActivity(): here I agree, it's probably not relevant. I could probably use the normal insert
- addGraph(): here I need it as it update the configs of each activity when we import a graph (from postgrest and from computer). The only thing that has been done previously is to update the global graphVersion

frog/imports/api/graphs.js
romainAA: I don't know, I really only moved this function as it made more sense for it to be in api/graphs instead of api/activities, but I did not change it's definition

romainAA:
@houshuang should be enough

PR #1217, opened 2018-07-18
Li dashboard

houshuang:
Adds a default dashboard which shows all LIs that are part of the graph. Currently show only LIs stored as ShareDB documents, which is usually what we want (ie creative works and not tweets etc). When zoomed in, you can view the history of any element, you can do a free-text search (not case insensitive yet), and filter by specific activities (where the LI was created). I'm sure we can do more around this in the future, but it's a good start.

### DIFF
--- a/frog/imports/ui/Dashboard/ImageBox.js
+++ b/frog/imports/ui/Dashboard/ImageBox.js
@@ -13,18 +13,6 @@ const styles = theme => ({
   })
 });
 
-const getStyle = styleCode =>
-  ({
-    chosen_by_team: {
-      border: 'solid 4px #009900',
-      borderRadius: '5px'
-    },
-    chosen_partially: {
-      border: 'solid 4px #FFFF00',
-      borderRadius: '5px'
-    }
-  }[styleCode] || { border: 'solid 2px #a0a0a0' });
-
 const ImgButton = styled.button`
   position: relative;
   border: none;
@@ -63,6 +51,5 @@ const ImageBox = ({
     </Paper>
   </ImgButton>
 );
-// style={getStyle(styleCode)}
 ImageBox.displayName = 'ImageBox';
 export default withStyles(styles)(ImageBox);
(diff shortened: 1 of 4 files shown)


frog/imports/ui/Dashboard/LIDashboard.js
```diff
@@ -0,0 +1,224 @@
...
+              style={{ position: 'absolute', right: '0px' }}
+            >
+              <span className="glyphicon glyphicon-remove" />
+            </button>
+          </div>
+          <div style={{ layout: 'flex', flexDirection: 'row' }}>
+            <div style={{ marginBottom: '50px' }}>{props.children}</div>
+            <hr />
+            <div>
+              <div>
```
lfaucon: Looks to me that these two divs are useless

frog/imports/ui/Dashboard/ImageBox.js
```diff
@@ -0,0 +1,55 @@
...
+  background: none;
+  max-width: 250px;
+  height: 250px;
+  width: 100%;
+  margin: 5px;
+  padding: 0px;
+  flex: 0 1 auto;
+`;
+
+export const CenteredImg = styled.div`
```
lfaucon: This is not imported anywhere, we might as well get rid of it.

frog/imports/ui/Dashboard/ImageBox.js
```diff
@@ -0,0 +1,55 @@
...
+
+const styles = theme => ({
+  root: theme.mixins.gutters({
+    marginLeft: 20,
+    paddingTop: 36,
+    paddingBottom: 16
+  })
+});
+
+const ImgButton = styled.button`
```
lfaucon: You use both `mui-withStyles` and `styled` in the same file.

lfaucon (approved):
This works very well. I only fixed CSS details

PR #1225, opened 2018-07-25
continue removing bootstrap

romainAA:
Remove bootstrap completely from autocode, brainstorm and dual


houshuang:
Good contribution. Could you make the buttons always be located in the top right, not in the middle? Also, I would prefer the number of votes inside the box and not in the corner, not clear which box it belongs to now.

screen shot 2018-07-25 at 13 21 23

Also in ac-dual, could you make the buttons larger
screen shot 2018-07-25 at 13 22 56

There's also a problem when the two activities are shown next to each other (pingball and geometry), the view is larger than the screen, and the left and right keys actually scroll the screen instead of moving the paddle.

houshuang (changes requested):
See my comments

houshuang:
@romainAA Could you address my concerns so we can get this merged? Thanks

PR #1228, opened 2018-07-25
Fix upgrade

romainAA:


houshuang:
Can you explain what the problem was - was it trying to upgrade activities/operators that did not yet have an ac/op type (just added to the graph)? So I guess it didn't really have anything to do with git clean (git clean doesn't actually empty the database, only meteor reset does that)?

houshuang (changes requested):
Check comment

PR #1235, opened 2018-07-26
li-spreadsheet

houshuang:
Needs some cleanup to be green


houshuang:
The CSS doesn't look too good (not sure what the issue is, I am importing the bundled CSS, but maybe it gets overwritten). Also, in the future I want this to be more flexible (ability to add rows or columns by student, create template for teacher etc). For now, it does work and should be merged.

lfaucon (approved):

PR #1247, opened 2018-08-02
Fix upgrade ac without activityType

### LINKED ISSUE #1241: upgrade empty activity
Importing a graph that contains activities w/o any activity_type opens a warning modal saying that the config couldn't be updated

romainAA:
fix #1241  but I get a weird flow error that seems not to be specific to this branch


romainAA:
apparently its only on my computer

houshuang (approved):

PR #1248, opened 2018-08-03
LI spreadsheet improvement

romainAA:
Adding and deleting rows and columns on the edit mode is now available


houshuang (changes requested):
Seems to work great, however I managed to get it into this state. I think I first added some columns, then deleted, then added some rows.

I'd also love for the buttons (+/-) to be less obtrusive, but still easy to find, but not sure how to do that. If you can fix the first bug, I'll merge, but if you have ideas for the second issue, I'd be interested.

houshuang:
Here's the problem:

screen shot 2018-08-04 at 00 43 59

PR #1250, opened 2018-08-03
Improve the Image LI

romainAA:
Add an editor to the learning item

### DIFF
--- a/frog/imports/internalLearningItems/li-file/utils.js
+++ b/frog/imports/internalLearningItems/li-file/utils.js
@@ -25,16 +25,11 @@ const uploadBufferWithThumbnail = (
             if (cb) {
               cb();
             }
-            createLearningItem(
-              'li-image',
-              {
-                url,
-                thumburl,
-                filename
-              },
-              undefined,
-              true
-            );
+            createLearningItem('li-image', {
+              url,
+              thumburl,
+              filename
+            });
           });
         });
       });
--- a/frog/imports/internalLearningItems/li-image/ToolBar.js
+++ b/frog/imports/internalLearningItems/li-image/ToolBar.js
@@ -1,0 +1,32 @@
+// @flow
+
+import * as React from 'react';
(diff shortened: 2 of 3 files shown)


romainAA:
@houshuang After adding an Editor attribute to the li, when I console.log his props, I get the right data props, but dataFn is undefined. Do you know why ?

houshuang:
@romainAA Don't know why you don't see it, but if I open the image in edit mode, and look in the Chrome devtools, I see the dataFn very clearly as a prop. (I tried in session as well, not just in preview). If dataFn was not there, none of the other editing (li-idea etc) would work.

screen shot 2018-08-07 at 09 26 49
screen shot 2018-08-07 at 09 26 39

houshuang:
Sorry, I figured it out! We currently have two kinds of learningItem storage - one where each LI is a separate ShareDB document, and one called "immutable" where the content is stored directly in the reactive data of the activityType - which we use for images and files... Thus there is no dataFn. I made the images not immutable (I pushed a small change to your branch), so now it should worK! Didn't think about this at all.

romainAA:
It works, I made 5 possiblilities to edit, rotate both ways, zoom in, zoom out and reset.
However, I am quite bad with css and I could not get to modify the window size when the image is rotated and its width is smaller than its height, so for now it overflow the paper that I set as hidden.
You can test it and see what you want to do with it

houshuang:
This is great work. However, the changes you make in edit mode are not reflected in view or thumbView (but they are shown if I close edit view and open it again, so it's stored). I guess you need to modify the rendering in view and thumbView. I would suggest just keeping the two rotate options for now - then we also don't need a "revert" button, since you can just rotate four times. The zoom is not so useful right now - we might bring it back in the future.

Let's get this cleaned up and merged (ie. only rotate left, right, but also reflected in view/thumbView), and I'll add a separate issue for more annotation tools.

PR #1252, opened 2018-08-06
Ac select

romainAA:
Seems to work perfectly for simple text. For rich text, it is more complicated as the Highlighter component takes the text to display as a property of type string called textToHighlight.
I don't think we could use this module if we want to use rich text.


romainAA:
We would have the same problem with multiple color for different words. Indeed we would need to imbricate Highlighter components to change the style of the highlighted word for each word and Highlighter is not to have children

houshuang:
The dashboard doesn't work at all (just crashes). As I said, please use reactive dashboards (see how the word cloud dashboard from ac-chat is done - you don't have any mergeLog, but just a function that iterates over the reactive data, and a viewer). 

Also, it currently highlights parts of a word (for example selecting the, will also highlight the first three letters in theory). 

Let's get the dashboard working, and fix that issue, and we can get this merged for now. Then we can look at how to deal with more complex cases.

romainAA:
That's weird, the dashboard works perfectly for me (even if it's pretty ugly, I'm gonna change that) but it doesn't crashes and neither does it on Louis'computer

lfaucon:
@houshuang This does not crash for me (I tested in preview). Can you share what is the problem you had?

Problems I found: 
- The text should be scrollable 
- Should use Material-UI table for the dashboard
- The title can be clicked for highlighting but does not highlight itself, that's confusing
- When selecting more than one word an "empty" entry appears in the dashboard.
- https://regexr.com/ <- Can be used to tinker with regexp

houshuang:
Got it, maybe I was looking at an old version (without pulling). 

@lfaucon I can scroll OK, what is the problem?

if you can fix the dashboard (as Louis suggests), add highlighting for title, avoid highlighting partial words (a regexp that includes word breaks), we can accept.

I didn't see the issue with "empty" - maybe you fixed it?

Also you can change the cursor to a pointer (with CSS).

romainAA:
@houshuang done, if you don't like the question mark cursor, you can change its value to pointer.

PR #1255, opened 2018-08-08
CSS editing li spreadsheet

### LINKED ISSUE #1253: CSS in li-spreadsheet
When you are editing a cell, it looks really weird. Once you have entered a value, it looks fine. This was not the case before, not sure what happened to the CSS. Would be great to get it fixed to make it look a bit nicer.

romainAA:
Closes #1253


romainAA:
@houshuang I modified a bit flow config, because for a while I have been having flow errors at places 
that should have been ignored. Can you check if it causes any problem or unwanted behavior ?

Issue #1246, opened 2018-08-02
UI Changelog


houshuang:
We will have a file in api/ which defines a changelog based on versions. (For example: {1: "Added a new shortcut in graph editor, r, which automatically improves your graph"}. On login, it should check if the field lastVersionUpdateSeen exists (if it doesn't exist, it means it's a new user - show the help modal from the graph editor, and write lastVersionUpdateSeen = whatever is the current version, or 0). 

If it does exist, and there are "newer updates", show these updates in a modal (use HTML from frog-utils to display so we can add some HTML markup like links), in reverse order (ie. newest first, up until the last "new" item).  when entering the graph editor for the first time, and then write the new version number seen to the user database. (This only applies when logged into the teacher view of course). 

The button called Logout on the menu bar should be changed to a pulldown similar to what most sites have, where you can change your password (in the future etc), but also see a link to the FROG project site, and see a modal with all the changes (starting from 0).

Let me know if something isn't clear.

The use case is that we have CS411 students registered to the server for four months, and we make some changes (like "now you can copy an activity from one graph to another"), and we upgrade the server code without them noticing - when they log in, we need a way to communicate changes with them.

houshuang:
@romainAA maybe this next?

PR #1256, opened 2018-08-09
Changelog

romainAA:
#1246 
The UI of the modal isn't perfect (I guess width should be fixed) but it works well.
There are also 2 dummy buttons in the logout menu for redirecting to frog website and changing password.


houshuang:
Looks good, however you should keep track of the latest update seen by each user - also, they should not see any updates the first time they log in. So the first time they log in, you set the counter to the current update, and after that check if there's anything new after that.

For that you probably need version numbers in the api/versions.

houshuang:
You should probably merge in the multipleTeacher stuff in this branch, so that we can test that it keeps track of what different users have seen etc. This functionality only makes sense with multiple teachers anyway.

romainAA:
@houshuang Everything should be good, but since I merged MultiTeacher in it, we might want to merge multiTeacher before.

houshuang:
@romainAA With a short changelog message, the window looks comically small - needs minimum width/height:

screen shot 2018-08-20 at 12 48 42

Other than that seems to work well. We should also add a field for date in api/changelog.js, and when we display many, we should sort them in reverse order (newest first). 

Other than that, ready to merge once we get multiTeacher in.

PR #1259, opened 2018-08-13
Partial fix #1214

### LINKED ISSUE #1214: Cleaning up ac-gallery, learning items, comments
ac-gallery is used both as a stand-alone activity, and it is also used for the preview of operator content. It is a great example of how we show a wide variety of content using different learning items, but it needs some love.

1. Somehow the Masonry component we use nicely formats content in multiple columns in production, and in a Modal, but not in the Activity Preview itself (although sometimes even there). I think this might have something to do with how the learning items themselves are formatted, and something about how we do the CSS on the Preview page.

Firstly some items don't look so nice yet - for example images inside a frame. Secondly, if you go to the example Hypothesis LIs on the Activity Preview
screen shot 2018-07-17 at 11 34 34

If you then click on the + sign on the right, and add an idea, this idea becomes hidden under the top left item. 

screen shot 2018-07-17 at 11 34 34

Also, the commenting box looks horrible (it's superimposed over the zoomed in view), and the button to submit also looks bad.

screen shot 2018-07-17 at 11 33 46

@anujanegi This would be a great thing to help us with!

anujanegi:
Fixed look of images inside a frame, and comment box (including submit box).

### DIFF
--- a/ac/ac-gallery/src/components/ImageBox.js
+++ b/ac/ac-gallery/src/components/ImageBox.js
@@ -6,11 +6,7 @@ import { withStyles } from '@material-ui/core/styles';
 
 const styles = () => ({
   root: {
-    marginLeft: 20,
-    paddingLeft: 16,
-    paddingRight: 24,
-    paddingTop: 20,
-    paddingBottom: 16
+    marginLeft: 20
   }
 });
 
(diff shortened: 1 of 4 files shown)


houshuang:
Hi. The comment box is nice, but look at Hypothesis. This is how it used to look:

screen shot 2018-08-14 at 15 56 42

and now it looks like this:

screen shot 2018-08-14 at 15 56 38

I think the former looks nicer...

anujanegi:
Updated.

houshuang (approved):

PR #1269, opened 2018-08-21
Multi cloud improvements

### LINKED ISSUE #1264: multiTeacher cloud issues
- in graph cloud list, too many undefineds :)
- when I import a graph, and then export it, it should already know it's information (title etc), and it should give me the option to just save if I'm the owner
- when I preview an activity that I have sent to the cloud in activity creator, I should have the option to toggle public/non-public.
- reactive data for an imported activity doesn't work (Stian) -- fixed

romainAA:
Closes #1264


houshuang:
@romainAA 
Issues:
- when I first export a graph as private, and then make it public, although it stays as public in FROG, it doesn't seem to be written to database - because I cannot see it when logged in as another teacher. If I initially export as public, it works. For activities, this works perfectly.
- if I export an activity as public, and then import it as another teacher, and want to export it again, it is marked public, but it should be marked default as private.

houshuang:
@romainAA 
When trying to export an existing graph it crashes

screen shot 2018-08-23 at 17 06 37

houshuang:
@romainAA Can you fix this so we can get it merged?

houshuang (changes requested):
Not working

houshuang:
@romainAA 
Still some problems:

- graph import/export/public etc etc seems to work perfectly
- if I export an activity from the sidebar as private, and then export again, change to public and "overwrite", it isn't stored
- if I open a public activity from the library in the activity creator as another user, and modify it, I can save over the original activity!

houshuang:
@romainAA 

I export a graph to the server, as private. I change it to public, and overwrite. I can now see it as another user. I try to "export graph as server" again - the public field is unchecked (even though the graph is public - I can see it as the other teacher)...

I import a public activity from another teacher in the side panel, and try to export it - it is marked as public by default.

When I first open the library in the activity creator, the activity that I just exported from the side panel is marked as Owner: undefined. Once I click on it, it changes to Owner: peter (and let's me edit).

Can you please just go through all the cases and make sure it works, I'm getting a bit tired of this PR... (don't work on this before the meeting with Valerie)

PR #1289, opened 2018-08-29
Changes to lock files when running a clean install

lfaucon:
When opening the app which works fine, this appears in the browser console:
screen shot 2018-08-29 at 10 51 33

Maybe your local installation and the package.json are slightly out of sync @houshuang

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "yarn": {
-      "version": "1.7.0",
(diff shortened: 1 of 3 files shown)


houshuang:
The package files etc shouldn't change, however the warning I cannot do anything about - Jest requires this version, and Meteor falsely discovers it (it's not actual the one that gets bundled). Now that Babel 7.0 has finally been released, it will hopefully stabilize quite soon, once Jest and Meteor gets updated.

houshuang:
See if you can figure out a way for continuous integration to see if any files have changed during setup? Maybe a git status or something?

PR #1297, opened 2018-08-29
Add test for lock files in CI

lfaucon:
Add a test that fails if the git status contains the name of one of the lock files.


lfaucon:
Not working. Somehow the repo is not a git repository

PR #1318, opened 2018-09-03
Improvements to ac-stat

lfaucon:
- Add css in the right place to have the activity look more professional and intuitive
- Fix bug related to sorting
- Refactor the code to be more readable (separating large component into smaller ones/naming functions/...)


houshuang:
Would be easier to test if you could quickly summarize what has been improved

houshuang:
There are still some CSS issues:
- overlap with "select zone to zoom", graph selector, and "use 2nd column"
- sometimes delete action is hidden
- sometimes graphs are not properly zoomed in boxes, even when I made the preview full screen with two graphs

screen shot 2018-09-03 at 15 16 12
screen shot 2018-09-03 at 15 15 44
screen shot 2018-09-03 at 15 14 37
screen shot 2018-09-03 at 15 14 33

Also what does m mean? 1000? is this logical/well known to the students?
screen shot 2018-09-03 at 15 19 01

houshuang (dismissed):
CSS issues, overlap, see comment

lfaucon:
@houshuang You were too quick to review, I was not done yet ^^

- overlap is fixed by deleting the two lines of text. If you think they are helpful, I suggest adding back as a toolkit behind a question mark icon. This could be helpful also to explain what the transformations are/do
- you should be able to scroll to bring the delete action into view. Your screenshots do not allow me to see if the scroll bar at the bottom of the table is missing 
- size of graph should be correctly handled by the latest commit 
- m stands for one thousandth, this is well known normally. I used the library human-format to display numbers more nicely. We can find another one if you prefer.

houshuang:
Looks much nicer now.

the m doesn't make sense to me.

log(1.5) = 0.4054651081081644, however according to the table, it's 405.47 m

lfaucon:
> m stands for one thousandth, this is well known normally. I used the library human-format to display numbers more nicely. We can find another one if you prefer.

Indeed 0.405 = 405 thousandth 

Anyway, I changed it with my home made display function which I believe will be convenient in all cases.

PR #1321, opened 2018-09-04
Add experimental tag in activity package meta

lfaucon:
Activity packages can add the tag `experimental: true` in the meta data.

By default the experimental activities are hidden for users. To change it you must use the following commands in the console and refresh the page

To show the experimental activities
```
localStorage.setItem('isResearcher', 'whatever.not.empty')
```

To hide the experimental activities
```
localStorage.removeItem('isResearcher')
```

This could be a button in the top right gear-menu


houshuang:
First of all, it's funny, because I could have sworn that I already did this - together with a .frog config file, where you could permanently configure that you wanted to see experimental activities or not. Maybe it's on a branch that never got merged...

Secondly, I have a bit of a different idea about this. Here's what I wrote to Anuja:
> here’s another design challenge, which maybe is a little bit easier in terms of implementation… in the graph editor, you see the sidebar where we list activities and operators. Right now we’re starting to get a lot of activity types, and it’s very overwhelming for new teachers - especially since some are very general, and some are very specific….

> What I’d like is some kind of categories, like “General tools”, “Simulations”, “Experimental”. I’m not sure how this would look visually, I’m imagining something like an accordion that can be open and closed, the default would be that teachers only see “General tools”, but they should see that there are other categories, and be easily able to also browse Simulations etc. 

> Do you think you want to try this? Feel free to show me some mockups before you begin actually implementing. Actually changing the code should be much easier, since all the code for rendering that sidepanel is in the same file. (I will need to add categories to each individual activity type, but I will do that).

lfaucon:
I agree this is not a long term optimal solution, but it's a quick win, which is important to us before we send out the platform to CS411 students

PR #1322, opened 2018-09-05
Visual upgrade on preview

lfaucon:
The size of the box for the control panel and the content of the preview was often wrong depending on screen size.

For example: 
- Too much margin on the right:
screenshot from 2018-09-04 16-07-02
- Right side out of the window and inaccessible
screenshot from 2018-09-04 16-08-19

This PR fixes this.

Also:
- Removes bootstrap from the control panel
- Changes ugly list of button to a mui-menu


houshuang:
Thanks for working on this, great to get it fixed.

However, I'm not crazy about the pull-down menu. I often switch back and forth between different examples a lot, and this is now a lot slower. I also worry that it's a lot less easy to discover for new users - for whom the preview is one of the main introductions to FROG. But I can play with it to see how I put back the whole list of examples.

Also, when I first load the example logs dashboard, there is a bit of horizontal scrolling - probably just need to remove some width or CSS from that component.

lfaucon:
"A lot slower" is exaggerated, maybe it takes now 1.5 the time it used to take. A tradeoff the the saved space (useful when simulating 4 students). If cycling through examples quickly is very important we can make it a keyboard shortcut.

PR #1330, opened 2018-09-10
Add coordinates to ck-board to be used with argue-operator

lfaucon:
The changes in `op/op-argue` are simply copy pasting the more robust version of the algorithm implemented in `op/op-argue-constraints`.

Added a graph on the server to test


lfaucon:
I fixed the test that we already had. The new algorithm gave the same output groups except for sometimes the order of the groups, so I made it such that the test compares only the sorted list of output groups

PR #1335, opened 2018-09-12
Fix broken merge function

### LINKED ISSUE #1304: Third example  of ac-plot exampleData is broken

lfaucon:
Closes #1304

### DIFF
--- a/ac/ac-plot/src/index.js
+++ b/ac/ac-plot/src/index.js
@@ -34,6 +34,10 @@ const config = {
 const dataStructure = {};
 
 const mergeFunction = ({ data: incomingData }, dataFn, data) => {
+  if (!incomingData) {
+    return;
+  }
+
   const toInsert = {};
   const prepareMergeItem = (item: { trace: string, x: number, y: number }) => {
     if (item && !isEmpty(item) && isObject(item)) {
--- a/ac/ac-plot/src/index.js
+++ b/ac/ac-plot/src/index.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { type ActivityPackageT } from 'frog-utils';
+import { type ActivityPackageT, entries } from 'frog-utils';
 import { isEmpty, isObject } from 'lodash';
 
 import meta from './meta';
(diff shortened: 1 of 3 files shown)


ac/ac-plot/src/index.js
```diff
@@ -33,32 +33,36 @@ const config = {
...
 
-const mergeFunction = ({ data: incoming }, dataFn, data) => {
-  if (Array.isArray(incoming)) {
-    incoming.forEach(item => mergeFunction({ data: item }, dataFn, data));
-    return;
-  }
-  if (isEmpty(incoming) || !isObject(incoming)) {
-    return;
-  }
-  if (incoming['1']) {
```
lfaucon: This is weird, did you write this special case to accept incomingData with the format:
`{  '1': { x, y, trace } }` ?
That looks arbitrary and unnatural to me. 
Made it so that the accepted types are only  `{x, y, trace}` or  `Array<{x, y, trace}>`

ac/ac-plot/src/index.js
```diff
@@ -33,32 +33,36 @@ const config = {
...
-    mergeFunction({ data: incoming['1'] }, dataFn, data);
-  }
-  if (incoming.y) {
-    if (!data[incoming.trace]) {
-      dataFn.objInsert({ y: [] }, incoming.trace);
```
lfaucon, line 49: I believe this was a bug. 
Does the `data` object changes synchronously when a dataFn function is called? I think it does not. This would then create bugs if we insert several points as we reinitialize the dataset for each point.

houshuang (changes requested):
Does not work with streaming data from WISE. Probably something about how you're handling wrapped objects (taking away the 1: stuff).

Please test with graph WISE Simulation Demo.

If you think there's a good reason, you could suggest a change to the format of the data streamed by thermoCUp (although note that it's also compatible with ac-spreadsheet, which is currently working), or even with how the streaming function packages stuff (although again - a lot of activities are relying on the current way of working).

lfaucon:
The problem was that I was expecting input of type `{trace, x, y}` but WISE simulation only sends `{trace, y}` expecting that by default x takes values 1,2,3,4,5...

PR #1339, opened 2018-09-12
Changes usage of arrays to object to fix bug in quiz dashboard

lfaucon:
I do not know what was the problem.
Before the state was an array and the viewer was expecting arrays (it worked perfectly in preview)
Now the state is an object and the viewer expect an object


houshuang:
Doesn't seem to be necessary anymore, sorry for the extra work. At least we found a good bug and squashed it in general.

PR #1342, opened 2018-09-13
Small css refactor

lfaucon:


houshuang:
Does not work with streaming data from WISE. Probably something about how you're handling wrapped objects (taking away the 1: stuff). 

Please test with graph WISE Simulation Demo.

If you think there's a good reason, you could suggest a change to the format of the data streamed by thermoCUp (although note that it's also compatible with ac-spreadsheet, which is currently working), or even with how the streaming function packages stuff (although again - a lot of activities are relying on the current way of working).

houshuang (changes requested):
Not working with ac-thermoCup

PR #1346, opened 2018-09-14
Small fixes around restart session/no activity

### LINKED ISSUE #1344: Restarting graph should remove all "joinedSession"
To avoid ghost users in social operators

### LINKED ISSUE #1345: Show "Teacher has not started session"
Or waiting for teacher to start session if timeInGraph === 0

houshuang:
The messages when there is no activity are now different whether Teacher has not yet started graph, No activity, or Graph is finished.

(in the future, I'd like to make this prettier. Maybe just a very big icon (Starting, finished) in a soft color, together with the text. 

We could also change the text of No activity to Projector mode (with a big picture of Projector), since that's now the only way we cannot have an activity (given that we now skip empty spaces).

We should also add nicer formatting (similar) to the Pause view.

Also added a separate menu item to remove all students from current graph. This will make all currently logged in student screens spin, and they have to be reloaded (reason I did not enable this by default for restart session). 

Closes #1345
Closes #1344

### DIFF
--- a/frog/imports/ui/StudentView/SessionBody.jsx
+++ b/frog/imports/ui/StudentView/SessionBody.jsx
@@ -109,7 +109,15 @@ const StudentView = ({ activities, session, classes }) => (
     <div className={classes.mainContent}>
       {(() => {
         if (!activities || activities.length === 0) {
-          return <h1>No Activity right now</h1>;
+          if (session.state === 'READY' || session.state === 'CREATED') {
+            return <h1>Waiting for teacher to start the session</h1>;
+          }
+          if (session.state === 'STARTED') {
+            return <h1>No activity right now</h1>;
+          }
+          if (session.state === 'FINISHED') {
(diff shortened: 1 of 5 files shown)


frog/imports/ui/StudentView/SessionBody.jsx
```diff
@@ -109,7 +109,15 @@ const StudentView = ({ activities, session, classes }) => (
     <div className={classes.mainContent}>
       {(() => {
         if (!activities || activities.length === 0) {
-          return <h1>No Activity right now</h1>;
+          if (session.state === 'READY' || session.state === 'CREATED') {
+            return <h1>Waiting for teacher to start the session</h1>;
+          }
+          if (session.state === 'STARTED') {
+            return <h1>No activity right now</h1>;
```
lfaucon, line 116: Could change that text to "Activity with the professor"

lfaucon (approved):

PR #1348, opened 2018-09-16
Several fixes around graph editor

lfaucon:
The problem with Undo was that the constructor for the activities in graphEditor was changed so it was not being used correctly and it was sending new activity ids for all activities. That lead to deleting everything and recreating the activities (losing activityTypes and config in the process). As this could have been caught by flow, I also made the effort of adding flow everywhere. 

Fixed the GraphName component in the process.

Adding back the HelpModal. The ChangeLog is accessible with "!"

### DIFF
--- a/frog-utils/src/TextInput.js
+++ b/frog-utils/src/TextInput.js
@@ -9,7 +9,7 @@ type TextInputPropsT = {
   onSubmit?: Function,
   onCancel?: Function,
   onFocus?: Function,
-  style?: string
+  style?: Object
 };
 
 export class TextInput extends React.Component<
@@ -84,7 +84,7 @@ export class TextInput extends React.Component<
 export class ChangeableText extends React.Component<
   TextInputPropsT & {
     EditComponent?: React.ComponentType<*>,
-    onlyHover?: Boolean
+    onlyHover?: boolean
   },
   {
     edit: boolean,
@@ -94,7 +94,7 @@ export class ChangeableText extends React.Component<
   constructor(
     props: TextInputPropsT & {
       EditComponent?: React.ComponentType<*>,
-      onlyHover?: Boolean
+      onlyHover?: boolean
     }
   ) {
     super(props);
--- a/frog/imports/api/graphs.js
+++ b/frog/imports/api/graphs.js
@@ -164,7 +164,7 @@ export const setCurrentGraph = (graphId: string) => {
   });
 };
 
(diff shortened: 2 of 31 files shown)


houshuang:
Thank you so much for taking the hard work to clean up a bit of this mess... I know that we (I :)) were quite proud of the flow typing in Store, with the way we modeled state etc, however when we had to remodel how we did mobx because changes in Babel etc, I had to create all those definitions at the top of the file, and I hated the idea of it, and never got around to it. It's actually possible that Mobx has been updated to allow us to do it in the old way now, but I'll leave that for the next refactor :)

It's also great that we finally fixed the graph name selector!! I pushed a commit which also ensures that if you import a graph from the server, it will be named after the server name, and not whatever internal name it had when it was exported.

One thing about Flow, if you define a function, then a(x: ?string) means that x can be a string or null, whereas a(x?: string) means that x should be a string, but there might not be an x (ie undefined). This makes a difference, at least I got errors saying "x needs an argument" when it was defined in your way, until I changed it some places.

Thanks a lot for putting the help modal back, although the title wasn't showing. I copied some style code from the changelog modal, and now it works... I want to factor out all the modal code into a FrogModal component, generally we should just provide title and contents and not all this copied+pasted stuff, but not right now.

--

One problem with making the selected activities look "nicer" is that you now cannot see if a selected activity has an error. Here's the difference between a selected activity with an error and one without in the current setup:

screen shot 2018-09-16 at 22 15 12
screen shot 2018-09-16 at 22 15 10

Not perfect, and I'd love something clearer/nicer, but I don't really want to totally loose this information.

--

Sometimes when I undo, I get connections connected to nothing, but still rendered. This is really weird - never seen that before.

screen shot 2018-09-16 at 22 17 10
 
screen shot 2018-09-16 at 21 57 04

Actually I guess the reason I've not seen it before, is because you put in a failsafe when rendering connections - before it would have just failed? But either way this shouldn't be happening...

--

The third mode in organize next is broken. If you click z three times, it looks like this

screen shot 2018-09-16 at 22 19 13

This is from the organize function in activityStore, which you modified. I guess nobody has ever really used this functionality, but still would be nice to have it unbroken.

houshuang (dismissed):
comment

lfaucon:
Thank for catching a few bugs that I created. It's lucky that we test things differently.

Ready for second round of tests

I have also change a few default values, tell me if you don't like it.

PR #1356, opened 2018-09-17
Used Material UI for the preview control options

tusharpahuja:
Phased out Bootstrap and used Material UI for the preview control options.

### DIFF
--- a/frog/imports/ui/Preview/Controls.jsx
+++ b/frog/imports/ui/Preview/Controls.jsx
@@ -1,10 +1,8 @@
 // @flow
 
 import * as React from 'react';
-import Paper from '@material-ui/core/Paper';
-import { withStyles } from '@material-ui/core/styles';
-import CloseIcon from '@material-ui/icons/Close';
-import Button from '@material-ui/core/Button';
+
+import { Button, List, ListItem, Paper, Grid } from '@material-ui/core';
 
 import { uuid } from 'frog-utils';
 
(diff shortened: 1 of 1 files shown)


frog/imports/ui/Preview/Controls.jsx
```diff
@@ -18,26 +16,6 @@ import { initActivityDocuments } from './Content';
...
-    color: 'white'
-  },
-  exampleButton: {
-    textTransform: 'none'
-  },
-  exampleButtonDeeplink: {
-    textTransform: 'none',
-    fontStyle: 'italic'
-  }
-});
```
lfaucon, line 39: We have decided to use css through material-ui's `withStyle` for the whole project. We gradually remove all of the inline css to be replaced by withStyle. The advantages I see are that it is easier to work with a single way, over the whole project instead of having several ways of doing it, also using `withStyles` makes the component easier to read. I prefer to see "This component has the css called `closeButton`" rather than "This component has the css `{ float: 'right', marginRight: '5px', backgroundColor: 'red', color: 'white' }`" when I read the code of a component.

lfaucon (commented):
Hello @tusharpahuja ,

Thank you for your contribution, we really appreciate it. Sorry to answer you late, we have been rushing a lot of things recently as it is the beginning of the school year and FROG is being used already in several classrooms all over the world :) 

Before you submitted this PR, Stian and I went through the same file (`.../Preview/Controls.js`) to fix some bugs and we also got rid of got rid of Bootstrap in it. I left a comment on the code, but actually I think the diff does not show the actual modifications that you have done. 

I would gladly discuss with you either by chat or videocall if you want to plan your future contributions. Send me an email

frog/imports/ui/Preview/Controls.jsx
tusharpahuja: Makes sense to me. Thanks !

PR #1418, opened 2018-10-02
Making sure social operators always fulfil their output contract

houshuang:
I found a bug where a graph had a op-social-name operator, feeding into an aggregate-p2 operator. The user had forgotten to configure the username list in the session settings, and ran the graph with no logged in users. This meant that the output from op-social-name was empty (no group: {}, but just {}), and this crashed aggregate-p2. I believe that if an operator states, through it's outputDefinition, that it will produce a social operator with a certain key, then it should always do so. I've gone through all the social operators and added tests for this condition.

(Note, of course in the case above, the graph would not be very successful because no data would be distributed etc, but at least we don't want runDataflow to crash).


op/op-create-groups/src/operatorRunner.js
```diff
@@ -24,7 +32,7 @@ const operator = (configData, object) => {
     struct = splitArray(ids, globalnum);
   } else {
     if (ids && configData.groupsize >= ids.length) {
-      return { group: { '1': ids } };
+      return { [newGrouping]: { '1': ids } };
```
houshuang, line 35: This was actually a bug

PR #1428, opened 2018-10-03
Op dataset

lfaucon:
Sorry, I branched from the wrong branch, but it should be correctly merged.

I updated the format product of ac-quiz to work with the new format of the reactive data (since we allowed multiple select). All the test fail because we were still testing on old format.


ac/ac-stat/src/Components/FilteringPanel.js
```diff
@@ -25,7 +25,7 @@ const styles = () => ({
 const transfo = ['log', 'exp', 'outliers', 'sqrt', 'x100', '+50', '11x-10E[x]'];
 
 const disabledFun = (data, tr) => {
-  if (!data.values) return true;
+  if (!data || !data.values) return true;
```
houshuang, line 28: You can also do !data?.values :)

houshuang:
If we show the two data sets together, we get this:

screen shot 2018-10-03 at 11 25 27

I guess the point was that we'd have one variable on the x-axis and the other on the y-axis, but this requires some transformation, especially since we are inputting months as a string...

Maybe the operator should have a text field with transformations like:
January,1
February,2

or we hard-code - I'd like to make it flexible, but not if it's very improbable that people will use it again?

houshuang:
Would also be nice if we could name the datasets (birth month, gender)

lfaucon:
@houshuang You can review it again

houshuang:
@lfaucon Now you are turning the month names into numbers, and I get 1 for October and 2 for November, I don't think that's what we wanted? I tried sending both answers, and it looks like it works OK at least for barplot (the only one meaningful)?

houshuang:
@lfaucon 	Merged this, doesn't mean we can't make it better !:)

lfaucon:
Yes, sorry, you are right, I got mixed up between different states of my code

PR #1429, opened 2018-10-03
Fix numbering of old sessions

### LINKED ISSUE #1417: Many sessions have same SLUG
This should not be possible (on icchilisrv3)

screen shot 2018-10-02 at 04 44 54

lfaucon:
Closes #1417

### DIFF
--- a/frog/imports/api/sessions.js
+++ b/frog/imports/api/sessions.js
@@ -43,9 +43,7 @@ Meteor.methods({
       let nextNum = 1;
       if (prev && prev.length > 0) {
         const findIndex = s => parseInt(s.slug.split('-').pop(), 10);
-        console.log(prev.map(findIndex));
         const prevNum = Math.max(...prev.map(findIndex));
-        console.log(prevNum);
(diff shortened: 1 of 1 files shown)


frog/imports/api/sessions.js
```diff
@@ -42,14 +42,9 @@ Meteor.methods({
       }).fetch();
       let nextNum = 1;
       if (prev && prev.length > 0) {
-        prev.sort((x, y) => y.slug - x.slug);
```
lfaucon, line 45: Here you sort alphabetically, that must have caused problems with "10" being less that "2"

houshuang:
Nice, but I think this would make more sense (need to test it first):

```js
        const findIndex = s => parseInt(s.slug.split('-').pop(), 10);
        const prevNum = Math.max(...prev.map(findIndex));
        nextNum = prevNum + 1;
```

houshuang:
@lfaucon Just made a small change, tested it, see if you approve before we merge.

lfaucon:
Should be good

PR #1434, opened 2018-10-05
Making ac-stats more configurable

houshuang:


lfaucon:
Looks like there is a problem with the configuration for "Kind of plot to display". I think it says bar instead of histogram. 
Also the preview shows only an empty graph for the example you added, until I change a few things, then it display the correct data.

PR #1446, opened 2018-10-08
Aggregate multiple

lfaucon:


op/op-aggregate-p2/src/operatorRunner.js
```diff
@@ -1,29 +1,38 @@
...
-    res[group] = [
-      ...(res[group] || []),
-      ...(configData.wholeElement ? [items] : items)
-    ];
+  const res = groups.reduce((acc, g) => {
+    acc[g] = [];
+    return acc;
+  }, {});
+
+  const hasMultipleIn = !activityData.payload;
```
lfaucon, line 13: Is it the right way to check for mutiple inputs?

PR #1462, opened 2018-10-09
Fix config of ac-stat

lfaucon:


lfaucon:
Fixed: 
- Works when only one plot is selected
- Does not default to showing histogram at first when histogram is not selected

PR #1476, opened 2018-10-11
Fix sorting acstat

lfaucon:


jkolsen (approved):
The sorting is fixed for the months and did not impact data that did not have sort selected.

PR #1489, opened 2018-10-16
Fix CSS for progress dashboard

houshuang:


frog-utils/src/dashboards/progress.js
```diff
@@ -18,7 +18,10 @@ const Viewer = (props: Object) => {
   const nowLine = [{ x: state.now, y: 0 }, { x: state.now, y: 1 }];
   return (
     <React.Fragment>
-      <VictoryChart theme={VictoryTheme.material}>
+      <VictoryChart
+        style={{ height: 'calc(100vh - 50px)' }}
```
lfaucon: This seems weird to me. I thought the unit `vh` was for the full height of the window, but here we would want the graph to be sized relatively to its parent container. I pulled the code and I don't see what changed from before. Maybe the component does not understand the style as it is specified. 

You could try putting the victory chart component in a div for which you control the size. 
What I like to do is the property `flex: 1` if the container has `display: "flex"` and contains other elements of fixed size. `flex: 1` makes your component take all the remaining space.

houshuang:
@lfaucon 	See the difference below. I'm merging this for now since it took me a bit of fiddling to get it to work, and I've tested in all views that it works. Happy for someone to keep improving it.

screen shot 2018-10-18 at 12 45 18
screen shot 2018-10-18 at 12 42 50

PR #1499, opened 2018-10-19
RichText Learning Item

Ruchiranga:
The `ReactiveRichText` component provides collaborative RichText editing by integrating `react-quill`  RichText editor with `ShareDB`. This component accepts an optional `toolbarOptions` prop that allows customizing the toolbar options shown to the user in the editor.

The `li-richText` learning item contains a `ReactiveRichText` component.


frog-utils/src/ReactiveRichText.js
```diff
@@ -0,0 +1,118 @@
...
+
+  render() {
+    return (
+      <div>
+        <ReactQuill
+          defaultValue={get(this.props.dataFn.doc.data, `payload.${this.props.path}`)}
+          ref={element => { this.quillRef = element }}
+          readOnly={this.props.readOnly}
+          onChange={this.handleChange}
+          modules={this.modules()}
```
houshuang: Here you are running the modules function every time the component is re-rendered, which might be with every keypress... Since it never changes, you should instead just define a static const outside of the class.

frog-utils/src/ReactiveRichText.js
```diff
@@ -0,0 +1,118 @@
...
+      [{ 'header': [1, 2, 3, 4, 5, 6, false] }],
+
+      [{ 'color': [] }, { 'background': [] }],
+      [{ 'font': [] }],
+      [{ 'align': [] }],
+
+      ['clean']
+    ];
+    return {
+      toolbar: this.props.readOnly ? null : toolbarOptions
```
houshuang: Define toolbarOptions externally, and then do this ternary inside the render function.

frog-utils/src/ReactiveRichText.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+  };
+
+  opListener = (op, source) => {
+    if (source === this.quillRef) return;
```
houshuang: I prefer { } even around single statements, like { return } - makes it easier to read.

frog-utils/src/ReactiveRichText.js
```diff
@@ -0,0 +1,118 @@
...
+    if (source !== 'user') return;
+
+    this.props.dataFn.doc.submitOp([op], {source: this.quillRef});
+  };
+
+  render() {
+    return (
+      <div>
+        <ReactQuill
+          defaultValue={get(this.props.dataFn.doc.data, `payload.${this.props.path}`)}
```
houshuang: I'm trying to figure out why you need to specify payload here, but this is never done in ReactiveText or li-textarea?

houshuang (changes requested):
Hi, great milestone, looks very good!

In addition to some comments, questions in the code, some general observations:

- it does not work well with two components on the same page. I pushed a commit where I added another RichText to the Learning Item, and it does not work well. This is important.

- We'll probably reduce the number of buttons in the toolbar, but we can look at it together to figure out which are more important etc

- Make sure to get it to green in the CI - ask me if there's anything you don't understand

- There's a few additional things that would be very nice to expect from the Learning Item. One is search - the ability to pass a search term, and only render if that search term is present, and even highlight that search term. The other is shortening, to only display a subset of the text in thumbnail mode.

I think it would be much easier to do these operations on HTML than on deltas (we already have functions for all of this - and do it on for example li-hypothesis, in the op-hypothesis package). I see that you can use Quill.root.innerHTML to get the HTML from a document. I'm not sure what's more efficient - keeping a separate HTML representation of the document up to date in the reactive text, or having the viewer keep rendering the document in a hidden Quill and then getting the HTML and rendering that, or maybe there's a much better way. Let's discuss.

But anyway, the only thing I need to merge this for now is green CI, and multiple components on the same page working.

frog-utils/src/ReactiveRichText.js
Ruchiranga: sure, makes sense. will do that.

frog-utils/src/ReactiveRichText.js
Ruchiranga: Yeah. In the `update` method of the `ReactiveText` component it calls `bindTextField()` method in `generateReactiveFn.js` with `this.path` which in our case evaluates to `'text'`. And in `generateReactiveFn.js` within the Doc component, `this.path` contains the string `'payload'`. When the string binding is created in `bindTextField()` method, the `path` param passed into it is a combination of both those values. 

But since in the Richtext implementation we do not call any method in `generateReactiveFn.js` I could not find a way to get the `'payload'` string dynamically. Please let me know if there is a better way to do that.

frog-utils/src/ReactiveRichText.js
Ruchiranga: okay. will do that.

frog-utils/src/ReactiveRichText.js
Ruchiranga: okay.

frog-utils/src/ReactiveRichText.js
Ruchiranga: yeah it does. Will change that.

Ruchiranga:
Hi,

Thanks for the comments and the suggestions. I will look into the improvements and update the PR.

Ruchiranga:
@houshuang I have updated the PR with the required fixes for supporting two components on the same page along with fixes for build errors. 

Also, incase you didn't notice, please let me know what your input is on my [response](https://github.com/chili-epfl/FROG/pull/1499#discussion_r227596585) regarding using `'payload'` in the `ReactiveRichText` component.

PR #1540, opened 2018-11-21
Add prisoner's dilemma activity

lromerio:
Implements a new activity called `ac-prisoner-dilemma` and its dashboard.

As suggested by its name, it allows to play the prisoner's dilemma. The number of rounds can be customized, while the gain matrix is always the one below. In case of a group composed by more than two people, additional members will be spectators.

|           | Cooperate | Cheat |
|-----------|:---------:|:-----:|
| **Cooperate** |    2/2    |  -1/3 |
|   **Cheat**  |    3/-1   |  0/0  |

All test are passing locally (`npm start test`).


houshuang:
See my PR and please improve it according to my instructions @lromerio

PR #1547, opened 2018-11-28
Conditional dashboards

houshuang:
The dashboard object takes an optional value called displayCondition, which is similar to how configUI works - it can either take a string, the key of a config field which is turned into a boolean check, or a function which takes formdata as input and returns true/false.

Currently I've only made the coordinates dashboard for ac-quiz dependent on whether coordinates are added to questions. We might also check if the right answer has been added, for the Leaderboard, but I guess maybe sometimes we use it just to see who finishes first? Also have to look at the other dashboards to see if there are other things we can make conditional.

(The key objective is to avoid confusion among teachers).

Since we're only modifying the list of dashboards shown, but none of the logic about subscriptions etc, it shouldn't introduce any new bugs :)

### DIFF
--- a/frog/imports/ui/Dashboard/MultiWrapper.jsx
+++ b/frog/imports/ui/Dashboard/MultiWrapper.jsx
@@ -110,7 +110,8 @@ const MultiWrapper = (props: {
   dashboardData?: Object,
   session?: Object,
   object: Object,
-  ready?: boolean
+  ready?: boolean,
+  config?: Object
 }) => {
   const {
     dashboardData,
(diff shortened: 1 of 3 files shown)


frog/imports/ui/Dashboard/MultiWrapper.jsx
```diff
@@ -127,7 +128,24 @@ const MultiWrapper = (props: {
...
+    names ||
+    Object.keys(aTDash).filter(name => {
+      const cond = aTDash[name].displayCondition;
+      if (!cond) {
+        return true;
+      }
+      if (typeof cond === 'string') {
+        return !!activity.data[cond];
+      }
+      return cond(activity.data);
```
lfaucon, line 146: Do you want to add a try/catch around this? As it's code from an external activity package it will not always be implemented by one of us and we don't want it to crash the whole app.

lfaucon (approved):

frog/imports/ui/Dashboard/MultiWrapper.jsx
houshuang: We can't put try catch everywhere, and this is just based on config, so it should be very easy to test in preview. We don't do try catch for configUI, which is much more common.

lfaucon: 👌

PR #1563, opened 2018-12-10
Fix style for quiz choices

lfaucon:
The choices for MCQ where not aligned properly (I think since we upgraded MUI)


houshuang:
Added all these 3 into my larger PR.

PR #1572, opened 2018-12-16
Shifting the action buttons to the bottom

ajainuary:
In order to make the brainstorm activity look less busy (as suggested by @houshuang ) I've shifted the action buttons to bottom of content so that they don't compete for space. Also increased padding and white space to make it look neater.
Shifting it to bottom instead of keeping it on the top also reduces movement as the user goes through the contents first, then performs action instead of doing the action first.


houshuang:
Hey, great start!

Looks what happens if I zoom in one item:
screen shot 2018-12-17 at 06 59 39

Your proposal also hurts scannability quite a lot, especially notable with shorter items, compare

screen shot 2018-12-17 at 07 01 19
screen shot 2018-12-17 at 07 00 59

(PS: There seems to be some bug with the config that it doesn't update automatically, I sometimes have to click in the side panel to have options appear, I'll look into that).

I'm wondering if we should have the zoom/edit/delete only appear on hover. The upvote/downvote should probably always be shown for discoverability... or maybe not. I also think we should choose nicer icons - like a cross for delete?

ajainuary:
How about I hide all the buttons by default (only when the user hovers above an Idea listing buttons are shown for that item)?

houshuang:
Try it! We need to kind of play with it to get an intuition.

ajainuary:
I have given the buttons a bit more minimalism (to fit them in less space) and made them appear only on hover. It is expected that buttons can be on top of content because buttons have no real estate of their own but share it with the content (but I've set the timeout so that it is not very obtrusive).

PR #1580, opened 2018-12-18
Import of Anurag's branch

houshuang:


houshuang:
Hey, thanks for all your effort. Hope you don't mind that this takes some time, sometimes we have to see things to really get a sense of it.

First of all, the way you do the mouse-over right now with a bit of lag after you remove the mouse does not look good - see what happens when I just quickly drag my mouse across the different items. I think you should change it to just immediately appear and disappear based on where the mouse is, with no lag etc. This is easy to do with a hover css property, you don't need any fancy Material UI interactions for that. 

screen shot 2018-12-18 at 16 28 41

Secondly, I think that voting up and down is a really key action, and should be there permanently. However, what if we had something like a an integrated counter and upvote/downvote, something like this?

screen shot 2018-12-18 at 16 33 41

Something simpler to implement would be to put them all on the same line I guess, downvote, number of votes, upvote... but it would take more space.

houshuang:
Note that I imported your branch, checkout this branch and continue work on it, then you get continuous integration. You'll see that there are a bunch of tests failing. Flow can be a bit tricky, ask me if you get stuck.

houshuang:
I turned off the delay, and it already became much more usable.

Note that there is still some CSS issue with different width items, see examples

screen shot 2018-12-18 at 16 42 19
screen shot 2018-12-18 at 16 42 16

actually I guess it's just because one is zoomable and one is not, I guess that's OK

ajainuary:
I could try to build a prototype with upvote/downvote button with integrated counter. In order to retain the same height (in order to maintain scannability) I would need to make the buttons of height 13px, which is 2/3rd of current size. Not only will they get small, they will be permanent (as the counter needs to be visible at all times) so again look would be busy. If that's fine can the counter be visible only on hover?

houshuang:
Hi,
here's how it looks (only on hover) in the rich-text (quill-embed-react-comp branch) which I think looks decent. We can look at whether the icons should be top/bottom etc.

screen shot 2018-12-22 at 14 33 57

What I'm planning to do is to extract this code (whichever part of yours and Ruchiranga's which works best) and make this available from core, so we won't have to render this from the ac-brainstorm. However the voting still needs to be rendered, and it should "work" with these... 

I'm not exactly sure what would look good. I do think it should always be visible, especially since you might have several with the same score etc. Can you try and show me a prototype, worrying just about the up/downvote and score?

Thanks!!

Issue #1574, opened 2018-12-17
ac-gallery layout


houshuang:
The adding Learning Item component is locked to the bottom of the page, even if the items themselves make the component scroll.

Either we should have it at the bottom no matter if it's initially hidden by scroll (or at top?), or it should be outside the scroll zone. 

<img width="695" alt="screen shot 2018-12-17 at 16 38 58" src="https://user-images.githubusercontent.com/61575/50097487-59c50480-021a-11e9-82bc-01ffa77547f2.png">

PR #1586, opened 2018-12-28
Issue #1574 - Shifted Add Learning Item component to top right 

RachitMalik12:
Hi, 
Based on the description of the issue, I shifted the Add Learning Item component to the top right of the screen as shown below, outside the scroll zone. I felt it was better than keeping it locked to the bottom right as it is more visible to the user and also it doesn't cover the content at some points on scrolling down.
screen shot 2018-12-28 at 8 23 26 pm

### DIFF
--- a/ac/ac-gallery/src/ActivityRunner.js
+++ b/ac/ac-gallery/src/ActivityRunner.js
@@ -138,7 +138,7 @@ class ActivityRunner extends Component<
           </div>
         )}
         {activityData.config.allowAny && (
-          <div style={{ position: 'absolute', bottom: '10px', right: '10px' }}>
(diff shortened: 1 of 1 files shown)


houshuang:
Does it also work if you specify a default Learning Item to add? It shouldn't be over the menu - look at how it looks if you make it fullscreen... 

I was thinking we could actually use some of the CSS from the chat activity, which has a specific learning item adder and an add button below the chat messages, with the chat messages scrolling freely.

houshuang:
Let me know if you can give this another go, @RachitMalik12

RachitMalik12:
> Let me know if you can give this another go, @RachitMalik12

Yes! I'm stuck on a couple of things, I have emailed you..

PR #1594, opened 2019-01-04
Updating all deps, and moving files around - major reorganization

houshuang:


lfaucon:
Testing with ac-single-li in preview the activity crashes with some li-types (files and images for example)

houshuang:
Great work! I started testing activities, and fixed a bunch of deprecations, and some libraries that changed their API... Including files and images :) Will continue looking through all the activities. Also know that I need to fix something with any operator that uses LearningItems. Hopefully can merge early next week.

PR #1609, opened 2019-01-10
Fixes CI errors

lfaucon:


lfaucon:
@houshuang what is this "normal" mode for Eslint? On my machine it runs forever and apparently in the CI too.

PR #1650, opened 2019-03-05
Reactive Rich Text Refactoring

dinomujki:
Overview:
- This PR aims to refactor the existing code for the reactive rich text functionality. 
- The main changes are the splitting of the previous monolith single file ReactiveRichText.js into multiple components contained in a folder of the same name within 'frog-utils'
- The importing of other supporting components is done through direct file imports since going through imports from 'frog-utils' caused errors due to the structure of the FROG app
- Several additions and modifications were done to the original code file, namely better type checking, optimization of redundant loop retrievals, and code clean-up

There are a few issues remaining from the original file that will be fixed in future commits:
- After initial load, the main component for reactive rich text editing seems to redraw in a poor manner which causes the cursor to disappear from the input as well as causing other issues such as invalid LI rendering and forcing closing of LI+ dropdown menu.
- The 'Idea with Attachment' LI remains broken and crashes the 'activity' if selected.
- There is still significant overhead in many cases of re-rendering and updating


houshuang:
Great work. It would be very nice if you could also have a chance to look into the reloading issue, it's quite disruptive for students. I wonder if it's related to any of the timeouts that Ruchiranga set.

PR #1675, opened 2019-05-08
Wiki

dinomujki:
PR relating to the recent bug fixes based on tasks from Asana and other styling/minor fixes that were in the pipeline.

### DIFF
--- a/frog-utils/package.json
+++ b/frog-utils/package.json
@@ -9,7 +9,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@houshuang/react-quill": "^1.3.1",
+    "@houshuang/react-quill": "^1.3.2",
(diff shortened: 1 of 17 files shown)


houshuang:
The modal etc for dealing with deleted pages looks great, however if I create a new link after a page has been deleted, it doesn't connect it to the deleted page (a new link created after deletion should work exactly the same as a link created before deletion). Otherwise you get very weird behaviour - like asking me to restore a page after it has already been created anew, creating two pages with the same name, etc.

Screenshot 2019-05-08 at 13 34 22
Screenshot 2019-05-08 at 13 34 18
Screenshot 2019-05-08 at 13 33 55

houshuang:
The slider CSS, and opening newly created page in edit mode work well. Deleted pages still need to be resolved. Anything else I should test?

PR #1680, opened 2019-05-13
Wiki New page modal

ajainuary:
This improves the design of the Modal to create a New Page in the wiki.
This also adds operator selector.


houshuang:
Great start! 

Page Title should autofocus when you open (goal is to be able to do ctrl+n, type something, enter - and you have a new page). 

I would like to make the tabs much more distinct, right now there's almost no visual difference between Settings and Page Title... Can we have a different background for the entire tab row, or a different background for the currently selected tab or what?

I would use the whitespace much more on the first tab - moving page title down a bit, and then the plane and checkboxes much more - have the page title "alone". Also why is there so much gap between the two checkboxes?

The mouseover stays with the item, not the mouse, when you scroll. Anyway this is less urgent, we should find a better solution for mouseover in general.

Screenshot 2019-05-13 at 11 06 03

ajainuary:
Fixed all CSS bugs, wired the Activity form (although some activities have error with delayed state updates)

- [ ]  Need to fix state updates for some activities (like Text Component)
- [ ] Need to wire up Operator Form and Social Plane selection

frog/imports/client/GraphEditor/SidePanel/OperatorPanel/EditOperator.jsx
```diff
@@ -94,6 +94,7 @@ export default ({
 }: StoreProp & {
   operator: Object
 }) => {
+  console.log('Edit');
```
houshuang: Make sure to clean up console.log before opening PR

houshuang (dismissed):
Until I get the continuous integration to work again, please run eslint etc locally. 

Screenshot 2019-05-19 at 23 36 44
Screenshot 2019-05-19 at 23 36 38

houshuang (changes requested):
Fix eslint (don't worry about the merge conflict)

ajainuary:
I have now fixed the linting errors.

PR #1688, opened 2019-05-20
Single Activities

ajainuary:
This is a basic version of Single Activities which can be accessed at `/single_activity`.


houshuang (approved):
I didn't review your code in Single_Activity carefully, but the functionality looks good for now, and since you're not touching any other files, I think this is OK to merge. THanks.

Issue #1691, opened 2019-05-22
Installer script doesn't acknowledge errors


ajainuary:
The installer script doesn't check the exit status of commands to present the success message which might be misleading.

- [x]  Check for exit code and if necessary present error message before terminating
- [ ]  Check for python version before installation (Required version is 2.x for `bcrypt`, using any other version will present an error)

PR #1692, opened 2019-05-22
Issue #1691 - Improves installation script

ajainuary:
As described in #1691 the PR includes some changes which allow for checking exit code before presenting the success message.

### DIFF
--- a/initial_setup.sh
+++ b/initial_setup.sh
@@ -6,20 +6,24 @@ if [ -d "frog/node_modules" ]; then
 fi
 shopt -s dotglob
 
-FROG="`pwd`"
 YARN_VERSION='1.15.2'
 if which yarn && [[ `yarn --version` == $YARN_VERSION ]]; then
     echo 'Using pre-installed global Yarn'; YARN=yarn
 else
-    if [ -f "$FROG/node_modules/.bin/yarn" ] && [[ `"$FROG/node_modules/.bin/yarn" --version` == $YARN_VERSION ]]; then
-        echo 'Using pre-installed local Yarn'; YARN="$FROG/node_modules/.bin/yarn"
+    if [ -f "./node_modules/.bin/yarn" ] && [[ `./node_modules/.bin/yarn --version` == $YARN_VERSION ]]; then
+        echo 'Using pre-installed local Yarn'; YARN="./node_modules/.bin/yarn"
     else
-        echo 'Installing Yarn'; npm install yarn@$YARN_VERSION --no-package-lock && YARN="$FROG/node_modules/.bin/yarn"
+        echo 'Installing Yarn'; npm install yarn@$YARN_VERSION --no-package-lock && YARN="./node_modules/.bin/yarn"
     fi
 fi
 echo "Yarn: $YARN"
 
-"$YARN" install
+if ! "$YARN" install; then
+	echo "ERROR DETECTED, Removing installed node_modules"
+	git clean -fdx
+	echo "There was some error installing the required packages, please rectify the errors and try again"
+	exit 1
+fi
 rm -rf frog/node_modules
 ln -s `pwd`/node_modules frog
 ln -s frog/babel.config.js .
--- a/initial_setup.sh
+++ b/initial_setup.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 if [ -d "frog/node_modules" ]; then
   echo "Please run git clean -fdx before running initial_setup.sh"
   exit 0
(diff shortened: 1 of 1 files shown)


initial_setup.sh
```diff
@@ -1,32 +1,44 @@
-#!/bin/bash
```
qgolsteyn: We should keep the shebang in the script
https://stackoverflow.com/questions/13872048/bash-script-what-does-bin-bash-mean

initial_setup.sh
```diff
@@ -1,32 +1,44 @@
...
-ln -s `pwd`/node_modules frog
-ln -s frog/babel.config.js .
-node linkFiles.js
+if python -c 'import sys; print(sys.version_info[0] == 2)' > /dev/null == 'True';
+then
+    YARN_VERSION='1.15.2'
+    if which yarn && [[ `yarn --version` == $YARN_VERSION ]]; then
+        echo 'Using pre-installed global Yarn'; YARN=yarn
+    else
+        if [ -f "./node_modules/.bin/yarn" ] && [[ `./node_modules/.bin/yarn --version` == $YARN_VERSION ]]; then
```
qgolsteyn: What happens if the script is not run from the correct directory?

initial_setup.sh
```diff
@@ -1,32 +1,44 @@
...
+    fi
+    echo "Yarn: $YARN"
 
-echo
-echo -e "\xE2\x9C\xA8  Finished Initial Setup."
-echo -e "\xF0\x9F\xA4\x93  Run ${LBLUE}npm start server ${NC}to begin hacking!"
+    if ! "$YARN" install; then
+    	echo "Sorry, there seems to be something wrong, please correct the errors and try again"
+    	echo "There was some error installing the required packages, please rectify the errors and try again"
+	    exit 1
```
qgolsteyn: We should be a bit more descriptive with our error message. Maybe something along the lines of "Unable to run yarn install successfully. You might not be connected to the internet, or there might be a problem with your FROG folder"

initial_setup.sh
```diff
@@ -1,32 +1,44 @@
...
-    echo 'Using pre-installed global Yarn'; YARN=yarn
-else
-    if [ -f "$FROG/node_modules/.bin/yarn" ] && [[ `"$FROG/node_modules/.bin/yarn" --version` == $YARN_VERSION ]]; then
-        echo 'Using pre-installed local Yarn'; YARN="$FROG/node_modules/.bin/yarn"
-    else
-        echo 'Installing Yarn'; npm install yarn@$YARN_VERSION --no-package-lock && YARN="$FROG/node_modules/.bin/yarn"
-    fi
+if ! PYTHON_VERSION=$(python --version);
+then
+    echo "Cannot find Python, please install python 2 and try again"
```
qgolsteyn:
``` suggestion
    echo "Cannot find Python, please install Python 2 and try again"
```

initial_setup.sh
```diff
@@ -1,32 +1,44 @@
...
+    fi
+    rm -rf frog/node_modules
+    ln -s `pwd`/node_modules frog
+    ln -s frog/babel.config.js .
+    node linkFiles.js
 
-exit 0
+    echo
+    echo -e "\xE2\x9C\xA8  Finished Initial Setup."
+    echo -e "\xF0\x9F\xA4\x93  Run ${LBLUE}npm start server ${NC}to begin hacking!"
```
qgolsteyn: nit: We could move the emojis to constants to help with readability

initial_setup.sh
```diff
@@ -1,32 +1,44 @@
...
+    exit 0
+else
```
qgolsteyn: This else statement could be changed to an if statement checking if the script failed and exiting early.

qgolsteyn (changes requested):
Left a couple of comments.

Generally:
- Would be good to add comments to describe the various commands in this script
- Maybe we should clean up the script to remove nested if statements. Try to keep error checking contained and exit early.

initial_setup.sh
ajainuary: Sorry, I didn't mean to remove it. It was an inadvertent error.

initial_setup.sh
ajainuary: That would be displayed by `yarn` itself. This just informs the user that his action wasn't successful.

initial_setup.sh
ajainuary: > go to the root repository

It is mentioned in the instructions for `README.md`
We could have error handling for that but seems a bit far-fetched idea to me.

initial_setup.sh
ajainuary: We could introduce a flag to remove emoji.

initial_setup.sh
ajainuary: Made the respective changes

initial_setup.sh
qgolsteyn: Good point, maybe add a quick comment in the file itself to keep this assumption close to the source code

initial_setup.sh
qgolsteyn: Unclear if there is benefit in doing so, this file will normally only be run once

qgolsteyn (approved):
Looks good! We could add some comments indicating the various steps we go through in this script for readability. Something like:

```
## INSTALLING PYTHON ##

<code goes here>
```

Very nitpicky so feel free to ignore if you don't have time

PR #1693, opened 2019-05-25
Wiki refactoring proper

dinomujki:
- Refactoring of the old 'index.js' for the Wiki
- Separated out the content part and the WikiLink component
- Many fixes and state update optimizations
- Implemented properly showing two wiki pages side-by-side
- Fixed linking between pages and made a lot simpler (all updates are handled through the state, URL only used for 'back' functionality and updated to allow users to share)
- NOTE: The functionality to add different activities/operators was temporarily removed (some relevant functions from before can be found in the deleted file 'SAVED_FUNCTIONS.js' in the commit history


dinomujki:
Following bugs fixed:

- all pages view completely broken - seems like query parameter doesn't get sent, so it returns all LIs in the database
- going to a URL of a deleted page yields a completely grey screen
- when deleting a page, someone viewing another instance doesn't get thrown out of that instance

houshuang:
Feedback from 28.05.2019:
- I can create multiple pages with the same name from createModal
- when a page is open in one browser, and gets deleted from another browser, the page completely reloads
- when I create a p1 page with user teacher, and log in as user B, and click on the link to that p1 page in the sidebar, nothing happens (also no error)
- all pages look good!
- revisions still broken (clicking on links)
- creating activity type broken
- logic around deletion, restoring etc all seem good. (except for triggering complete reload for other user when deleting a page).

PR #1695, opened 2019-05-28
Fixes for Wiki

ajainuary:
1. Removes console statements
2. Removes middle plane from social plane selector


ajainuary:
Fixed minor issue with value for select

PR #1696, opened 2019-05-29
Anonymous login for wiki and single activity

ajainuary:
Creates an anonymous user when a non-logged in user visits `wiki` or `single_activity`

### DIFF
--- a/frog/imports/client/Wiki/WikiTopNavbar.js
+++ b/frog/imports/client/Wiki/WikiTopNavbar.js
@@ -6,11 +6,18 @@ import History from '@material-ui/icons/History';
 import Delete from '@material-ui/icons/Delete';
 import ImportContacts from '@material-ui/icons/ImportContacts';
 
-class WikiTopNavbar extends React.Component<> {
-  constructor(props) {
-    super(props);
-
-    this.state = {};
+class WikiTopNavbar extends React.Component<{ username: string }> {
+  constructor() {
+    super();
+    this.state = {
+      username: '...'
+    };
+    Meteor.subscribe('userData', () => {
+      const user = Meteor.users.findOne();
+      this.setState({
+        username: user.isAnonymous ? 'Anonymous Visitor' : user.username
+      });
+    });
   }
 
   render() {
@@ -20,7 +27,6 @@ class WikiTopNavbar extends React.Component<> {
       deleteLI,
       moreThanOnePage
     } = this.props;
-
     const topNavBarStyle = {
       display: 'flex',
       flex: '0 0 50px',
@@ -119,7 +125,7 @@ class WikiTopNavbar extends React.Component<> {
         ) : (
           <div />
         )}
-        <div style={topNavBarItemStyleName}>{Meteor.user().username}</div>
+        <div style={topNavBarItemStyleName}>{this.state.username}</div>
       </div>
     );
   }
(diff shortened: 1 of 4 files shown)


houshuang:
This is a good start. 

It doesn't seem like you're setting any flag on the user that they're an automatically generated? Can you add that, and also check in the wiki whether that flag is set, and don't display the username in the menubar if it's an anonymous user.

Also, ideally we would not let anonymous users access other paths... Currently if you go directly to for example /teacher, it will not let you in unless you're logged in or use ?login. However, if you first go to the wiki, and then go to /teacher, it will let you in as that anonymous user...

This is more complex and should be a separate PR. First finish up the first PR with the anonymous ID and removing it from the wiki menu.

Then a separate PR where you try to see if someone is logging in using a token from localSession, and whether that user is an anonymous user, then you should delete the token and treat it like user is not logged in.

You'll have to think a bit about how to do this - maybe the logic should be on the server when you call the login function, maybe you should try to store information in localSession about whether it's an anonymous user or not - see if you can figure it out, otherwise I can help.

frog/server/user_accounts.js
```diff
@@ -1,22 +1,28 @@
...
+      return Accounts._loginUser(self, alreadyUser._id);
+    }
   }
-
-  const { userId } = Accounts.updateOrCreateUserFromExternalService('frog', {
-    id: user
-  });
-  Meteor.users.update(userId, { $set: { username: user } });
+  const userServiceData = {
+    isAnonymous: user ? false : true,
```
ajainuary: Here a flag is set stating that the user is Anonymous

ajainuary:
Requirements:
- [x] Hide username for anon user in wiki
- [x] Don't allow other paths without login

frog/server/user_accounts.js
```diff
@@ -1,22 +1,30 @@
...
-    id: user
+  const userServiceData = { id: user || uuid() };
+  const { userId } = Accounts.updateOrCreateUserFromExternalService(
+    'frog',
+    userServiceData
+  );
+  Meteor.users.update(userId, {
+    $set: {
+      username: userServiceData.id,
+      isAnonymous: user ? false : true
```
ajainuary: The flag for the anonymous user is set here

frog/imports/client/Wiki/WikiTopNavbar.js
```diff
@@ -7,20 +7,14 @@ import Delete from '@material-ui/icons/Delete';
 import ImportContacts from '@material-ui/icons/ImportContacts';
 
 class WikiTopNavbar extends React.Component<> {
-  constructor(props) {
-    super(props);
-
-    this.state = {};
-  }
-
```
houshuang: Why did you remove this? it can cause a crash if a component tries to access state without state being initialized.

frog/server/main.js
```diff
@@ -95,7 +95,7 @@ Meteor.publish('userData', function() {
     return this.ready();
   }
   return Meteor.users.find(this.userId, {
-    fields: { username: 1, joinedSessions: 1, role: 1 }
+    fields: { username: 1, isAnonymous: 1, joinedSessions: 1, role: 1 }
```
houshuang, line 98: good catch

frog/server/user_accounts.js
houshuang: why not replace with !!user

houshuang (changes requested):

frog/imports/client/Wiki/WikiTopNavbar.js
ajainuary: Component doesn't require a state anymore, I turned it into a functional component now

houshuang:
@ajainuary When I try the latest version, I see the random student ID in the wiki?

ajainuary:
I've fixed the typo, please clear the localStorage for testing (close all incognito windows and open again).

houshuang (approved):

PR #1697, opened 2019-05-31
Tooltip now moves with mouse

ajainuary:


houshuang:
Looks good. Can you make the styling slightly different, right now it's hard to see whether it's a tooltip or a part of the menu.

Screenshot 2019-05-31 at 07 20 04

houshuang:
Also is it possible to have a little bit longer delay before the tooltip appears? only if it's easy.

houshuang:
Exactly how I wanted it, thanks!.

Issue #1705, opened 2019-06-04
[Bug] Search in Activity Preview 


RachitMalik12:
**Description**
Searching for learning items on Activity preview causes screen to go blank. 

**Steps to Reproduce** 

1. Open ActivityPreview side panel 
2. Type text in the search bar 

**Expected behaviour** 
The list of activities should filter based on the search query. 

**Observed behaviour**
The page goes blank and the search box is unresponsive.

PR #1706, opened 2019-06-04
Fixes the issue with search

ajainuary:
The bug was caused due to change in Tooltip. ( #1705)


houshuang:
@RachitMalik12 Can you review please?

RachitMalik12 (approved):
Looks good! The bug was that the ToolTips were white before and then they were changed to black but the white tool tip dependency was removed. It was not changed from black to white in one place which caused the crash on searching.

PR #1713, opened 2019-06-08
Added logout feature to Wiki 

RachitMalik12:
This PR introduces the ability for a user to logout from the wiki. 

**How to test?**

- Open the wiki and login as any user.
- Click logout and you should be redirected to the FROG user not logged in page.

**Asana Issue** 

https://app.asana.com/0/1121331595459324/1125930882647545/f


ajainuary (approved):
Looks good to me

frog/imports/client/GraphEditor/Graph.js
```diff
@@ -23,7 +23,6 @@ const scrollMouse = e => {
 };
 
 const mousemove = e => {
-  // We do -100 here because there is 100px above the graph editor
```
qgolsteyn, line 26: I am not sure if we should remove this comment, it provides an explanation for the magic number below.

qgolsteyn (commented):
Just a tiny nit comment, but otherwise looks good

houshuang:
This feature doesn't make any sense until we have the login screen where you choose or type in a name, and once we have that feature, logging out should bring you back to that screen, not to the root URL. Once that feature has been implemented, I will review this PR again and see if it's doing the right thing.

qgolsteyn:
#1716 will break this PR, but can be updated by adding a secondary nav item

houshuang:
@RachitMalik12 	You can see if updating this PR makes sense given all your new work, or if it should be closed.

PR #1715, opened 2019-06-11
Removed confirmation modal when restoring pages

qgolsteyn:
This PR removes the confirmation modal that appears when trying to restore a page. Instead, the page restored immediately.

**Testing done**

Manual testing

**Issue completed**
https://app.asana.com/0/1121331595459324/1126262288715643


houshuang (changes requested):
If I delete an individual page with some instances, and when following a link say "create new" instead of "restore", all the old instances are still there. (This doesn't happen with a whole class page).

houshuang:
Also we should switch to the restored page immediately.

houshuang:
@qgolsteyn 	Ping, we should try to get this merged in.

qgolsteyn:
We now go to the page after it is restored.

qgolsteyn:
Regarding the other issue, I opened a new task in Asana:
https://app.asana.com/0/1121331595459324/1127117535559358

PR #1717, opened 2019-06-12
Fix long links in Reactive Rich Text

qgolsteyn:
This PR fixes the issue with Reactive Rich Text in which mention items with long text overlaped with other items. 

The fix now allows each mention element to take the height it needs to display its text content.

Font was reduced to increase the number of elements that can be displayed and max height of the mention panel was limited to 300px to avoid it going outside the bounds of the screen.

**Testing done**
Manual testing

**Issues completed**
https://app.asana.com/0/1121331595459324/1126732819188826


houshuang (approved):
Great fix!!

PR #1718, opened 2019-06-12
Make long wiki page titles more 'separate' and visible on the sidebar

RachitMalik12:
This PR adds a separator between each title on the sidebar, added some margin and added an additional highlight when the current page is selected. Hopefully this makes long titles more distinguishable. See screenshot

**Testing done**
Manual testing 

**Asana issue**
https://app.asana.com/0/1121313931563676/1126579168821756/

**Future work**
- I would potentially want to work on this more and make it look nicer by using Material UI ListItem's 

Screen Shot 2019-06-12 at 5 02 25 PM

### DIFF
--- a/frog/imports/client/Wiki/ModalFind.js
+++ b/frog/imports/client/Wiki/ModalFind.js
@@ -45,7 +45,6 @@ export const PagesLinks = ({
             fontSize: '14px',
             marginTop: '12px',
             backgroundColor: i === index ? 'cornflowerblue' : (currentPageBool ? '#e6e6e6': undefined), 
-            borderBottom: '2px solid #bfc7d0'
       }
       return (
         <li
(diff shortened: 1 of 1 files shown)


houshuang:
@RachitMalik12 	I removed the border, and merged this. Pushing it to prod now. I'm still not crazy about it, but maybe it's OK to leave it for now until we redesign it completely (with more hierarchy etc). One small change that would be nice would be to extend the click target to 100% of the horizontal length no matter how long the page name actually is.

RachitMalik12:
@houshuang Same would like to work on redesigning this. I pushed some changes to make the highlight go to 100% width of the parent. I don't think it reflects here since this has been merged.  I made a new PR If you would like to merge that. Thanks!

PR #1719, opened 2019-06-12
Random username generation for the anonymous user

ajainuary:
This replaces the 'Anonymous Visitor' greeting in the Wiki's Top Navbar to the user's randomly computer generated human friendly english username composed of 2 words.

**How to test?**
`git pull`
`git checkout random-username`
Since it introduces a new package, please do
`npm start yarn`
`npm start server`
View the wiki as an Anonymous user (i.e. without login)

### DIFF
--- a/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
+++ b/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
@@ -1,4 +1,4 @@
-//@flow
+// @flow
 
 import * as React from 'react';
 
@@ -24,7 +24,7 @@ export default (props: PrimaryButtonPropsT) => {
         alignItems: 'center',
         justifyContent: 'center',
         fontSize: '14px',
-	fontStyle: i ? 'italic' : 'normal',
+        fontStyle: i ? 'italic' : 'normal',
         cursor: callback && 'pointer',
         padding: '20px 0'
       }}
--- a/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
+++ b/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
@@ -6,11 +6,12 @@ type PrimaryButtonPropsT = {
   title: string,
   active?: boolean,
   icon?: React.ComponentType<*>,
-  callback?: () => void
+  callback?: () => void,
+  i?: boolean
 };
 
 export default (props: PrimaryButtonPropsT) => {
-  const { active, title, icon, callback } = props;
+  const { active, title, icon, callback, i } = props;
 
   const Icon = icon;
 
@@ -23,6 +24,7 @@ export default (props: PrimaryButtonPropsT) => {
         alignItems: 'center',
         justifyContent: 'center',
         fontSize: '14px',
+	fontStyle: i ? 'italic' : 'normal',
         cursor: callback && 'pointer',
         padding: '20px 0'
       }}
--- a/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
+++ b/frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
@@ -7,11 +7,11 @@ type PrimaryButtonPropsT = {
   active?: boolean,
   icon?: React.ComponentType<*>,
   callback?: () => void,
-  i?: boolean
+  italics?: boolean
 };
 
 export default (props: PrimaryButtonPropsT) => {
-  const { active, title, icon, callback, i } = props;
+  const { active, title, icon, callback, italics } = props;
 
   const Icon = icon;
 
@@ -24,7 +24,7 @@ export default (props: PrimaryButtonPropsT) => {
         alignItems: 'center',
         justifyContent: 'center',
         fontSize: '14px',
-        fontStyle: i ? 'italic' : 'normal',
+        fontStyle: italics ? 'italic' : 'normal',
         cursor: callback && 'pointer',
         padding: '20px 0'
       }}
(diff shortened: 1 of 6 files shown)


houshuang:
Note that you don't need to do git clean, etc. Simply do npm start yarn, and it will update all the packages.

frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
```diff
@@ -1,16 +1,17 @@
...
 
 import * as React from 'react';
 
 type PrimaryButtonPropsT = {
   title: string,
   active?: boolean,
   icon?: React.ComponentType<*>,
-  callback?: () => void
+  callback?: () => void,
+  i?: boolean
```
houshuang: this variable is not very descriptive, maybe rename as italic

frog/imports/client/Wiki/components/TopNavbar/index.js
```diff
@@ -7,8 +7,9 @@ import OverflowPanel from './OverflowPanel';
 
 type TopNavBarPropsT = {
   /** The current Meteor username */
-  user: string,
-
+  username: string,
+  /* Status of the user, username will be displayed in italics if true */
```
houshuang: we prefer // comments unless they are multiline

frog/imports/client/Wiki/components/TopNavbar/index.js
```diff
@@ -7,8 +7,9 @@ import OverflowPanel from './OverflowPanel';
 
 type TopNavBarPropsT = {
   /** The current Meteor username */
```
houshuang: this kind of comment is not very useful. Try to limit comments to things that are not obvious.

frog/imports/client/Wiki/index.js
```diff
@@ -110,7 +111,8 @@ class WikiComp extends Component<WikiCompPropsT, WikiCompStateT> {
       deletedPageModalOpen: false,
       currentDeletedPageId: null,
       currentDeletedPageTitle: null,
-      rightSideCurrentPageObj: null
+      rightSideCurrentPageObj: null,
+      user: Meteor.user()
```
houshuang: why would you put this in state? Why can't you just query Meteor.user() when you need it? It might make sense if you were actually listening to the Meteor.user and updating state when it changes, but you don't (and we don't expect it to change, except if we allowed renaming in the future, as discussed on Slack)

frog/imports/client/Wiki/index.js
```diff
@@ -223,7 +225,7 @@ class WikiComp extends Component<WikiCompPropsT, WikiCompStateT> {
     if (!currentPageObj) {
       if (!fullPageObj.noNewInstances) {
         this.initialLoad = true;
-        const instanceName = Meteor.user().username;
+        const instanceName = this.state.user.username;
```
houshuang: why this change?

frog/server/user_accounts.js
```diff
@@ -13,7 +17,7 @@ const doLogin = (user, self) => {
       return Accounts._loginUser(self, alreadyUser._id);
     }
   }
-  const userServiceData = { id: user || uuid() };
+  const userServiceData = { id: user || uniqueNamesGenerator(usernameConfig) };
```
houshuang: can we make the username capital first letters, like Happy Panda instead of happy panda?

houshuang (changes requested):

frog/imports/client/Wiki/index.js
ajainuary: On the server, this will fetch the record from the database. To improve the latency since it is being called multiple times, we could save the returned record to a variable instead of re-calling `Meteor.user()`.

frog/imports/client/Wiki/components/TopNavbar/index.js
ajainuary: This comment is not a part of the PR.

frog/server/user_accounts.js
ajainuary: Made the changes

frog/imports/client/Wiki/components/TopNavbar/index.js
ajainuary: Removed the comment

frog/imports/client/Wiki/index.js
ajainuary: In order to improve the performance, we re-use the value instead of calling `Meteor.user()` again.

frog/imports/client/Wiki/components/TopNavbar/index.js
ajainuary: Made the changes

frog/imports/client/Wiki/components/TopNavbar/PrimaryButton.js
ajainuary: Made the changes

frog/imports/client/Wiki/index.js
houshuang: That's not how Meteor works - Meteor uses something called Minimongo, which subscribes to certain collections and always keeps them in sync locally. The only time a server call is made is when you make a new subscription (with withTracker) or when you call a Meteor method. Please revert this.

houshuang (changes requested):

houshuang:
Note that you don't need to do git clean, etc. Simply do npm start yarn, and it will update all the packages.

frog/imports/client/Wiki/index.js
ajainuary: Reverted the changes

ajainuary:
Reverted to using `meteor.user()`

houshuang (approved):

PR #1720, opened 2019-06-13
Make highlight extend to 100% of width of parent container in sidebar of Wiki

RachitMalik12:
Reference [https://github.com/chili-epfl/FROG/pull/1718](url).
This PR makes the changes discussed in the link above.


houshuang:
@RachitMalik12 	I made some quick changes to actually get the pointer and the onClick to also work in the whole horizontal width.

PR #1721, opened 2019-06-13
Refactor/wiki modal

qgolsteyn:
This PR adds logic to handle modal state, allowing us to move it away from index.js. It exposes two methods, `showModal` and `hideModal` allowing us to display a particular React component as a modal, and then hide it when appropriate. Further work to move the Wiki modals to this new logic is needed.

**Testing done**
Manual testing

### DIFF
--- a/frog/imports/client/Wiki/components/Modal/withModalController.js
+++ b/frog/imports/client/Wiki/components/Modal/withModalController.js
@@ -1,0 +1,52 @@
+//@flow
+
+import * as React from 'react';
+
+import {
+  type ShowModalFunctionT,
+  type HideModalFunctionT,
+  type ModalParentPropsT
+} from './types';
+
+/**
+ * HOC that adds ability to create modals via provided prop function
+ */
+export const withModalController = <T: {}>(
+  Component: React.AbstractComponent<T>
+): React.AbstractComponent<
+  $Diff<
+    T,
+    {
+      showModal: ShowModalFunctionT | void,
+      hideModal: HideModalFunctionT | void
+    }
+  >
+> => {
+  return (
+    props: $Diff<
+      T,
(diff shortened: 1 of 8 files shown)


frog/imports/client/Wiki/components/Modal/withModalController.js
```diff
@@ -0,0 +1,52 @@
...
+
+import * as React from 'react';
+
+import {
+  type ShowModalFunctionT,
+  type HideModalFunctionT,
+  type ModalParentPropsT
+} from './types';
+
+/**
```
houshuang, line 11: I prefer single line // comments

houshuang:
In general looks very good to me, nice use of advanced React features. 

I wonder if in the future this could be generalized and moved to frog-utils.

I still feel like the look and feel of the modals is a bit too minimalistic, for example the restore modal.

Anyway if you can fix the merge conflicts I'm happy to merge this one, and we can make more improvements in the future.

houshuang (changes requested):

frog/imports/client/Wiki/components/Modal/withModalController.js
qgolsteyn: Javadoc style comments can be picked up by IDEs and make it easier to understand the code by hovering over variables and function calls. Happy to revert to `//` however if you want.

qgolsteyn:
Fixed the merge conflicts

frog/imports/client/Wiki/components/Modal/withModalController.js
houshuang: I think we should use Flow for that instead?

houshuang (approved):

PR #1723, opened 2019-06-14
Activities in wiki

ajainuary:
**Description**
This PR refactors the code for creating new pages in the Wiki, for creating a LI in the Wiki and most importantly the `duplicateLI` function. Luckily, any old pages/wikis made previously would still work flawlessly.
**How to test**

1. Create an activity in the 'Individual Plane'
2. Use the activity from 2 different users (they will not share the data)

**Resolved Issues**
Asana task for reference: https://app.asana.com/0/1121331595459324/1125930882647548


frog/imports/api/generateReactiveFn.js
```diff
@@ -187,7 +187,14 @@ export class Doc {
       'li',
       id
     );
-
+    if (LIData.liType === 'li-activity') {
+      // In this case the activity also needs to be duplicated
+      const acId = uuid();
+      const activityPointer = connection.get('rz', acId);
+      // New instance starts with empty data
+      activityPointer.create({});
```
houshuang, line 195: this should be activityTypesObj[acId].dataStructure || {}

houshuang: so the need for a specific duplication mechanism is exactly why we have the option of having a duplicate function in the liType... so can you move this logic to li-activity, adding a duplicate function to the main export?

houshuang: also, should a new instance start with empty data even if there is already a lot of data? that's not my intuition of cloning - if I have a rich text with lots of text, and clone it, I get two lis with lots of text... cloning a ck-board should be the same?

frog/imports/client/Wiki/index.js
```diff
@@ -466,45 +448,33 @@ class WikiComp extends Component<WikiCompPropsT, WikiCompStateT> {
...
 
-  createLI = (newTitle, plane, config, operatorConfig) => {
+  // Creates a new page entry in ShareDB and navigates to it.
+  createPage = (title, socialPlane, activityConfig, operatorConfig) => {
     const error =
-      checkNewPageTitle(wikiStore.parsedPages, newTitle) ||
-      (config.invalid && 'Activity config is not valid') ||
-      (operatorConfig.invalid && 'Operator config is not valid');
+      checkNewPageTitle(wikiStore.parsedPages, title) ||
+      (activityConfig &&
```
houshuang: why not activityConfig?.invalid

houshuang (changes requested):
Great start. See my comments on code - also when I try to create a new Individual Page with an activity type, the page gets created but the create modal stays open...

frog/imports/api/generateReactiveFn.js
ajainuary: Where should the instance get the new data from? (This is for creating a new instance when a new individual comes)

ajainuary: If we export it into the `li-activity` we won't be able to assign the `id` of the activity to the `rz` field of the LI.

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

houshuang (approved):

PR #1725, opened 2019-06-17
Cloning data with activities

ajainuary:
**Description**
This PR modifies the cloning logic for activities in the wiki to allow cloning of the data with the activities

**Resolved Issues**
This is a follow up from #1723


houshuang:
@ajainuary Can you fix merge conflict please?

frog/imports/api/generateReactiveFn.js
```diff
@@ -189,10 +189,11 @@ export class Doc {
     );
     if (LIData.liType === 'li-activity') {
       // In this case the activity also needs to be duplicated
+      const ac = connection.get('rz', LIData.payload.rz);
+      ac.fetch();
```
houshuang: This is asynchronous, so you will probably have timing issues here in production (not on localhost). You should use a callback or Promise). See other places where we use connection.get - we always wait for onload etc.

frog/imports/client/Wiki/WikiTopNavbar.js
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
```
houshuang: What happened to this file - is this because of a merge conflict?

houshuang (changes requested):

frog/imports/client/Wiki/WikiTopNavbar.js
ajainuary: Yes, somehow git merge re-introduced this file

ajainuary: (Removed now)

frog/imports/api/generateReactiveFn.js
ajainuary: Now using await function

houshuang (approved):

PR #1726, opened 2019-06-18
Creating and Displaying Single activities

ajainuary:
**Description**
This PR introduces the functionality in the Single Activity, users can create and use Single Activities and the interface for creation is accessible at `/single_activity` and the activity will run at `/ac/:id`.

**Resolved Issues**
This completes the asana task:
https://app.asana.com/0/1115664036175110/1124152769431682

**How to test?**
Go to `/single_activity` and choose an activity type and configure it suitably, then navigate to the presented URL to view the activity.


ajainuary:
The curent implementation by using LI from the Wiki is not suitable, instead Graph sessions must be used as suggested by @houshuang.

PR #1727, opened 2019-06-18
Cannot rename page to an invalid page in wiki 

RachitMalik12:
The following PR fixes the bug of pages getting lost if you rename it to an existing page title. Also it fixes the bug of letting the user rename the page title to empty. 

**Steps to reproduce**

- Create a page on the wiki with title X 

- Create another page on the wiki with title Y 

- Rename page Y to X. There will be 2 pages with title X on the sidebar. Furthermore, Y can be renamed to anything including an empty title. 

**Testing done**
Manual testing 

**Future thoughts**
There is a lot of similarities with the error handling logic when Creating a page and renaming it.
So I would possibly like to make another PR that abstracts this out in some form.

### DIFF
--- a/frog/imports/client/Wiki/WikiContentComp.js
+++ b/frog/imports/client/Wiki/WikiContentComp.js
@@ -77,15 +77,15 @@ class WikiContentComp extends React.Component<> {
   handleErrorClearing(currentTitle) {
     if (
       currentTitle === '' ||
-      (currentTitle.length > 0 && this.state.error === 'Title cannot be empty')
-    )
+      (currentTitle.length > 0 && this.state.error === 'Title cannot be empty')  )
       this.clearErrors();
   }
 
   saveNewPageTitle = () => {
     const { pageTitleString } = this.state;
     const error = this.handleErrors(pageTitleString);
-    if (error === null) {
+  
+    if (error === null || pageTitleString === this.props.currentPageObj?.title) {
       this.props.changeTitle(this.props.currentPageObj.id, pageTitleString);
       this.setState({
         pageTitleString,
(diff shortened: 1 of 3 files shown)


frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -130,20 +151,30 @@ class WikiContentComp extends React.Component<> {
...
         </Button>
       );
-    })();
+    };
 
     const titleDiv = this.state.editingTitle ? (
       <div style={titleDivStyle}>
-        <div>
-          <input
+        <FormControl margin="normal">
```
houshuang, line 164: Seems like you are also making UI changes, while trying to fix a bug? Is this integral to fixing the bug? Otherwise would have been better to have two separate PRs, easier to test/review.

houshuang:
Remember to run prettier before making PR, for example using npm start eslint.fix. I've fixed it for you.

houshuang (changes requested):
Currently you cannot rename a page to the same title - ie if I open the rename module, and decide not to rename, I have to reload the page... We should accept renaming to the same page as the current page. Otherwise looks good.

frog/imports/client/Wiki/WikiContentComp.js
RachitMalik12: I needed to add helper text to the input to display the errors. I felt it was better to use the Material UI's TextField and FormHelperText than the HTML input to make it more uniform with all of the other UI, so I decided to include that in. I could make a separate PR for this but then for this PR (if I wanted to use the standard HTML input) I would have to use the standard HTML label and include some CSS to display the error properly. Regardless things would have to change on the UI side of things. Let me know what you think.

RachitMalik12:
> Remember to run prettier before making PR, for example using npm start eslint.fix. I've fixed it for you.

oh sorry about that I think I ran it at first and then changed something and forgot to run it in the end. Will take care next time. Thanks!

frog/imports/client/Wiki/WikiContentComp.js
houshuang: That's totally fine if it's integral to fixing the bug. Thanks

PR #1728, opened 2019-06-20
Fixes exporting activities from the Wiki

ajainuary:
**Description**
Due to changes in the structure of documents in Wiki, the exporting activities to wiki from the TeacherView was broken. This PR fixes that.

**Resolved Issues**
Asana task for reference: https://app.asana.com/0/1115664036175110/1127349467155372

**How to test?**
Run a graph and click on "Export activities to wiki", the activities should be imported into the chosen wiki page along with their data.

**Pending Changes**
- [ ] **Individual Plane**: Export individual activities as individual pages. (Currently, individual activities  don't export properly)


ajainuary:
Error handling implemented, does not crash the server in case of Error on the RPC call.

PR #1730, opened 2019-06-20
Wiki page title can now contain special characters except % and /

RachitMalik12:
This PR allows wiki titles to contain special characters except % and / by encoding the title. Earlier FROG would crash if you make a page with title containing a ? for example. Also, this PR adds error handling if the title contains %.

**Testing instructions**
Create a wiki page title with a special character and it should not crash.

**Testing done**
Manual testing - entered a lot of random combinations of text with random special characters.


ajainuary (approved):
**Testing Done**
Tested with multiple inputs and special characters, handles all test cases.

houshuang (approved):

PR #1732, opened 2019-06-21
Wiki Embed Option

ajainuary:
**Description**
This PR introduces and option to embed the wiki from anywhere by specifying the `WikiId`, `page` and (optionally) the `instance`.

**Resolved Issues**
This is the first part of bringing the Wiki to the FROG Graph via an activity.

**How to test?**
Passing `embedPage` Object to the Wiki should cause only the content to be displayed.


houshuang:
Please fix merge conflict. 

How do I test this? Do you have any code anywhere that actually tries renders this component?

frog/imports/client/Wiki/index.js
```diff
@@ -52,13 +52,19 @@ export const LearningItem = dataFn.LearningItem;
 
 type WikiCompPropsT = {
   location: *,
-  match: {
+  match?: {
```
qgolsteyn: If we are making this field optional, what happens if functions that use this prop receive `undefined`?

frog/imports/client/Wiki/index.js
```diff
@@ -74,7 +80,8 @@ type WikiCompStateT = {
 };
 
 class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
-  wikiId: string = this.props.match.params.wikiId;
+  wikiId: string =
```
qgolsteyn: Type annotation should not be necessary here. Flow should type infer.

frog/imports/client/Wiki/index.js
```diff
@@ -52,13 +52,19 @@ export const LearningItem = dataFn.LearningItem;
...
     params: {
       wikiId: string,
-      pageTitle: ?string
+      pageTitle: ?string,
+      instance?: string
     }
   },
-  history: Object
+  history: Object,
+  embedPage?: {
```
qgolsteyn: What is the difference between embedPage and match? Could we merge the two if they contain the same props?

frog/imports/client/Wiki/index.js
```diff
@@ -52,13 +52,19 @@ export const LearningItem = dataFn.LearningItem;
...
+    wikiId: string,
+    page: string,
```
qgolsteyn: Could rename to pageTitle for more clarity.

frog/imports/client/Wiki/index.js
```diff
@@ -74,7 +80,8 @@ type WikiCompStateT = {
 };
 
 class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
-  wikiId: string = this.props.match.params.wikiId;
+  wikiId: string =
+    this.props.match?.params.wikiId || this.props.embedPage.wikiId;
```
qgolsteyn: embedPage is type Maybe, what happens if it is undefined?

frog/imports/client/Wiki/index.js
qgolsteyn: We are setting a prop value as a class global. This will break React rerendering on prop change. If the prop changes, this value will not get changed, causing a potential bug.

frog/imports/client/Wiki/index.js
```diff
@@ -102,7 +109,8 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
       dashboardSearch: null,
       pageId: null,
       currentPageObj: null,
-      initialPageTitle: this.props.match.params.pageTitle || null,
+      initialPageTitle:
+        this.props.match.params.pageTitle || this.props.embedPage?.page || null,
```
qgolsteyn: What happens if initialPageTitle becomes null?

frog/imports/client/Wiki/index.js
```diff
@@ -255,7 +266,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
           (instanceId && instanceId !== this.getInstanceId(fullPageObj)
             ? '/' + instanceName
             : '');
-        this.props.history.push(link);
+        if (!this.props.embedPage) this.props.history.push(link);
```
qgolsteyn: NIT: Add curly braces to single if statement to help with readability

frog/imports/client/Wiki/index.js
```diff
@@ -347,7 +358,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
 
   changeTitle = (pageId, newPageTitle) => {
     changeWikiPageTitle(this.wikiDoc, pageId, newPageTitle);
-    const instanceId = this.props.match.params.instance;
+    const instanceId = this.props.match?.params.instance;
```
qgolsteyn: What happens if instanceId is undefined?

frog/imports/client/Wiki/index.js
```diff
@@ -413,10 +424,10 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
             ? '/' + instanceName
             : '');
         if (cb) {
-          this.props.history.replace(link);
+          if (!this.props.embedPage) this.props.history.replace(link);
```
qgolsteyn: Add curly braces

frog/imports/client/Wiki/index.js
```diff
@@ -413,10 +424,10 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
           return cb();
         }
-        this.props.history.push(link);
+        if (!this.props.embedPage) this.props.history.push(link);
```
qgolsteyn: Add curly braces

frog/imports/client/Wiki/index.js
qgolsteyn: Make sure to update other instances of this variable

``` suggestion
  embeddedPage?: {
```

qgolsteyn (changes requested):
It seems we are doing too much work in this file, as we are heavily relying on conditional statements to drive component behaviour. I would consider refactoring this file to remove as much common logic as possible, and creating two components, one for the primary wiki view, and one for the embedded view.

Also, it would be good to reduce our reliance on the Maybe type when possible to avoid having to type guard and reduce the risk of `null` or `undefined` related bugs.

frog/imports/client/Wiki/index.js
ajainuary: The only case that this could happen is when it is not called by the `react-router` (which is what happens when a user goes to the URL). This would imply that the Wiki is being embedded.

frog/imports/client/Wiki/index.js
ajainuary: We would still need a prop to signify whether the Wiki is embedded or not. Also `match` is kind of a special prop passed by the router so I wouldn't recommend to override that.

frog/imports/client/Wiki/index.js
ajainuary: It is the pageId, I'll change the variable name and write comments

frog/imports/client/Wiki/index.js
ajainuary: Which would mean it is not an embedded page but called by the `react-router`. (Which implies the existence of `props.match`)

frog/imports/client/Wiki/index.js
qgolsteyn: What would be the expected behaviour in this case? Would enabling a flag (`isWikiEmbeded` or similar) work better?

frog/imports/client/Wiki/index.js
ajainuary: Fair point, I faced a bug due to this one!

frog/imports/client/Wiki/index.js
ajainuary: It would result in an error, it was an inadvertent error. This is a good reason to not use maybe types to control the logic.

frog/imports/client/Wiki/index.js
ajainuary: There are no immediate bugs due to this issue, though this could be addressed later in a refactor of the Wiki.

frog/imports/client/Wiki/index.js
qgolsteyn: Yes, but what happens if they are both undefined, which is a possible state

ajainuary: It is possible but should never happen (you would either access the wiki via `react-router` or embed it, no other possible way to use it).

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

ajainuary:
@qgolsteyn Thank you for your feedback, I have made some changes.
You could now test the embed view on the branch `wiki-pageview-testing` and go to `/wiki_embed/<wiki-id>/<page-id>/<instance-id>`

frog/imports/client/Wiki/index.js
```diff
@@ -213,7 +224,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
     const instanceId =
       this.getInstanceIdForName(
         fullPageObj,
-        this.props.match.params.instance
+        this.props.match?.params.instance || this.props.embeddedPage?.instance
```
qgolsteyn: this.props.embeddedPage.instance does not exist in embeddedPage

frog/imports/client/Wiki/index.js
```diff
@@ -255,7 +266,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
           (instanceId && instanceId !== this.getInstanceId(fullPageObj)
             ? '/' + instanceName
             : '');
-        this.props.history.push(link);
+        if (!this.props.embeddedPage) this.props.history.push(link);
```
qgolsteyn: Add curly braces

frog/imports/client/Wiki/index.js
```diff
@@ -130,27 +138,30 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+        this.setState({ findModalOpen: true })
+      );
+    }
   }
 
   componentDidUpdate(prevProps) {
-    const pageTitle = decodeURIComponent(this.props.match.params.pageTitle);
+    const pageTitle =
+      decodeURIComponent(this.props.match?.params.pageTitle) ||
+      this.props.embeddedPage.pageId;
```
qgolsteyn: We are setting this.props.embeddedPage.pageId to pageTitle

frog/imports/client/Wiki/index.js
```diff
@@ -130,27 +138,30 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
 
     if (
       (pageTitle !== this.state.currentPageObj?.title &&
-        decodeURIComponent(prevProps.match.params.pageTitle) !==
-          decodeURIComponent(this.props.match.params.pageTitle)) ||
-      prevProps.match.params.instance !== this.props.match.params.instance
+        decodeURIComponent(prevProps.match?.params.pageTitle) !==
+          decodeURIComponent(this.props.match?.params.pageTitle)) ||
+      prevProps.match?.params.instance !== this.props.match?.params.instance
     ) {
```
qgolsteyn, line 152: Clean up conditional for clarity

frog/imports/client/Wiki/index.js
```diff
@@ -347,7 +358,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
 
   changeTitle = (pageId, newPageTitle) => {
     changeWikiPageTitle(this.wikiDoc, pageId, newPageTitle);
-    const instanceId = this.props.match.params.instance;
+    const instanceId = this.props.match?.params.instance || this.props.embeddedPage.instance;
```
qgolsteyn: Should be this.props.embeddedPage.instanceId

qgolsteyn (changes requested):
1. Simplify props. For the router, wrap the Wiki in a function and pass the correct parameters so we can reduce our reliance on the Maybe type
2. If we are not splitting up code, MAKE SURE that all instance where behaviour should be different are checked and double checked. There are a lot of conditional behaviour and we do not want to miss everything
3. Bring back yarn.lock :slightly_smiling_face:

houshuang:
@ajainuary 	Merge conflicts

houshuang:
Functionality looks good. Let me know when you've fixed merge conflict and given the code the final pass (addressing comments) and I will review/merge.

qgolsteyn:
@ajainuary Please resolve comments that you completed, and message me when you are ready for another round of code review :slightly_smiling_face:

frog/imports/client/Wiki/index.js
ajainuary: Reverted to using this.props.embeddedPage.instance to match the naming convention of the Router

frog/imports/client/Wiki/index.js
ajainuary: We can keep this for when we refactor the wiki

frog/imports/client/Wiki/index.js
ajainuary: It was an error, `embeddedPage` should have the string `pageTitle` not `pageId`

frog/imports/client/Wiki/index.js
ajainuary: Reverted to using `instance` instead of `instanceId`

frog/imports/client/Wiki/index.js
ajainuary: Made the respective changes

ajainuary:
@qgolsteyn It's ready for review, I appreciate your valuable feedback!

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -40,12 +40,11 @@ class WikiContentComp extends React.Component<> {
...
 
-  componentDidUpdate(prevProps) {
-    if (prevProps.currentPageObj.id !== this.props.currentPageObj.id) {
-      // eslint-disable-next-line react/no-did-update-set-state
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.currentPageObj.id !== this.props.currentPageObj.id) {
       this.setState({
         docMode: 'view',
-        pageTitleString: this.props.currentPageObj.title,
+        pageTitleString: nextProps.currentPageObj.title,
```
qgolsteyn, line 47: Unless necessary, we could simply rely on the prop being passed instead of saving to the state

frog/imports/client/Wiki/index.js
```diff
@@ -45,13 +45,19 @@ import {
 
 type WikiCompPropsT = {
   location: *,
-  match: {
+  match?: {
     params: {
       wikiId: string,
-      pageTitle: ?string
+      pageTitle: ?string,
```
qgolsteyn: Why is `pageTitle` Maybe here, but not on `embeddedPage`

frog/imports/client/Wiki/index.js
qgolsteyn: We can make use of type inference

frog/imports/client/Wiki/index.js
```diff
@@ -681,6 +699,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
                 goToPage={this.goToPage}
                 dashboardSearch={this.state.dashboardSearch}
                 side={this.state.mode === 'splitview' ? 'left' : null}
+                embed={!!this.props.embeddedPage}
```
qgolsteyn: Use `!== undefined` instead

frog/imports/client/Wiki/index.js
```diff
@@ -694,6 +713,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
                   goToPage={this.goToPage}
                   dashboardSearch={this.state.dashboardSearch}
                   side="right"
+                  embed={!!this.props.embeddedPage}
```
qgolsteyn: Use `!== undefined` instead

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -159,7 +158,8 @@ class WikiContentComp extends React.Component<> {
       );
     };
 
-    const titleDiv = this.state.editingTitle ? (
+    const titleDiv = 
+    this.state.editingTitle && !this.props.embed ? (
```
qgolsteyn, line 162: `embed` is not defined as part of the props, there are no guarantees that it will be there

qgolsteyn, line 162: NIT: Instead of embed, we could use a more descriptive name to indicate what this prop does (eg. `disableTitleEdit`). This will reduce coupling as WikiContentComp will not need to know whether it is embedded or not

frog/imports/client/Wiki/index.js
```diff
@@ -69,7 +75,8 @@ type WikiCompStateT = {
 };
 
 class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
-  wikiId: string = this.props.match.params.wikiId;
+  wikiId: string =
+    this.props.match?.params.wikiId || this.props.embeddedPage?.wikiId;
```
qgolsteyn: Merging `match` and `embeddedPage` would negate the need for this conditional

qgolsteyn (changes requested):
This PR adds significant amount of coupling between components. Now `index.js` and `WikiContentComp` need to manage what an embedded view is. I am afraid we are exposing ourselves to a number of defect with this PR by creating a second operating mode and handling it through conditional logic throughout `index.js`. If we forget to add or remove a conditional, this will lead to a bug. If changes are made to the sidebar or header, we will need to take into consideration when the Wiki is embedded.

frog/imports/client/Wiki/WikiWrapper.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import * as React from 'react';
+import { withRouter } from 'react-router';
+import Wiki from './index';
+
+type PropsT = {
+    location: *,
```
qgolsteyn, line 10: We should specify the type

frog/imports/client/Wiki/WikiWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+    match: {
+      params: {
+        wikiId: string,
+        pageTitle: ?string,
+        instance?: string
+      }
+    },
+    history: Object,
+    embed: boolean
```
qgolsteyn, line 15: We can safely remove this

frog/imports/client/Wiki/WikiWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+class WikiWrapper extends React.Component<PropsT> {
```
qgolsteyn: Change this to a function component

frog/imports/client/Wiki/index.js
```diff
@@ -44,14 +43,13 @@ import {
 } from './components/Modal';
 
 type WikiCompPropsT = {
-  location: *,
-  match: {
-    params: {
-      wikiId: string,
-      pageTitle: ?string
-    }
+  setPage: Function,
```
qgolsteyn: We should use the `() => *` notation to specify parameters and return value

frog/imports/client/Wiki/WikiWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+type PropsT = {
+    location: *,
+    match: {
+      params: {
+        wikiId: string,
+        pageTitle: ?string,
+        instance?: string
+      }
+    },
+    history: Object,
```
qgolsteyn, line 14: We should specify the type

frog/imports/client/Wiki/index.js
```diff
@@ -44,14 +43,13 @@ import {
...
+  pageObj: {
+    wikiId: string,
+    pageTitle: string,
+    instance?: string
   },
-  history: Object
+  embed: boolean
```
qgolsteyn: Make this a Maybe type for implicit false or use defaultProps

frog/imports/client/Wiki/index.js
```diff
@@ -44,14 +43,13 @@ import {
...
 } & ModalParentPropsT;
```
qgolsteyn, line 55: Include query in props

frog/imports/client/Wiki/index.js
```diff
@@ -91,8 +87,8 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
       addNonActivePage: this.addNonActivePage
     };
 
-    const query = queryToObject(this.props.location.search.slice(1));
-
+    const query = queryToObject(this.props.query);
```
qgolsteyn: Move to WikiWrapper

frog/imports/client/Wiki/WikiWrapper.js
qgolsteyn: Add a quick Jsdoc comment explaining what it does

frog/imports/client/Wiki/index.js
```diff
@@ -131,27 +127,29 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
-
-    Mousetrap.bindGlobal('ctrl+n', () =>
-      this.setState({ createModalOpen: true })
-    );
-    Mousetrap.bindGlobal('ctrl+s', () => this.setState({ docMode: 'view' }));
-    Mousetrap.bindGlobal('ctrl+e', () => this.setState({ docMode: 'edit' }));
-    Mousetrap.bindGlobal('ctrl+f', () =>
-      this.setState({ findModalOpen: true })
-    );
+    if (!this.props.embeddedPage) {
```
qgolsteyn: `embeddedPage` is not a prop

frog/imports/client/Wiki/index.js
```diff
@@ -131,27 +127,29 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+    const pageTitle =
+      this.decodePageTitle(this.props.pageObj.pageTitle);
 
     if (
       (pageTitle !== this.state.currentPageObj?.title &&
-        this.decodePageTitle(prevProps.match.params.pageTitle) !==
-          this.decodePageTitle(this.props.match.params.pageTitle)) ||
-      prevProps.match.params.instance !== this.props.match.params.instance
+        this.decodePageTitle(prevProps.pageObj.pageTitle) !==
+          this.decodePageTitle(this.props.pageObj.pageTitle)) ||
```
qgolsteyn: Can be changed to `pageTitle`

frog/imports/client/Wiki/index.js
qgolsteyn: Add a JSDoc comment explaining what this does

frog/imports/client/Wiki/index.js
```diff
@@ -361,14 +357,11 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
   changeTitle = (pageId, newPageTitle) => {
     changeWikiPageTitle(this.wikiDoc, pageId, newPageTitle);
-    const instanceId = this.props.match.params.instance;
-    const link =
-      '/wiki/' +
-      this.wikiId +
-      '/' +
-      encodeURIComponent(newPageTitle) +
-      (instanceId ? '/' + instanceId : '');
-    this.props.history.replace(link);
```
qgolsteyn, line 371: We remove replace behaviour, make sure this is intentional

frog/imports/client/Wiki/index.js
```diff
@@ -694,14 +688,14 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
                   side="right"
+                  embed={this.props.embed}
                 />
               )}
             </div>
           </div>
         </div>
         {this.state.findModalOpen && (
           <FindModal
-            history={this.props.history}
```
qgolsteyn, line 704: Does FindModal still need history?

qgolsteyn (approved):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

qgolsteyn (changes requested):
I like the direction, make sure to add comments on props and `WikiWrapper`. For `setPage` maybe include the possibility to replace the URL rather than push it.

frog/imports/client/Wiki/index.js
ajainuary: No, it is not a part of its arguments

houshuang (approved):

PR #1733, opened 2019-06-21
Add scripts to run Cypress

qgolsteyn:
This PR adds Cypress as a dependency and adds scripts to run cypress in interactive and headless mode. Test script has been modified to include Cypress, which will make it possible to run it as part of the CI build.

**Testing done**
Manual testing


houshuang:
Our problem with this before was that we needed to ensure FROG was running properly before we ran the Cypress test scripts - otherwise they would fail or wait forever in a CI mode. Have you resolved this?

qgolsteyn:
The CI script should start the server, wait until `localhost:3000` becomes reachable, and start cypress in headless mode. Once Cypress completes its work, the server is shutdown. If the server hangs for any reason, the script will timeout after 5 min. Unfortunately, this value is not configurable.

houshuang (approved):

houshuang:
@qgolsteyn Would be great if you could go back and get the Preview Activity script to work again as well. Added to Asana.

PR #1738, opened 2019-06-24
Reloading a wiki page whose title contains special characters 

RachitMalik12:
This PR addresses the bug where on reloading a page/ opening a page in a new tab whose title contains a special character causes a blank screen. 

**Testing instructions**
- Create a new wiki page with a special character in the title, eg: "Hello?" 
- Reloading the page or opening it in a new window should lead to the expected result. 

**Asana Issue**
https://app.asana.com/0/1121331595459324/1128067724130213/f


houshuang (changes requested):
It handles the issue you're trying to handle well, but if you leave the page name empty, it goes to undefined rather than Home. I guess because urldecode(undefined) = 'undefined'. So you might need to move the logical check to check for a string before you urldecode.

houshuang (approved):

PR #1744, opened 2019-06-24
Fix new wiki link

houshuang:
This PR fixes the ability to create a link to a page that does not yet exist, and create it by clicking on it.

A key idea is that if you add three links to the same non-existant page, they will all link to the same object in the background, so that once you create the page, all links will be updated and point to the newly created page - if you then rename the page, all links will point to the renamed page etc.

Note that we have decided that creating a page by link always creates a li-richText on plane 3. 

https://app.asana.com/0/1121331595459324/1127850709273037/f

### DIFF
--- a/frog-utils/src/ReactiveRichText/WikiLink/WikiLinkModule.js
+++ b/frog-utils/src/ReactiveRichText/WikiLink/WikiLinkModule.js
@@ -222,10 +222,7 @@ class WikiLinkModule {
       );
     } else {
       if (!data.id) {
-        const { pageId, liId } = window.wiki.createNewGenericPage(
-          data.title,
-          false
-        );
+        const { pageId, liId } = window.wiki.addNonActivePage(data.title);
         data.id = pageId;
         data.liId = liId;
       }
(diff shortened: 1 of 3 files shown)


RachitMalik12 (commented):
One small edge case. I tried to create a new page with a special character and it did not work. Since we are allowing page names with special characters we should allow it with the links too. Other than that it seems good..

ajainuary:
What do you think about using the format `[[<Page name here>]]` to link to pages inside the wiki?
The same format is also used by Wikipedia.

ajainuary:
Is it possible to delay the creation of the new page until someone clicks on the link, which is when a new page modal could be shown to allow for the customization of the new page?
If possible, I would like to work on this.

houshuang:
@ajainuary 	Re your suggestion about delaying creation of pages, for now I've decided to make these kinds of pages "simple pages". We basically want to make it very quick to use the wiki for normal wiki stuff (normal pages with links etc), while still enabling complex stuff... 

We might revisit this in the future if we test it with users who request changes, but for now we'll leave it as it is. Thanks for your suggestion though!

PR #1746, opened 2019-06-26
Make sure allowed characters for link creation in RRT is in sync with wiki policy

RachitMalik12:
This PR makes sure allowed characters for link creation is consistent when you create a page in RRT and through the create modal. 

**Steps to test**
Type @ in the rich text editor and try to create a wiki page with a special character, for example: 
"Hello! are you there?". If the page doesn't exist a green link should pop up and clicking it should create a page with the same name. If the page exists then it should link to the page. 

**Asana Issue**
https://app.asana.com/0/1121331595459324/1128677555115569/f


houshuang (changes requested):
Currently impossible to create a link that includes a @ tag (but you can select one that already exists). If this is hard to fix, never mind, if it's easy, would be nice. (10 minutes max :))

PR #1747, opened 2019-06-26
Export activities graph->wiki

houshuang:
This PR reimplements the functionality to export FROG activities/graphs to the wiki:

- create a graph with multiple activities. Make sure some students are part of the graph (for example by opening "4 students")
- you can export activities individually by clicking on them in the orchestration view, or the entire graph from the menu. You can export to an existing wiki or create a new wiki. 

It should correctly export all activities that have already run, on p1/p2/p3. When accessing in the wiki, it should not create new instances - if you are a user that were not part of the graph, it should show a message and let you select instances, otherwise it should directly drop you into the correct instance. 

Known bug (I will create separate Asana task for this): It does not check if a wiki page already exists with the same name, and will create two pages with the same name.


frog/imports/client/Wiki/index.js
```diff
@@ -279,7 +279,13 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
     if (!fullPageObj || fullPageObj.plane === 3) return fullPageObj;
 
     const instanceObj = fullPageObj.instances[instanceId];
-    if (!instanceObj) return null;
+    if (!instanceObj) {
+      if (fullPageObj.noNewInstances) {
```
qgolsteyn, line 283: Nit: perhaps merge the two if statements

frog/imports/client/Wiki/index.js
```diff
@@ -279,7 +279,13 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+        return fullPageObj;
+      } else {
+        return null;
+      }
+    }
 
     const mergedObj = Object.assign({}, fullPageObj, instanceObj);
```
qgolsteyn, line 290:
```suggestion
    const mergedObj = {...fullPageObj, ...instanceObj};
```

frog/imports/client/Wiki/index.js
```diff
@@ -389,7 +407,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
           },
           () => {
             if (foreignInstanceId) return;
-            const instanceName = Meteor.user().username;
+            const instanceName = this.getInstanceNameForUser(fullPageObj);
```
qgolsteyn, line 410: `getInstanceNameForUser` can return null, maybe we need a null check here

frog/server/wiki.js
```diff
@@ -25,13 +30,11 @@ const exportSessionWiki = (sessionId, wiki, userId) => {
...
     const obj = Objects.findOne(act._id);
-    if (act.plane === 3) {
-      importWikiFromFROG(act, obj, wiki, act.title, userId);
-    }
+    if (obj) importWikiFromFROG(act, obj, wiki, act.title, userId);
   });
 };
 
-export const importWikiFromFROG = async (item, object, wiki, page, userId) => {
+export async function importWikiFromFROG(item, object, wiki, page, userId) {
```
qgolsteyn: NIT: could use arrow function

frog/server/wiki.js
```diff
@@ -69,29 +72,41 @@ export const importWikiFromFROG = async (item, object, wiki, page, userId) => {
...
+      instanceData.all?.liId || instanceData[userId]?.liId || null,
       item.plane,
+      item.plane === 3
+        ? undefined
+        : Object.keys(instanceData)
+            .filter(key => key !== userId)
+            .reduce((obj, key) => {
+              obj[key] = instanceData[key];
+              return obj;
+            }, {}),
```
qgolsteyn, line 106: NIT assign to variable to help with readability

qgolsteyn (changes requested):
Left some comments

frog/imports/client/Wiki/index.js
ajainuary: In theory, it is possible for it to return `null` but it would happen only if the page is p3, and in that case, we will never call the function. Therefore in practice, it should never cause any issues

ajainuary (approved):
**Testing Done**
Other than the naming issue, everything works as expected

PR #1749, opened 2019-06-26
Enter now works from anywhere in create modal in the wiki

RachitMalik12:
This PR fixes the problem of enter not always working in the create modal. 

**How to test**

- Create a new wiki page, type a title and go configure the page by expanding the modal. Hit enter and the page will be created if there are no errors. 

**Asana issue**
https://app.asana.com/0/1121331595459324/1125445940645551/f

### DIFF
--- a/frog/imports/client/Wiki/ModalCreate.js
+++ b/frog/imports/client/Wiki/ModalCreate.js
@@ -169,10 +169,8 @@ class NewPageModal extends React.Component<PropsT, StateT> {
         }}
         onEscapeKeyDown={() => this.props.setModalOpen(false)}
         onKeyDown={e => {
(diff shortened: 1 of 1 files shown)


houshuang (changes requested):
Does not work for me. Type down, tab tab until you get to the Individual/All pages selector, select individual pages and press enter. No result. Go to Component, choose Brainstorm etc, enter, no response.

houshuang:
Merging since I think this is probably better than nothing, and doesn't break anything. We can look at it again later if we have lots of time :)

qgolsteyn:
Added a task for future reference: https://app.asana.com/0/1121331595459324/1129065858488801

PR #1750, opened 2019-06-26
Fix anonymous users in wiki, and updating user account with ShareDB

houshuang:
This PR adds a way of updating the current user in the ShareDB module after the websocket has connected. This is a bit of a hack - we make a specific fetchQuery (we could have made an op, but then we would need a document, fetchQuery can be done directly on the connection), but then intercept it using middleware and instead update the userid. 

I also made it possible to rename the wiki user. This is in anticipation of all the RFC work around accounts, just a quick fix. And I made sure that each user in the wiki is actually different (previously all anonymous users were the same user). 

If you wish to log out the user when you are testing, you can open dev tools, go to Application/Local Storage, and delete the data. (Or add ?reset=true to the URL)

Desired functionality:
- the user in ShareDB is always up to date, even when logging in and logging out without reloading the page, when being automatically logged in as an anonymous user, etc. (You can see this on the op.m.userid field in o_li, or in the Revisions of the wiki).
- when a user is anonymous, we always show "Anonymous User" and never the uuid
- a user can try to change their username. if the username exists, they get a warning, and the displayed username stays Anonymous...
- when a user changes their username, you can see the new username in Revisions etc.
- when two users logs in anonymously in two different browsers, they are logged in as two different anonymous users

Known bugs:
The intercepted fetchQuery results in a console.error because we are not properly catching the error. This doesn't seem to harm anything else, I will add an Asana task to look into this in the future. 

Other concerns:
- I modified the ChangeableText component to be able to place the pen on the left, to correctly hide until hover, and to cancel updating if the onChange returns false

### DIFF
--- a/frog-utils/src/TextInput.js
+++ b/frog-utils/src/TextInput.js
@@ -12,7 +12,8 @@ type TextInputPropsT = {
   clearOnEnter?: boolean,
   style?: Object,
   noBlur?: boolean,
-  placeholder?: string
+  placeholder?: string,
+  penOnLeft?: boolean
 };
 
 export class TextInput extends React.Component<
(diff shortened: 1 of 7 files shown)


ajainuary (approved):
**Testing Done**
Tested with multiple scenarios (on Google Chrome, Linux) and some minor issues were documented.

Works fine!

houshuang:
@ajainuary 	Thanks for your careful testing and documenting the issues you found.

PR #1752, opened 2019-06-27
Function to list pages from a Wiki

ajainuary:
**Description**
This is a minor PR which introduces a function to list all the pages of a wiki.

**How to test?**
This would be used in future PRs, but could be tested by calling `listPages()` from `frog/imports/client/Wiki/helpers.js`

### DIFF
--- a/frog/imports/client/Wiki/helpers.js
+++ b/frog/imports/client/Wiki/helpers.js
@@ -106,7 +106,7 @@ const listWikis = async () => {
 };
 
 /**
- * Function to get the pages of all pages in a wiki
+ * Function to get the pages of all (valid) pages in a wiki
  * @param {wikiDoc} ShareDB document corresponding to the Wiki
  * @return{Array} Array of pages in the form of [Title, ID] 
  */
--- a/frog/imports/client/Wiki/helpers.js
+++ b/frog/imports/client/Wiki/helpers.js
@@ -108,13 +108,12 @@ const listWikis = async () => {
 /**
  * Function to get the pages of all (valid) pages in a wiki
  * @param {wikiDoc} ShareDB document corresponding to the Wiki
- * @return{Array} Array of pages in the form of [Title, ID] 
+ * @return{Promise} A Promise that resolves into an array of pages in the form of [Title, ID] 
  */
 const listPages = wikiDoc => {
   return new Promise((resolve, reject) =>
     wikiDoc.fetch(err => {
       if (err) {
-        // eslint-disable-next-line no-console
         reject(Error('Unable to fetch the Wiki Document'));
       }
       resolve(
--- a/frog/imports/client/Wiki/helpers.js
+++ b/frog/imports/client/Wiki/helpers.js
@@ -107,10 +107,11 @@ const listWikis = async () => {
 
 /**
  * Function to get the pages of all (valid) pages in a wiki
- * @param {wikiDoc} ShareDB document corresponding to the Wiki
+ * @param {string} wikiId: id of the wiki to list
  * @return{Promise} A Promise that resolves into an array of pages in the form of [Title, ID] 
  */
-const listPages = wikiDoc => {
+const listPages = (wikiId: string) => {
+  const wikiDoc = connection.get('wiki', wikiId);
   return new Promise((resolve, reject) =>
     wikiDoc.fetch(err => {
       if (err) {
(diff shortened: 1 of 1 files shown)


frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,35 @@ const listWikis = async () => {
   });
 };
 
+// Function to get the pages of all pages in a wiki, returns [Title, ID]
```
qgolsteyn: Would recommend Javadoc comments so that we are compatible with IDEs

frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,35 @@ const listWikis = async () => {
...
+const listPages = async wikiDoc => {
+  const list = await new Promise(resolve =>
+    wikiDoc.fetch(err => {
+      if (err) {
+        // eslint-disable-next-line no-console
+        console.log('Unable to fetch');
```
qgolsteyn: Instead of logging, we should reject the promise and provide an error message

frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,35 @@ const listWikis = async () => {
   });
 };
 
+// Function to get the pages of all pages in a wiki, returns [Title, ID]
+const listPages = async wikiDoc => {
+  const list = await new Promise(resolve =>
```
qgolsteyn: No need to async/await for this function, we can simply return the Promise.

frog/imports/client/Wiki/helpers.js
qgolsteyn: We should also specify this function return only valid pages.

qgolsteyn (changes requested):
Left some comments

frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,37 @@ const listWikis = async () => {
...
+/**
+ * Function to get the pages of all pages in a wiki
+ * @param {wikiDoc} ShareDB document corresponding to the Wiki
+ * @return{Array} Array of pages in the form of [Title, ID] 
+ */
+const listPages = wikiDoc => {
+  return new Promise((resolve, reject) =>
+    wikiDoc.fetch(err => {
+      if (err) {
+        // eslint-disable-next-line no-console
```
qgolsteyn: Remove

frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,37 @@ const listWikis = async () => {
   });
 };
 
+/**
+ * Function to get the pages of all pages in a wiki
```
qgolsteyn: The pages of all pages? Unclear.

frog/imports/client/Wiki/helpers.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/helpers.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/helpers.js
ajainuary: Made the respective changes

frog/imports/client/Wiki/helpers.js
```diff
@@ -105,12 +105,37 @@ const listWikis = async () => {
...
+ * @param {wikiDoc} ShareDB document corresponding to the Wiki
+ * @return{Array} Array of pages in the form of [Title, ID]
```
qgolsteyn: NIT: it technically returns a promise

ajainuary:
@qgolsteyn Thank you for your feedback, I have addressed all your comments.

qgolsteyn (changes requested):
Almost there :slightly_smiling_face:

frog/imports/client/Wiki/helpers.js
ajainuary: Addressed the comment

frog/imports/client/Wiki/helpers.js
ajainuary: Addressed the comment

frog/imports/client/Wiki/helpers.js
ajainuary: Addressed the comment

houshuang (changes requested):
I need a function that deals with wikidoc itself. The caller should not have to worry about sharedb etc. The signature should be listPages(wikiId: string).

ajainuary:
@houshuang Made the respective changes

houshuang (approved):

PR #1753, opened 2019-06-27
Activity to session function

houshuang:
From the browser, you can use 

```js
Meteor.call('create.graph.from.activity', 'ac-chat', {title: 'hi'}, 3, (err, succcess) => console.log(err,success))
```
(activityType, config object, plane - defaults to 3). 

The return value is an object with 4 keys: slug, activityId, graphId and sessionId.

Teacher should go to /teacher/orchestration/{slug}?login={teachername}, and student to /{slug}.


frog/imports/api/activities.js
```diff
@@ -93,11 +93,12 @@ export const findOneActivityMongo = (id: string) => {
 export const addActivity = (
   activityType?: string,
   data: ?Object = {},
-  id: string,
+  id?: string,
   configVersion: ?number,
   groupingKey: ?string,
   parentId: ?string
 ) => {
```
qgolsteyn, line 100: All fields of `addActivity` are Maybe type. If it is within the scope of this PR, we could identify some fields that are required to avoid a potential undefined bug.

frog/imports/api/activities.js
```diff
@@ -118,7 +119,7 @@ export const addActivity = (
     });
   else
```
qgolsteyn, line 120: NIT: add curly braces

frog/imports/api/activities.js
```diff
@@ -93,11 +93,12 @@ export const findOneActivityMongo = (id: string) => {
...
   activityType?: string,
   data: ?Object = {},
-  id: string,
+  id?: string,
   configVersion: ?number,
   groupingKey: ?string,
   parentId: ?string
 ) => {
+  const actId = id || uuid();
   if (id)
```
qgolsteyn, line 102: NIT: Add curly braces

frog/imports/api/activities.js
```diff
@@ -118,7 +119,7 @@ export const addActivity = (
     });
   else
     insertActivityMongo({
-      _id: uuid(),
```
qgolsteyn, line 121: Here, I would recommend keeping this line and returning the generated id. In the `if` branch above, we can return the id provided

frog/imports/api/graphs.js
```diff
@@ -15,6 +20,36 @@ import {
...
+export const createSessionFromActivity = (
+  activityType: string,
+  config: Object,
+  plane: number = 3
+): {
+  slug: string,
+  sessionId: string,
+  graphId: string,
+  activityId: string
+} | void => {
```
qgolsteyn, line 33: NIT: Could move to its own type to improve readability and to allow callers to import the type

frog/imports/api/sessions.js
```diff
@@ -227,7 +227,7 @@ export const updateOpenActivities = (
 export const removeSession = (sessionId: string) =>
   Meteor.call('flush.session', sessionId);
 
-const addSessionFn = (graphId: string, slug: string): string => {
+export const addSessionFn = (graphId: string, slug?: string): string => {
```
qgolsteyn, line 231: NIT: We could add a function comment describing why slug can be optional

frog/imports/client/TeacherView/OrchestrationView.jsx
```diff
@@ -82,7 +82,7 @@ class OrchestrationViewController extends React.Component<any, {}> {
           <Grid item xs={12} className={classes.maybeScaled}>
             <Grid container className={classes.subroot}>
               <Grid item xs={12}>
-                {visible ? (
+                {visible || session.singleActivity ? (
```
qgolsteyn, line 85: Because session is of type `any`, we have no guarantee that singleActivity would exist. If within the scope of this PR, we could add a prop type for this component specifying the type of the session object

frog/imports/client/TeacherView/SessionUtils.jsx
```diff
@@ -177,7 +177,7 @@ const SessionUtils = ({
         </Typography>
       </Grid>
       <Grid item xs={4} className={classes.textCenter}>
-        <DashToggle visible={visible} toggleVisible={toggle} />
+        {!session.singleActivity && <DashToggle visible={visible} toggleVisible={toggle} />}
```
qgolsteyn, line 180: Because session is of type any, we have no guarantee that singleActivity would exist. If within the scope of this PR, we could add a prop type for this component specifying the type of the session object

qgolsteyn (changes requested):
Left some comments

frog/imports/api/activities.js
houshuang: ???

frog/imports/client/App/index.js
```diff
@@ -28,6 +29,8 @@ import { LocalSettings } from '/imports/api/settings';
 import WikiRouter from '../Wiki/WikiRouter';
 import SingleActivity from '../SingleActivity';
 
+window.create = createSessionFromActivity
```
ajainuary: Why is the function attached to the window?

frog/imports/client/TeacherView/OrchestrationView.jsx
ajainuary: In that case, `undefined` would still be false-y and the logic should work fine.
I think no change is required here.

frog/imports/client/App/index.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import path from 'path';
+import { createSessionFromActivity} from '/imports/api/graphs'
```
ajainuary: This could be removed.

frog/imports/client/TeacherView/SessionUtils.jsx
ajainuary: With reference to my previous comment, I think no change is required here.

ajainuary (dismissed):
**Testing Done**
Tested with `ac-chat` only with the specified configuration, could be tested further after implementing single activity.
It seems good to me

frog/imports/client/App/index.js
ajainuary: (Was present for testing purposes)

PR #1754, opened 2019-06-28
Minimal FROG to speed development

houshuang:
This PR attempts to reduce reload times for FROG by removing all operators, reducing the number of activities, and moving all activity types inline. 

Please test by git clean -fdx, and initial_setup. All functionality should be there (with fewer activities/no operators). Editing front-end code including activity runners in activities should be fastest, just causing a refresh, but even editing backend/shared code should be significantly faster.

I am planning to merge this to develop and use this as our main development branch. I've tagged an allActivities release (the current develop) which we can always merge back in before sending to production. In the future we might automate this, for now let's see if this works well.


RachitMalik12 (changes requested):
Some issues: 

- Gallery doesn't load 

- The example data in all activity previews are not displayed properly 

Logs for build time: 
- Initial build time after git clean: 03:16
- Time to build after making a small UI change in the wiki (after hitting save): 35 s
- Time to build after making a small UI change in an activity ~ 45s

houshuang:
I fixed issues from @RachitMalik12 . Note that some learning items will show an error message, because they are not included anymore (in Preview). This is expected, and should not lead to crashes or other loss of functionality.

RachitMalik12 (approved):
Logs for build time: 
- Initial build 01:40 
- After modifying an activity: ~40 secs 
- After modifying the wiki: ~ 30 secs 
**_Testing completed:_**
- Tested activities: brainstorm, chat, gallery, quiz, ck- board. Tested default configs in Preview.  
Just thought I would reiterate and mention some things that look like bugs but are expected: 
   li-twitter and li-hypothesis are not included and thus in the preview with some pre loaded configs, these will not be displayed, this is expected behaviour.
Since operators are removed, you cannot run a graph. 
- Tested making UI changes to the wiki and noted significant speed improvements.

Looks good. Thanks for this :)

qgolsteyn (commented):
LGTM :+1:

PR #1756, opened 2019-07-01
Pulling wiki pages into graph

houshuang:
Let's you open p1/p3 wiki pages in FROG. The config should only let you select existing wikis/pages, and students should see the correct instance in the case of p1.

To test: 
- add Wiki activity in graph, and run graph
- can also use Preview

You probably need to run `npm start cleanCache` first, since it's a new activity type

Future work:
- only let you select p1 pages in a p1 activity, same for p3
- how to deal with p2?
- option to automatically open in edit mode
- option to allow following links?


frog/imports/client/Wiki/helpers.js
```diff
@@ -123,7 +123,11 @@ const listPages = (wikiId: string) => {
             key =>
               wikiDoc.data.pages[key].created && wikiDoc.data.pages[key].valid
           )
-          .map(key => [wikiDoc.data.pages[key].title, key])
+          .map(key => [
+            wikiDoc.data.pages[key].title,
+            key,
+            wikiDoc.data.pages[key].plane
```
houshuang, line 129: Adding plane for use in the future in the config, not using currently

frog/imports/internalActivities/ac-wiki/config.js
```diff
@@ -0,0 +1,116 @@
+// @flow
+import * as React from 'react';
+import Form from 'react-jsonschema-form';
+import { FormControl } from 'react-bootstrap';
+import { listWikis, listPages } from '/imports/client/Wiki/helpers';
+
+const SelectWiki = ({ formContext, onChange, value = '' }: any) => (
```
houshuang, line 7: It would probably be possible to combine SelectWiki and SelectPage, but easier to leave it separate for now.

ajainuary (commented):
**Testing Issues**
Even if no wiki page is selected, the config appears valid but when the graph is run, new Wikis without any `wikiId` get created (which cannot be accessed again).

PR #1757, opened 2019-07-01
Remove cyclic dependencies in `frog-utils`

qgolsteyn:
This PR removes cyclic dependencies in `frog-utils` by moving all function declaration outside of `index.js`.

**Out of scope**
This PR makes no attempt at commenting or adding type information to `frog-utils`. This should be the focus of a separate PR.

**Testing done**
Manual testing


qgolsteyn:
The CI seems to be broken due to another issue. Will push a separate PR to address this problem.

houshuang:
Looks good to me, let's merge and keep an eye if there are any issues emerging. Thanks for this work.

PR #1758, opened 2019-07-01
Implementing Single Activities via Sessions

ajainuary:
**Description**
This PR implements the Single Activity (accessible at `/single_activity`) by creating Graph sessions for the activities using the activity to session function introduced in #1747 
The file at `frog/imports/client/SingleActivity/index.js` is also refactored and broken into smaller files.

**How to test?**
Visit `/single_activity`


frog/imports/client/SingleActivity/Finish.js
```diff
@@ -0,0 +1,79 @@
...
+    <Card raised className={classes.card}>
+      <IconButton
+        onClick={() => props.onReturn()}
+        className={classes.back_button}
+      >
+        <ArrowBack />
+      </IconButton>
+      <Typography variant="h3" component="h2" className={classes.padded_text}>
+        You're all set, please share this link with the participants:{' '}
+        <a
```
houshuang: Can you extract all the URL calculation logic into a separate function? In fact it could even be in frog-utils, but at least it shouldn't be inthe JSX and it shouldn't be repeated.

houshuang:
I'm sometimes getting this. index.js:305 is the single_activity router in App/index.
Screenshot 2019-07-01 at 16 33 05

houshuang:
Right now there is no error checking of the config, meaning you can add a Quiz with no questions, which is invalid - if you try to go to the next step, it crashes... You should be showing the red error bar, and not allow user to proceed if the activity is invalid. We already do this in the wiki.

houshuang (changes requested):
Good start - see comments in code and below.

frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -0,0 +1,63 @@
...
+  const { classes, onSubmit } = props;
+  // You can add the permitted activities for the single activity here
+  const allowed = [
+    'ac-quiz',
+    'ac-ck-board',
+    'ac-chat',
+    'ac-brainstorm',
+    'ac-ranking',
+    'ac-video'
+  ];
```
qgolsteyn: Move to a constant file or `index.js` as it significantly affects SingleActivity behaviour

frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -0,0 +1,63 @@
...
+  const list = activityTypes.filter(x => allowed.includes(x.id));
```
qgolsteyn: This logic should live in its own file or in `index.js` and passed as a prop. Everytime React rerenders this component, this method will be called.

frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -0,0 +1,63 @@
...
+      </Typography>
+      <GridList cols="4" spacing="8">
+        {list.map(x => (
+          <a
+            href="#"
+            onClick={e => {
+              e.preventDefault();
+              onSubmit(x);
+            }}
+          >
```
qgolsteyn, line 48: Remove this element and place the onClick in the `GridListTile`

frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -0,0 +1,63 @@
...
+            <GridListTile key={x.id} classes={{ root: classes.tile }}>
+              <img
+                src={'/' + x.id + '.png'}
+                alt={x.id}
+                className={classes.icon}
+              />
+              <GridListTileBar
+                title={x.meta.name}
+                subtitle={<span>{x.meta.shortDesc}</span>}
```
qgolsteyn: Why do we need a `<span>` here?

frog/imports/client/SingleActivity/types.js
```diff
@@ -0,0 +1,14 @@
...
+
+export type StateT = {
+  stage: number,
+  activityType?: ActivityPackageT,
+  activity?: ActivityDbT,
+  slug?: string
+};
+
+export type PropsT = {
+  classes: Object
```
qgolsteyn, line 13: We can use the `style` constant defined in `style.js` to make this type more specific. For example:
``` javascript
import {style} from './style';
type StyleObjectT = $Keys<typeof style>;
```

frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -0,0 +1,63 @@
...
+import { activityTypes } from '/imports/activityTypes';
+import { type PropsT } from './types';
+
+/**
+ * The icon-based selection form for choosing the activity type
+ * @param {Function} onSubmit - Callback function called with the selection by the user.
+ */
+export default function ChooseActivityType(
+  props: {
+    onSubmit: Function
```
qgolsteyn, line 31: We can make this type more specific

frog/imports/client/SingleActivity/ConfigPanel.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * The final screen displayed after the creation of a single activity.
+ * @param {ActivityPackageT} activityType - The object containing the details of the activity to be configured
+ * @param {Function} onSubmit - Callback function for handling the 'Submit' button
+ * @param {Function} onReturn - Callback function for handling the 'Back' button
+ */
+export default class ConfigPanel extends React.Component<
+  {
+    activityType: ActivityPackageT,
+    onSubmit: Function,
```
qgolsteyn, line 25: We can make this type more specific

frog/imports/client/SingleActivity/ConfigPanel.js
```diff
@@ -0,0 +1,63 @@
...
+    onReturn: Function
```
qgolsteyn, line 26: We can make this type more specific

frog/imports/client/SingleActivity/Finish.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * The final screen displayed after the creation of a single activity.
+ * @param {Object} url - An object containing the paths of public and dashboard view
+ * @param {Function} onReturn - Callback function for handling the 'Back' button
+ */
+export default function Finish(
+  props: {
+    url: {
+      public: String,
```
qgolsteyn: You may want to use the `string` type instead, the primitive instead of the wrapper object

frog/imports/client/SingleActivity/Finish.js
```diff
@@ -0,0 +1,79 @@
...
+      dashboard: String
+    },
+    onReturn: Function
```
qgolsteyn, line 26: We can make this type more specific

frog/imports/client/SingleActivity/TopBar.js
```diff
@@ -0,0 +1,26 @@
...
+import Button from '@material-ui/core/Button';
+import { type PropsT } from './types';
+
+/**
+ * Navigation bar displayed at the top
+ */
+export default function TopBar(props: PropsT) {
+  const { classes } = props;
+  return (
+    <AppBar position="static" color="default">
```
qgolsteyn, line 18: Why `color="default"`?

frog/imports/client/SingleActivity/TopBar.js
```diff
@@ -0,0 +1,26 @@
...
+      <Toolbar classes={{ root: classes.navbar }}>
+        <Typography variant="h6" color="inherit" className={classes.logo}>
```
qgolsteyn, line 20: Why `color="inherit"`?

frog/imports/client/SingleActivity/index.js
```diff
@@ -263,22 +44,37 @@ class SingleActivity extends React.Component<PropsT, StateT> {
           <ConfigPanel
             activityType={activityType}
             classes={classes}
-            onSubmit={conf =>
-              this.setState({ stage: this.state.stage + 1, activity: conf })
-            }
+            onSubmit={conf => {
```
qgolsteyn, line 42: Move Meteor related logic to its own function to separate UI and state concerns.

frog/imports/client/SingleActivity/index.js
```diff
@@ -263,22 +44,37 @@ class SingleActivity extends React.Component<PropsT, StateT> {
...
+              this.setState({ stage: this.state.stage + 1, activity: conf });
+              Meteor.call(
+                'create.graph.from.activity',
+                conf.activityType,
+                conf.config,
+                3,
+                (err, res) => {
+                  if (err) {
+                    // eslint-disable-next-line no-console
+                    console.log(err);
```
qgolsteyn: We should throw instead of logging as this puts the application in a non-deterministic state.

frog/imports/client/SingleActivity/index.js
```diff
@@ -263,22 +44,37 @@ class SingleActivity extends React.Component<PropsT, StateT> {
...
-            onSubmit={conf =>
-              this.setState({ stage: this.state.stage + 1, activity: conf })
-            }
+            onSubmit={conf => {
+              this.setState({ stage: this.state.stage + 1, activity: conf });
+              Meteor.call(
+                'create.graph.from.activity',
+                conf.activityType,
+                conf.config,
+                3,
```
qgolsteyn, line 48: Magic number should be moved to a constant

frog/imports/client/SingleActivity/index.js
```diff
@@ -239,15 +28,7 @@ class SingleActivity extends React.Component<PropsT, StateT> {
     const { stage, activityType } = this.state;
```
qgolsteyn: Add `slug`

qgolsteyn (changes requested):
Good job! Just a couple comments regarding separation of concerns and code style.

frog/imports/client/SingleActivity/ChooseActivityType.js
houshuang: just export from index.js is enough.

houshuang: but if we export from index.js then we have circular dependency! :) I think keeping it in this file, but making it a constant at the top of the file is OK.

frog/imports/client/SingleActivity/index.js
houshuang: Better to show error message and let user try again? Currently it gets stuck on the loading screen.

houshuang (changes requested):

frog/imports/client/SingleActivity/ChooseActivityType.js
ajainuary: Made the respective changes

frog/imports/client/SingleActivity/index.js
ajainuary: Should I display a Modal in this case?

frog/imports/client/SingleActivity/Finish.js
ajainuary: I'll currently extract it into a variable in the same file but later on, it could be moved into `frog-utils`

frog/imports/client/SingleActivity/ChooseActivityType.js
ajainuary: Made the respective changes

frog/imports/client/SingleActivity/ChooseActivityType.js
ajainuary: Made the respective changes

frog/imports/client/SingleActivity/ChooseActivityType.js
ajainuary: No we don't :+1:

frog/imports/client/SingleActivity/types.js
ajainuary: This is not the most appropriate solution, a better one would be to use a package like `css-modules-flow-types-loader`. Since this is not a critical issue, I'll skip this one for now.

frog/imports/client/SingleActivity/index.js
ajainuary: Made the respective changes

ajainuary:
@houshuang and @qgolsteyn Thank you for your valuable comments, I have addressed them in my latest commit.

houshuang (approved):

PR #1759, opened 2019-07-02
Splitting collections out of /imports/api

qgolsteyn:
**NON FUNCTIONAL - DO NOT MERGE**

This PR is a first attempt at splitting Mongo collections away from the api folder, to reduce cyclic dependencies, to simplify the dependency graph, and to work towards separating controller logic from the persistence layer.

A large number of files depend on the Mongo collections, from React components to server functions, making it difficult to perform this operation with confidence. At this point, FROG is unstable but is able to compile.


qgolsteyn:
dependencygraph (1)

PR #1760, opened 2019-07-02
Adding templates to FROG

houshuang:
I know I'm doing a lot of hacky things here. I will probably change/normalize how I define templates in the future, and some of the way I process them might be improved. For now, I'm just worried about breaking anything else, or having bugs.


frog/imports/client/SingleActivity/ChooseActivityType.js
```diff
@@ -21,6 +22,7 @@ const allowed = [
   'ac-video'
 ];
 const list = activityTypes.filter(x => allowed.includes(x.id));
+list.push(templatesObj['te-peerReview']);
```
ajainuary, line 25: Could be changed from hard coding names to including it in the `allowed` array

ajainuary (commented):
Functionality-wise this PR works as expected.

The logic could be more generalized (replacing hard-coding of template names).
Also, thumbnails could be moved into their own folder instead of having them in the `public` folder. (I will be working on a separate PR for that)

frog/imports/internalTemplates/te-peerReview/index.js
```diff
@@ -0,0 +1,83 @@
...
+      title: 'How many items should each student/group review',
+      type: 'number',
+      default: 1
+    },
+    reviewPrompt: { title: 'Review prompt', type: 'rte' },
+    reviseInstructions: { title: 'Revision prompt', type: 'rte' }
+  }
+};
+
+// showAll: {
```
ajainuary, line 26: Should commented out code be discarded?

PR #1761, opened 2019-07-03
Trimming of leading, trailing and extra spaces on the Wiki page titles

RachitMalik12:
This PR fixes the issue of an extra leading and trailing space sometimes inserted at the end of the wiki page title. 

**Testing instructions**

- Create a wiki page title with leading,  trailing and extra spaces between the words using either the CreateModal, RRT links or renaming an existing page. 

- Now, the spaces at the beginning, the end and extra spaces between words of the title will be removed. (see the URL bar) 

**Asana issue**
https://app.asana.com/0/1121331595459324/1129368571245493/f

### DIFF
--- a/frog-utils/src/browserOnlyComponents.js
+++ b/frog-utils/src/browserOnlyComponents.js
@@ -1,5 +1,5 @@
 // @flow
-/***
+/** *
  * This file specifies a number of components that should only be available when
  * frog-utils is accessed from a browser instance. If this is not the case, the
  * components will be replaced with dummy alternatives.
--- a/frog-utils/src/browserOnlyComponents.js
+++ b/frog-utils/src/browserOnlyComponents.js
@@ -1,5 +1,5 @@
 // @flow
-/** *
+/***
  * This file specifies a number of components that should only be available when
  * frog-utils is accessed from a browser instance. If this is not the case, the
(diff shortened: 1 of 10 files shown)


frog-utils/src/browserOnlyComponents.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-/***
+/** *
```
houshuang: क्यूं?

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -83,15 +87,17 @@ class WikiContentComp extends React.Component<> {
 
     if (
       error === null ||
-      pageTitleString === this.props.currentPageObj?.title
+      sanitizeTitle(pageTitleString) === this.props.currentPageObj?.title
     ) {
       this.props.changeTitle(this.props.currentPageObj.id, pageTitleString);
```
houshuang, line 92: shouldn't you be sanitizing the pageTitleString that you are saving?

houshuang (approved):

frog-utils/src/browserOnlyComponents.js
RachitMalik12: Es lint.. :(

PR #1762, opened 2019-07-03
Login with User ID

ajainuary:
**Description**
This PR introduces a feature to login via `userid` param in the GET Request. The intended use is in the single activity where we wish to provide URLs to the anonymous user for viewing the dashboard. This PR also allows the Anonymous Visitor to access the TeacherView. It also makes the Public URL in the Finish section of Single Activity non-clickable.
Finally, this PR places the Single Activity at the root (`/`)

**Resolved Issues**
Referenced Asana Tasks:
https://app.asana.com/0/1115664036175110/1129513551042844
https://app.asana.com/0/1115664036175110/1129513551042843
https://app.asana.com/0/1115664036175110/1129522585812163
https://app.asana.com/0/1121313931987677/1129523836089717
https://app.asana.com/0/1121313931987677/1129755058291934

**How to test?**
***Single Activity***
Access `/single_activity` without login and create an activity. The dashboard URL will contain `?userid=` param. Use a separate session / window to access this URL, you should still be able to access your dashboard. (Without any prompt for logging in)
***Public URL***
The student URL should not be clickable
***URL***
Going to `/` should now load the Single Activity View


frog/imports/client/App/TopBar.js
```diff
@@ -115,7 +115,7 @@ class TopBarController extends React.Component<
 > {
   routes = [
     { name: 'Graph Editor', to: '/teacher/graph' },
-    { name: 'Sessions', to: '/teacher/orchestration' },
+    { name: 'Sessions', to: '/t/' },
```
qgolsteyn: Paths should be moved to a constant as we are having to modifying them in more than one place. If we forget an edit, it will break FROG

frog/imports/client/App/index.js
```diff
@@ -247,9 +266,9 @@ const FROGRouter = withRouter(
               }
             }
             if (
-              this.props.match.params.slug &&
-              (this.props.match.params.slug.slice(0, 4) === 'wiki' ||
-                this.props.match.params.slug.slice(0, 15) === 'single_activity')
+              !this.props.match.params.slug ||
+              this.props.match.params.slug.slice(0, 4) === 'wiki' ||
+              this.props.match.params.slug.slice(0, 1) === 't'
```
qgolsteyn: Here, we should use regex to avoid having to rely on specific string length. What happens if the slug starts with a "t" but should not lead to a login?

frog/imports/client/App/TopBar.js
qgolsteyn: Careful, this path is the only path with a `/` at the end of the string. Make sure this is not a typo

frog/imports/client/SingleActivity/Finish.js
```diff
@@ -37,11 +37,13 @@ function Finish(
       </IconButton>
       <Typography variant="h3" component="h2" className={classes.padded_text}>
         You're all set, please share this link with the participants:{' '}
-        <Link to={'/' + url.public}>{BASE_URL + '/' + url.public}</Link>
```
qgolsteyn, line 40: Removing the link tag would not allow the user to click on the link

frog/imports/client/SingleActivity/Finish.js
```diff
@@ -37,11 +37,13 @@ function Finish(
       </IconButton>
       <Typography variant="h3" component="h2" className={classes.padded_text}>
         You're all set, please share this link with the participants:{' '}
```
qgolsteyn, line 42: Remove `{' '}` and simply add a space

frog/imports/client/SingleActivity/index.js
```diff
@@ -71,7 +71,7 @@ class SingleActivity extends React.Component<{}, StateT> {
           <Finish
             url={{
               public: slug,
-              dashboard: 'teacher/orchestration/' + slug
+              dashboard: 't/' + slug
```
qgolsteyn, line 74: We should have a constant for the path

frog/imports/client/StudentView/SessionBody.jsx
```diff
@@ -90,7 +90,7 @@ const StudentView = withRouter(({ activities, session, classes, history }) => (
               className={classes.button}
               color="inherit"
               onClick={() => {}}
-              href="/teacher/orchestration"
+              href="/t/"
```
qgolsteyn: Caution, you are adding a `/` at the end of the URL which wasn't there before. Make sure this is intentional

qgolsteyn (changes requested):
Left some comments. A common pattern which would benefit from refactoring would be to move the `/t/` path to a constant as it is specified many times across many files. This approach would be less error prone as we would only have to change the path once

frog/imports/client/SingleActivity/Finish.js
ajainuary: This was intentional, teachers should not click the student URL.

frog/imports/client/SingleActivity/Finish.js
ajainuary: https://stackoverflow.com/questions/38547558/why-do-people-put-in-their-react-jsx

frog/imports/client/App/TopBar.js
ajainuary: Both are equally valid, will change for consistency

frog/imports/client/SingleActivity/index.js
ajainuary: This is not in the scope of the PR

frog/imports/client/App/TopBar.js
ajainuary: This is not in the scope of the PR

frog/imports/client/StudentView/SessionBody.jsx
ajainuary: Both are equally valid, will change for consistency

qgolsteyn:
@ajainuary 	I would argue that adding a constant for path `/t` should be a part of this PR due to the large number of edits to path related logic

PR #1763, opened 2019-07-04
Add first and second level navigation for activity configuration

qgolsteyn:
This PR adds first and second level navigation for activity and operator configuration. This is accomplished by adding a number of new UI to the ObjectField, conditional on its position within the configuration schema. This means that the default behaviour is backward-compatible with how FROG runs today.

_Important notes_
This PR upgrades material-ui meaning there could be changes in UI across FROG

**Testing done**
https://5d1cc7940fd48d596a62ea7c--boring-cray-c77b3b.netlify.com/

**Next steps**
- Preview page has its UI broken. We recommend revisiting this page to fix this issue and to better incorporate the changes in configuration UI.


houshuang (changes requested):
You knew about the extra whitespace in Preview, but also notice that one of the buttons have been pushed off the viewport.

Screenshot 2019-07-04 at 12 23 36

Should the entire header (General ... ) be clickable to unroll, rather than just the button (that's what I would expect).

How do we specify automatically unrolling a section? If not, I would default to unrolling everything for now. 

Also while you're at it, if you can quickly update the default width of the unrolled ModalCreate in the wiki so that it stays fixed between the three tabs.
Thanks!

qgolsteyn:
@houshuang Addressed these changes

frog-utils/src/FrogForm/components/PrimaryGrouping.jsx
```diff
@@ -0,0 +1,84 @@
...
+        {title && <Typography variant="h6">{title}</Typography>}
+        {description && (
+          <Typography className={classes.description} variant="body2">
+            {description}
+          </Typography>
+        )}
+        <ExpandMore
+          className={classes.expandIcon}
+          style={{
+            transform: expand ? 'rotate(180deg)' : undefined
```
ajainuary: Did you intend to have no transform in case `expand` is false? `transform: none` would be more suitable than `undefined`.

frog-utils/src/FrogForm/index.js
```diff
@@ -0,0 +1,3 @@
+import FrogForm from "./FrogForm";
```
ajainuary: Why not change the file name?

ajainuary (dismissed):
Looks better than existing config, I think sidebar space can be more efficiently utilized
Space Waste

frog-utils/src/FrogForm/index.js
qgolsteyn: Changed, makes more sense

frog-utils/src/FrogForm/components/PrimaryGrouping.jsx
qgolsteyn: You're right, changed to `rotate(0deg)` to improve readability

PR #1767, opened 2019-07-05
Allow sharing/duplicating activities/templates

houshuang:


frog/imports/client/SingleActivity/ConfigPanel.js
```diff
@@ -21,9 +21,11 @@ import { style } from './style';
  */
 class ConfigPanel extends React.Component<
   {
+    name: string,
     activityType: ActivityPackageT,
     onSubmit: Function,
-    onReturn: Function
+    onReturn: Function,
+    data?: Object
```
qgolsteyn, line 28: What is the format of data?

frog/imports/client/SingleActivity/Finish.js
```diff
@@ -40,14 +39,20 @@ function Finish(
...
         You're all set, please share this link with the participants:{' '}
-        {BASE_URL + '/' + url.public}
+        {BASE_URL + '/' + slug}
       </Typography>
       <Typography variant="h4" component="h3" className={classes.padded_text}>
         Here's a secret link to your behind-the-scene view of the dashboard:{' '}
-        <Link to={'/' + url.dashboard + '?u=' + Meteor.userId()}>
-          {BASE_URL + '/' + url.dashboard + '?u=' + Meteor.userId()}
+        <Link to={'/t/' + slug + '?u=' + Meteor.userId()}>
+          {BASE_URL + '/t/' + slug + '?u=' + Meteor.userId()}
```
qgolsteyn, line 47: We should move `/t` to a constant to simplify future work. A change to the path could break this link.

frog/imports/client/SingleActivity/index.js
```diff
@@ -9,20 +10,38 @@ import ConfigPanel from './ConfigPanel';
...
+          activityTypes.find(x => x.id === data.activityType) ||
+          templatesObj[data.activityType];
+        this.setState({
+          stage: 2,
+          activityType: type,
+          config: data.config
+        });
+      }
+    });
+  };
```
qgolsteyn, line 45: We should isolate this call outside of `index.js` to limit the number of responsibilities SingleActivity has.

qgolsteyn (changes requested):
Left some comments

frog/imports/client/SingleActivity/ConfigPanel.js
houshuang: This is the result of react-jsonschema-form for operators/activities, so Object is the most specific we can be.

frog/imports/client/SingleActivity/Finish.js
houshuang: Let's add another task to think about constants and URLs. Where should those constants be defined and exported from?

houshuang:
@qgolsteyn I've tried to address your comments, OK to merge?

qgolsteyn (approved):
LGTM :+1:

PR #1768, opened 2019-07-05
Accounts in FROG - functions to create a new user and login with password 

RachitMalik12:
This PR introduces 2 functions to create a new 'verified' user with a password and email as well as login to that user. We are no longer using the unique username but the email will now be unique and  we will have a profile object for each user that will have a field for DisplayName which is not unique but will be displayed on the UI. Please note in the Meteor.users collection,  the email field is an array of objects (with 2 fields: address and verified (bool)). Verified is used to indicate whether the email is verified but **has nothing to do** with the concept of a verified user in FROG. 

**Meteor Packages used:**

1. accounts-base 
2. accounts-password

**Background**

- There are 2 cases when you want to create a verified account.
  a) By default every user that goes to the landing page of FROG is logged in as an anonymous user. After using the product the user might decide to make an account with us. So when they fill out their information (email and password), we simply use the same anonymous userId and update it with their information. In this case we pass the userId of the anonymous user as a parameter to the create account function. 
b) If they are not anonymous and they want to create another account, then we create a new user in the users collection. 

**Testing instructions**
1. **Account creation**
For now these functions will be tested from the console. 

_Case 1_:

- Navigate to localhost:3000. Note the userid of the anonymous user. 

- Call `Meteor.call('create.account',Meteor.userId(),'testemail@gmail.com','somepassword',{DisplayName:'UserTester'},(err,succ) => console.log(err,succ))`
- The database should have the same userId noted in step 1, but with the updated email, password, profile information 

_Case 2_:
- Log in with a non anonymous user by using `Meteor.loginWithPassword('someuser@gmail.com', 'somepassword')`. Note the userId. 
- Call `Meteor.call('create.account',null,'testemail2@gmail.com','someotherpassword',{DisplayName:'UserTester2nd'}, (err,succ) => console.log(err,succ))`. We are passing null as the userId since the user is not anonymous and we want to create a new user. (This should work even without passing null since we have a check for anonymous user in the function)
- In the `Metoer.users` collection, there should be a new user with a different userId than before. 

**2.Logging in with password**
- Simply call `Meteor.loginWithPassword('email','password', (err,succ) => console.log(err,succ))` from the console. This function is meant to be used only on the client so it will be used directly on the client when we implement the front-end components . 
- Verify that the user is logged in by checking localStorage or using `Meteor.userId() or Meteor.user().emails[0].address`


frog/.meteor/packages
```diff
@@ -34,3 +34,4 @@ kadira:dochead
 mizzao:timesync
 mizzao:user-status
 underscore
+alanning:roles
```
houshuang: What are we using this for?

frog/server/accountManagement.js
```diff
@@ -0,0 +1,45 @@
...
+    Accounts.createUser({
+      email,
+      password,
+      profile,
+      isAnonymous: false
+    });
+  }
+};
+
+export const loginUser = (email: string, password: string) => {
```
houshuang: Do we need these 2 functions?

houshuang (dismissed):
LGTM - check 2 things I noted

frog/.meteor/packages
RachitMalik12: Resolved in latest commit

houshuang (changes requested):
You should add a check for whether email is valid, and return an error - currently it crashes the server if email already exists: 

```js
 Exception while invoking method 'create.account' { MongoError: E11000 duplicate key error collection: meteor.users index: emails.address_1 dup key: { : "testemail@gmail.com" }
I20190705-16:01:22.528(2)?     at Function.create (/Users/stian/.meteor/packages/npm-mongo/.3.1.2.vlpulc.v4258++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb-core/lib/error.js:43:12)
I20190705-16:01:22.529(2)?     at toError (/Users/stian/.meteor/packages/npm-mongo/.3.1.2.vlpulc.v4258++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/lib/utils.js:149:22)
I20190705-16:01:22.529(2)?     at coll.s.topology.update (/Users/stian/.meteor/packages/npm-mongo/.3.1.2.vlpulc.v4258++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/lib/operations/collection_ops.js:1414:39)
I20190705-16:01:22.529(2)?     at /Users/stian/.meteor/packages/npm-mongo/.3.1.2.vlpulc.v4258++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb-core/lib/connection/pool.js:532:18
I20190705-16:01:22.529(2)?     at _combinedTickCallback (internal/process/next_tick.js:132:7)
I20190705-16:01:22.529(2)?     at process._tickDomainCallback (internal/process/next_tick.js:219:9)
I20190705-16:01:22.530(2)?   driver: true,
I20190705-16:01:22.530(2)?   name: 'MongoError',
I20190705-16:01:22.530(2)?   index: 0,
I20190705-16:01:22.530(2)?   code: 11000,
I20190705-16:01:22.530(2)?   errmsg: 'E11000 duplicate key error collection: meteor.users index: emails.address_1 dup key: { : "testemail@gmail.com" }',
I20190705-16:01:22.531(2)?   [Symbol(mongoErrorContextSymbol)]: {} }
```

houshuang:
We should never crash the server because it disrupts everyone else. We can return an error code, and then the UI can say "invalid email" etc.

frog/.meteor/packages
```diff
@@ -20,7 +20,7 @@ shell-server@0.3.1            # Server-side component of the `meteor shell` comm
 
 insecure@1.0.7                # Allow all DB writes from clients (for prototyping)
 accounts-ui@1.3.0
-accounts-password@1.5.1
+accounts-password
```
qgolsteyn, line 23: Nit: specify which version we are using to avoid version related issues in the future

frog/server/accountManagement.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param: {string}, email - email to be associated with the acount
+ * @param: {string}, password - password of the new account
+ * @param: {Object}, profile - user profile which consists of the DisplayName
+ */
+
+export const createAccount = (
+  userid: string,
+  email: string,
+  password: string,
+  profile: Object
```
qgolsteyn: Nit: we should make this `Object` more specific

frog/server/accountManagement.js
```diff
@@ -0,0 +1,50 @@
...
+) => {
+  const user = Meteor.user();
+
+  if (userid && user?.isAnonymous) {
+    // checks for duplicate email
+    if (!Accounts.findUserByEmail(email)) {
```
qgolsteyn, line 24: What do we do if we find a duplicate email. Should we notify the user?

frog/server/accountManagement.js
```diff
@@ -0,0 +1,50 @@
...
+      });
+      Accounts.setPassword(userid, password, { logout: false });
+    }
+  } else {
+    Accounts.createUser({
+      email,
+      password,
+      profile
+    });
+    const userId = Accounts.findUserByEmail(email);
```
qgolsteyn: Caution: we are initialising a variable that has a name very similar to the parameter `userid`. Rename the variable to avoid confusion.

frog/server/accountManagement.js
```diff
@@ -0,0 +1,50 @@
...
+
+export const createAccount = (
+  userid: string,
+  email: string,
+  password: string,
+  profile: Object
+) => {
+  const user = Meteor.user();
+
+  if (userid && user?.isAnonymous) {
```
qgolsteyn: `userid` is always truthy

frog/server/accountManagement.js
```diff
@@ -0,0 +1,50 @@
...
+  } else {
+    Accounts.createUser({
+      email,
+      password,
+      profile
+    });
+    const userId = Accounts.findUserByEmail(email);
+    Meteor.users.update(userId, {
+      $set: {
+        isAnonymous: false
```
qgolsteyn, line 48: Nit: both branches make the user not anonymous. Could move this outside of the conditional block to keep the code DRY

qgolsteyn (changes requested):
Left some comments

frog/server/accountManagement.js
RachitMalik12: Yes we will display an error message on the client. I will move most of this logic to the client when I build the UI.

frog/server/accountManagement.js
RachitMalik12: Fixed

frog/server/accountManagement.js
RachitMalik12: resolved.. talked to you directly :)

frog/server/accountManagement.js
RachitMalik12: resolved in latest commit

frog/server/accountManagement.js
RachitMalik12: Thanks for that! Fixed in latest commit.

frog/.meteor/packages
RachitMalik12: I ran meteor-add and for some reason it removed the version..

qgolsteyn: Weird... I'll leave it up to you to fix it if you think it is important

frog/server/accountManagement.js
qgolsteyn: However this is still a case that might happen at the server level (as you implemented a conditional check for this exact scenario). I would suggest outputting a console.err to indicate this is an unwanted state.

qgolsteyn (approved):
LGTM :+1:

PR #1772, opened 2019-07-09
[Wiki Settings] Option for Read-only wiki

ajainuary:
**Description**
This is a minor PR which adds a settings field to the `wikiDoc` containing a `bool readOnly` which if set, will cause the wiki to be read-only.

**Resolved Issues**
This PR builds toward the Wiki Ownership feature.

**How to test?**
Cannot be tested via Browser, need to manually make changes to the database.

**Caveat**
This PR disables the option of editing the wiki but does not include any access-control functionality to prevent unauthorized access.

### DIFF
--- a/frog/imports/client/Wiki/WikiContentComp.js
+++ b/frog/imports/client/Wiki/WikiContentComp.js
@@ -136,7 +136,8 @@ class WikiContentComp extends React.Component<> {
     const docModeButton = () => {
       if (
         this.state.docMode === 'history' ||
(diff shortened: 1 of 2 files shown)


houshuang:
OK, but you should probably just have state.settings instead of having to individually set state.readonly state.this and state.that?

ajainuary:
In this case, no new field has been introduced in the state. A `readonly` value has been added to the `docMode` enum.

houshuang (approved):

PR #1774, opened 2019-07-09
Verify that Cypress works in CI

qgolsteyn:
This PR adds two small UI tests on the wiki and modifies the CI script to only run Cypress. This PR should set the foundation for more E2E tests.

**Testing done**
Cypress


houshuang:
Do you have any idea why it still doesn't work in Cypress interactive mode?

houshuang:
@qgolsteyn 	Can you merge this with the Eslint changes, and I think we can accept?

qgolsteyn:
This PR is now synced with develop

PR #1782, opened 2019-07-10
Introducing Git hooks

qgolsteyn:
This PR introduce two git hooks to run our test script when:
- commiting
- pushing

Let's discuss if we need to change this behaviour during the code review, but I believe introducing this change will help us avoid some common issues moving forward.


qgolsteyn:
Removed the commit hook, leaving only the push hook

PR #1784, opened 2019-07-11
Wiki Permissions

ajainuary:
**Description**
This PR completes the Wiki Ownership / Settings features:
1. Locking the Wiki
2. Making Wiki Read only
3. Requiring password to view / edit

**How to test?**
Open the Wiki Settings modal from Secondary Nav items in the Top Navbar in the wiki


frog/imports/client/Wiki/ModalPassword.js
```diff
@@ -0,0 +1,60 @@
...
+  callback,
+  hideModal
+}: ModalPasswordPropsT) => {
+  const [password, setPassword] = useState('');
+  const [error, setError] = useState(false);
+  const cancelAction = () => {
+    callback(false);
+    hideModal();
+  };
+  const validatePassword = async () => {
```
houshuang: you can remove the async now

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,142 @@
...
+    margin: '2vh'
+  },
+  label: {
+    paddingTop: '2vh'
+  },
+  helperText: {
+    display: 'inline',
+    paddingLeft: '1vw',
+    fontSize: '1rem',
+    letterSpacing: '0.001em'
```
houshuang: is this necessary?

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,142 @@
...
+  },
+  select: {
+    margin: '1vh'
+  }
+};
+
+export default withStyles(styles)(
+  ({ callback, hideModal, currentSettings, classes }: ModalSettingsPropsT) => {
+    const [settings, setSettings] = useState(currentSettings);
+    const handleChange = target => name => event => {
```
houshuang: I guess it doesn't matter, but why not pass the target, name and event as three props? Isn't this generating a bunch of new functions every time it gets called? (Again I know this won't impact performance here but still)

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -146,7 +145,9 @@ class WikiContentComp extends React.Component<> {
             style={docModeButtonStyle}
             color="primary"
             onClick={() => {
-              this.setState({ docMode: 'edit' });
+              this.props.checkEdit().then(x => {
```
houshuang, line 155: can you use a more descriptive variable than x?

houshuang (changes requested):
- hide the wiki settings in the menu when user is not wikiOwner, rather than showing it and just showing a modal saying you can't change settings after clicking
- I can still see the content on the home page even if wiki is locked while I'm editing
- read-only is not watching for double-clicking
- you can still create pages by adding a link to a non-existing page and clicking on the link (it's in wiki/WikiLink, which listens to the store, so you can update a flag there)
- enter should work in the password dialog
- users in the wikiObj is not updated automatically
- password when set to "for viewing" is not recognized

frog/imports/client/Wiki/ModalPassword.js
ajainuary: It was an inadvertent error

frog/imports/client/Wiki/ModalSettings.js
ajainuary: Removed `letter-spacing` but the other ones are necessary.

frog/imports/client/Wiki/ModalSettings.js
ajainuary: This is syntactically more concise and in our implementation generates the same number of function (1 for handling each change). 
This is because we would need to use arrow functions to pass the props otherwise.

frog/imports/client/Wiki/WikiContentComp.js
ajainuary: Thanks for pointing this out!

ajainuary:
@houshuang all comments have been addressed

houshuang:
@ajainuary 	Can you fix the merge conflict, and make CI green?

frog/imports/api/wikiDocHelpers.js
```diff
@@ -117,6 +117,14 @@ export const createNewEmptyWikiDoc = (wikiDoc, wikiId, liId, owner) => {
   const emptyDocValues = {
     wikiId,
     owners: Array.of(owner),
+    users: Array.of(owner),
+    editors: Array.of(owner),
+    settings: {
+      password: '',
```
qgolsteyn, line 125: Plain-text password is not a good idea

frog/imports/api/wikiDocHelpers.js
```diff
@@ -117,6 +117,14 @@ export const createNewEmptyWikiDoc = (wikiDoc, wikiId, liId, owner) => {
...
+      restrict: 'none',
```
qgolsteyn, line 126: What are the possible string values here?

frog/imports/api/wikiDocHelpers.js
```diff
@@ -141,3 +149,39 @@ export const completelyDeleteWikiPage = (wikiDoc, pageId) => {
 
   wikiDoc.submitOp(op);
 };
+
+export const addUser = (wikiDoc, userid) => {
```
qgolsteyn: Define parameter types

frog/imports/api/wikiDocHelpers.js
```diff
@@ -141,3 +149,39 @@ export const completelyDeleteWikiPage = (wikiDoc, pageId) => {
...
+  const op = {
+    p: ['users', 0],
+    li: userid
+  };
+  wikiDoc.submitOp(op);
+};
+
+export const addEditor = (wikiDoc, userid) => {
```
qgolsteyn: Define parameter types

houshuang:
We're getting there. But still a few small issues:

- You can still double-click to edit a Read Only wiki.
- You can still create new pages through creating a link to a non-existing page, even if "don't allow creation of new pages" is true.
- Password required modal should auto-focus on the password field.

frog/imports/api/wikiDocHelpers.js
```diff
@@ -141,3 +149,39 @@ export const completelyDeleteWikiPage = (wikiDoc, pageId) => {
...
+  const op = {
```
qgolsteyn, line 166: NIT: A comment explaining what we do would be useful, it isn't very clear directly from the code

frog/imports/api/wikiDocHelpers.js
```diff
@@ -141,3 +149,39 @@ export const completelyDeleteWikiPage = (wikiDoc, pageId) => {
...
+      p: ['editors'],
+      oi: []
+    };
+    wikiDoc.submitOp(opDropEditors);
+    const opDropUsers = {
+      p: ['users'],
+      oi: []
+    };
+    wikiDoc.submitOp(opDropUsers);
+  }
```
qgolsteyn: The content of this if statement should be moved to its own function. This function already does a lot.

frog/imports/api/wikiDocHelpers.js
```diff
@@ -141,3 +149,39 @@ export const completelyDeleteWikiPage = (wikiDoc, pageId) => {
...
+  wikiDoc.submitOp(op);
+};
+
+export const updateSettings = (wikiDoc, settings) => {
+  const op = {
+    p: ['settings'],
+    oi: settings
+  };
+  wikiDoc.submitOp(op);
+  if (wikiDoc.data.password !== settings.password) {
```
qgolsteyn: What is the purpose of this if statement?

frog/imports/client/Wiki/ModalPassword.js
```diff
@@ -0,0 +1,65 @@
...
+import { Modal } from './components/Modal';
+
+type ModalPasswordPropsT = {
+  actualPassword: string,
+  callback: boolean => void,
+  hideModal: () => void
+};
+
+export default ({
+  actualPassword,
```
qgolsteyn, line 15: Uhhh, we can't do this. Let's not pass plain-text passwords in props.

frog/imports/client/Wiki/ModalPassword.js
```diff
@@ -0,0 +1,65 @@
...
+          callback: () => {
+            validatePassword();
+          }
+        }
+      ]}
+    >
+      <FormControl
+        error={error}
+        onKeyDown={e => {
+          if (e.keyCode === 13) validatePassword();
```
qgolsteyn, line 49: NIT: Make keycode 13 its own constant. Helps with readability as I do not know what this code maps to which key.

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,141 @@
...
+import { type WikiSettingsT } from './index';
+
+type ModalSettingsPropsT = {
+  callback: WikiSettingsT => void,
+  hideModal: () => void,
+  currentSettings: WikiSettingsT,
+  classes: Object
+};
+
+const styles = {
```
qgolsteyn, line 35: NIT: You should check out the `makeStyle` approach, which uses hooks. This gives you access to the Material UI theme object, which contains useful functions to help normalize margin width, font size, colors, etc.
https://material-ui.com/styles/basics/

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,141 @@
...
+  group: {
+    margin: '1vh'
```
qgolsteyn: Prefer using percentage based values over viewport-based values

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,141 @@
...
+    display: 'inline',
+    paddingLeft: '1vw',
+    fontSize: '1rem'
+  },
+  select: {
+    margin: '1vh'
+  }
+};
+
+export default withStyles(styles)(
```
qgolsteyn: NIT: Define the component in its own object and pass it to `withStyles` to be consistent with the rest of the codebase.

frog/imports/client/Wiki/ModalSettings.js
```diff
@@ -0,0 +1,141 @@
...
+                </FormHelperText>
+              </>
+            }
+            labelPlacement="end"
+          />
+          <FormControlLabel
+            checked={!settings.allowPageCreation}
+            control={<Switch color="secondary" />}
+            onChange={e =>
+              setSettings({ ...settings, allowPageCreation: !e.target.checked })
```
qgolsteyn: Here you do not use `handleChange`

frog/imports/client/Wiki/ModalSettings.js
qgolsteyn: I agree with @houshuang. This implementation makes use of currying, which complicates our code and makes our code more fragile by breaking type checking. This is not a common approach that we use in our codebase and should be changed. I recommend using `setSettings` directly in the JSX tree.

frog/imports/client/Wiki/WikiContentComp.js
qgolsteyn, line 155: `checkEdit` is not defined in `WikiContentComp`

frog/imports/client/Wiki/WikiLink.js
```diff
@@ -12,6 +12,7 @@ const WikiLink = observer(
       const { data } = this.props;
       const pageObj = wikiStore.pages[data.id];
       const noFollowLinks = wikiStore.noFollowLinks;
+      const preventPageCreation = wikiStore.preventPageCreation;
```
qgolsteyn, line 15: NIT: you can use destructuring to merge the two lines of code.

frog/imports/client/Wiki/WikiLink.js
```diff
@@ -48,7 +49,7 @@ const WikiLink = observer(
         return (
           <span
             ref={e => (this.ref = e)}
-            onClick={noFollowLinks ? null : createLinkFn}
+            onClick={noFollowLinks || preventPageCreation ? null : createLinkFn}
```
qgolsteyn, line 52: If I understand correctly, we disable link creation entirely if page creation is disabled?

frog/imports/client/Wiki/index.js
```diff
@@ -59,8 +65,12 @@ type WikiCompPropsT = {
   query?: Object
 } & ModalParentPropsT;
 
-type WikiSettingsT = {
-  readOnly: boolean
+export type WikiSettingsT = {
```
qgolsteyn: This will cause a cyclic dependency and should be moved to a `types.js` file

frog/imports/client/Wiki/index.js
```diff
@@ -74,7 +84,8 @@ type WikiCompStateT = {
   noInstance: ?boolean,
   username: string,
   isAnonymous: boolean,
-  settings?: WikiSettingsT
+  settings?: WikiSettingsT,
```
qgolsteyn: We should make settings not optional to help reduce the code complexity.

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+      this.wikiDoc.data.settings?.locked &&
+      !this.wikiDoc.data.owners.find(x => x === Meteor.userId())
+    ) {
+      this.setState({ currentPageObj: null });
+      this.props.showModal(<LockedModal />);
+      return;
+    } else {
+      this.initialLoad = true;
+      this.props.hideModal();
+    }
```
qgolsteyn, line 189: Move this code to its own function to reduce the number of things `loadWikiDoc` does

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+      } else addUser(this.wikiDoc, Meteor.userId());
+      return;
+    } else {
+      if (
+        this.wikiDoc.data.users?.find(x => x === Meteor.userId()) === undefined
+      )
+        addUser(this.wikiDoc, Meteor.userId());
+      this.initialLoad = true;
+      this.props.hideModal();
+    }
```
qgolsteyn, line 224: Move this code to its own function to reduce the number of things loadWikiDoc does

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+      !this.wikiDoc.data.users?.find(x => x === Meteor.userId()) &&
+      !this.wikiDoc.data.owners.find(x => x === Meteor.userId())
+    ) {
+      this.setState({ currentPageObj: null });
+      const passwordPromise = new Promise(resolve => {
+        this.props.showModal(
+          <PasswordModal
+            callback={resolve}
+            hideModal={this.props.hideModal}
+            actualPassword={this.wikiDoc.data.settings?.password}
```
qgolsteyn: What happens if settings is undefined?

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
       );
     }
 
+    // Show locked modal if the wiki is locked
+    if (
+      this.wikiDoc.data.settings?.locked &&
+      !this.wikiDoc.data.owners.find(x => x === Meteor.userId())
```
qgolsteyn: This call should be moved to its own function or saved to a variable as we are repeating ourselves

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+          <PasswordModal
+            callback={resolve}
+            hideModal={this.props.hideModal}
+            actualPassword={this.wikiDoc.data.settings?.password}
+          />
+        );
+      });
+      const result = await passwordPromise;
+      if (!result) {
+        this.props.showModal(
```
qgolsteyn, line 208: We should move this modal to its own file to reduce code complexity and remove the dependency of `index.js` on Modal

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+            ]}
+          >
+            This wiki has been password protected.
+          </Modal>
+        );
+      } else addUser(this.wikiDoc, Meteor.userId());
+      return;
+    } else {
+      if (
+        this.wikiDoc.data.users?.find(x => x === Meteor.userId()) === undefined
```
qgolsteyn: Move this code to its own function to reduce the number of things loadWikiDoc does

frog/imports/client/Wiki/index.js
```diff
@@ -182,12 +194,76 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
       pagesData: wikiStore.pages,
-      settings: this.wikiDoc.data.settings
+      settings: this.wikiDoc.data.settings,
+      isOwner:
+        this.wikiDoc.data.owners.find(x => x === Meteor.userId()) !== undefined
     });
-
+    if (
+      (this.wikiDoc.data.settings?.readOnly ||
+        this.wikiDoc.data.settings.restrict === 'edit') &&
```
qgolsteyn: The allowed restrict values should be moved to constants to help with code maintenance

frog/imports/client/Wiki/index.js
```diff
@@ -520,6 +595,82 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
     return { liId, pageId };
   };
 
+  editAccess = async action => {
+    if (this.state.isOwner) return true;
+    if (action === 'createPage') {
+      if (!this.state.settings?.allowPageCreation) {
+        this.props.showModal(
```
qgolsteyn: This Modal should be moved to its own file. This indicates we might need a standard way to display error messages

frog/imports/client/Wiki/index.js
```diff
@@ -617,13 +768,36 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+          {
+            title: 'Restore deleted page',
+            icon: RestorePage,
+            callback: () => this.openRestorePageModal(invalidPages)
+          },
+          {
+            title: 'Wiki Settings',
+            icon: Tune,
+            callback: () =>
+              this.props.showModal(
```
qgolsteyn: Create a new function to show PermissionModal

frog/imports/client/Wiki/index.js
```diff
@@ -720,7 +898,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
                 goToPage={this.goToPage}
                 dashboardSearch={this.state.dashboardSearch}
                 side={this.state.mode === 'splitview' ? 'left' : null}
-                disableEdit={this.props.embed || this.state.settings?.readOnly}
+                checkEdit={() => this.editAccess('editPage')}
```
qgolsteyn, line 887: `'editPage'` does not appear to be a case checked by an if statement inside `checkEdit`

frog/imports/client/Wiki/index.js
```diff
@@ -617,13 +768,36 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
-        callback: () => this.openRestorePageModal(invalidPages)
-      }
-    ];
+    const secondaryNavItems = this.state.isOwner
+      ? [
+          {
+            title: 'Restore deleted page',
+            icon: RestorePage,
+            callback: () => this.openRestorePageModal(invalidPages)
+          },
```
qgolsteyn: This is not DRY. Make a push to `secondaryNavItems` instead of repeating the definition `'Restore deleted page'`

frog/imports/client/Wiki/index.js
```diff
@@ -520,6 +595,82 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
     return { liId, pageId };
   };
 
+  editAccess = async action => {
```
qgolsteyn, line 608: What does this function do?

qgolsteyn, line 608: Break this function into smaller parts that are responsible for editing access to different features

qgolsteyn (changes requested):
There is a lot of changes in this PR and will require another round of review after these are first addressed.

frog/imports/api/wikiDocHelpers.js
ajainuary: Implemented on the suggestion of @houshuang , we need to show the existing password to the user in case the owner has shared it with some subset of users and forgotten it. In this case, the owner would not want to change the password since it would cause that subset of users with the existing password to be communicated of this change.

frog/imports/api/wikiDocHelpers.js
ajainuary: `'none'`, `'edit'`, `'access'` for different levels of permissions

frog/imports/client/Wiki/index.js
ajainuary: It is implicitly being checked since there are only 2 possibilities `'editPage'` and `'createPage'`. If you insist, I could change the `if` statement to make this explicit

frog/imports/client/Wiki/index.js
ajainuary: I will add a JSDoc comment explaining the function

ajainuary: Why?

frog/imports/client/Wiki/index.js
ajainuary: Ok, fair point

frog/imports/api/wikiDocHelpers.js
ajainuary: What would be the advantage? It's required only in this function

frog/imports/api/wikiDocHelpers.js
ajainuary: To detect if the password was changed, in which case we would need to invalidate existing users who used the old password.

frog/imports/client/Wiki/ModalPassword.js
ajainuary: Please refer to my previous comment explaining why storing plain-text password is a requirement.

frog/imports/client/Wiki/ModalPassword.js
ajainuary: Fair point, maps to the `Enter` key (universally).

frog/imports/client/Wiki/ModalSettings.js
ajainuary: Fair point, I will make the respective changes

frog/imports/client/Wiki/ModalSettings.js
ajainuary: Thank you for your feedback

frog/imports/client/Wiki/ModalSettings.js
ajainuary: At some places in the existing codebase, we do use this kind of style.

frog/imports/client/Wiki/ModalSettings.js
ajainuary: This is because I wish to negate the input.

frog/imports/client/Wiki/WikiLink.js
ajainuary: No, only new pages cannot be clicked.

frog/imports/client/Wiki/index.js
ajainuary: Existing wikis don't have the settings object, this has been done to ensure backward compatibility

frog/imports/client/Wiki/index.js
ajainuary: Why would this be necessary?

frog/imports/client/Wiki/index.js
ajainuary: Then we would never reach this statement, I will remove Maybe conditional.

frog/imports/client/Wiki/index.js
ajainuary: We are checking the existence of the user in 3 different lists, `owners`, `editors` & `users`

frog/imports/client/Wiki/index.js
ajainuary: Why should we create a new file for 10 lines of code?

frog/imports/client/Wiki/index.js
ajainuary: It would **not** reduce the number of things loadWikiDoc does.

frog/imports/client/Wiki/index.js
ajainuary: Yes, we do need a standard way to display error/success messages, but in absence of such a method we would need to resort to modals and creating a file for 10 lines of code seems too much to me.

frog/imports/api/wikiDocHelpers.js
qgolsteyn: Okay, let's make this clear in a code comment

frog/imports/api/wikiDocHelpers.js
qgolsteyn: Perfect, let's make it clear

frog/imports/api/wikiDocHelpers.js
qgolsteyn: Because of testability, readability, extensibility.

frog/imports/client/Wiki/ModalSettings.js
qgolsteyn: Yes, but in recent code, we do not use this approach

frog/imports/client/Wiki/store.js
```diff
@@ -7,11 +7,13 @@ class WikiStore {
     extendObservable(this, {
       pages: {},
       noFollowLinks: false,
-
+      preventPageCreation: false,
```
qgolsteyn, line 10: We used `allowPageCreation` in `ModalSettings`, choose one

frog/imports/client/Wiki/index.js
qgolsteyn: Again, as previously mentioned, provide a default set of values for those pages to handle backward compability. The rest of the code should not be responsible for this behaviour.

frog/imports/client/Wiki/index.js
qgolsteyn: You are right, however, it makes the code DRY

frog/imports/client/Wiki/index.js
qgolsteyn: Functions should do one thing, and only one thing

frog/imports/client/Wiki/index.js
qgolsteyn: Make it a switch statement and move code away to separate functions

PR #1792, opened 2019-07-16
Moving frog-utils and changing ShareDB fork

houshuang:
This is unfortunately a huge PR, but it only does two things:

- moves frog-utils into the main FROG repo, which should improve load times in general, and specifically for stuff which we can put in a /client folder, which will prevent reloading the entire server (I have done this with ReactiveRichText). Unfortunately this leads to a huge number of files changed to update import statements.
- updating a bunch of ShareDB libraries to other forks - this is required for presence/shared cursor functionality that is coming in another PR. These are the same versions that we're using at Minerva.

All functionality should be identical.

You probably need to do git clean -fdx and initial_setup.


RachitMalik12 (dismissed):
LGTM

RachitMalik12 (dismissed):
Changed import statement.. now LGTM 🥇

PR #1798, opened 2019-07-19
Putting back all activities

houshuang:
Can you guys check if reload times are very different? I'll either merge this back, or keep it separate and only merge it before I push to production.


RachitMalik12 (commented):
Reload times are slower for me. I think it's best to not merge into develop..

PR #1800, opened 2019-07-23
Reformat Cypress tests

qgolsteyn:
This PR updates Cypress tests to fit the format described on this document: https://docs.google.com/spreadsheets/d/1-IntypKfr-e_6E_bD-QK5LvfJKXKM5phjzEyMUJabYw/edit#gid=1105876440

### DIFF
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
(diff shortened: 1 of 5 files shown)


houshuang (approved):

houshuang:
This is really great, I love how clean the tests are, the reporting when you run Cypress etc.

My only 2 concerns are:
- the cost of loading FROG for every single item
- how we handle things like fixtures, whether we want to modify the database or sharedb to test specific conditions that take too long to set up using Cypress actions

@ajainuary 	Could you look at this and add more tests for the wiki (settings etc)?

PR #1801, opened 2019-07-23
TopBar element

qgolsteyn:
This PR implements a common TopBar element with the goal to harmonise this UI widget, currently implemented three times in Wiki, Single Activity and Pro FROG. The TopBar is divided in two parts, navigation and actions.

This is how I see the various navigation views for each FROG app:

**Wiki**
- Single page
- Split view

**Single activity**
- Create/Edit activity/template
- Run session

**FROG Pro**
- Edit Graph
- Run session
- Preview activity

----

**To discuss:** 
- UI for user info, with login/sign up/logout action
- Custom views inside the top bar (for example to show the next activity when running a graph)

### DIFF
--- a/frog/imports/ui/TopBar/ActionMenu.js
+++ b/frog/imports/ui/TopBar/ActionMenu.js
@@ -14,7 +14,7 @@ import { MoreVert } from '@material-ui/icons';
 
 import type { TopBarActionT } from './types';
 
-const useStyle = makeStyles(theme => ({
+const useStyle = makeStyles(() => ({
   root: {
     display: 'flex'
   }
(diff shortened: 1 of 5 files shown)


houshuang:
I guess the only way to see this is Storybook? That's fine, but it would have been good to mention, I initially booted up FROG to see if anything had changed :)

houshuang:
This is neat, and I love playing with Storybook (great addition to our dev experience). However it's a bit hard for me to think abstractly about a sidebar across the three types of interfaces, since they are all different and what we put in the top bar depends a lot on how we design the entire interface. But this seems flexible enough that we can modify it depending on how we want to design? 

@ajainuary 	and @RachitMalik12 	any input?

houshuang:
@qgolsteyn 	Should this still be merged in, or has this code already been added through other PRs?

PR #1802, opened 2019-07-23
Users api 

RachitMalik12:
This PR adds a users API to simplify the way we get the username in different parts of FROG. Since we are in the middle of transitioning the accounts system and we are now moving to displayNames, the function `getUsername()` returns the right display name based on the type of the user logged in or passed in as an argument. 

**Testing instructions**
Functionality should be unchanged

**Asana task**
https://app.asana.com/0/1129712605359574/1132275539342866/f


frog/imports/api/remoteActivities.js
```diff
@@ -50,7 +51,7 @@ export const collectActivities = (callback: ?Function) =>
   fetch(
     RemoteServer +
       '?select=uuid,title,description,tags,activity_type,owner_id,timestamp,is_public&deleted=not.is.true&or=(is_public.not.is.false,owner_id.eq.' +
-      Meteor.user().username +
+      getUsername +
```
houshuang: should be getUsername()?

frog/imports/api/users.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Returns the appropriate user object based on the type of user. If no user is passed as args then will return the current user object.
+ *
+ * @param: {User=} user
+ */
+const getUser = (user: User) => {
+  if (!user) return Meteor.user();
+  else {
+    const { id, userObj } = user;
+    return Meteor.users.findOne(id) || userObj;
```
houshuang: This is a minor issue, but if id is undefined, better to avoid a database lookup? Maybe

`if(id) { return Meteor.users.findOne(id) } else if userObj { return userObj } else { return Meteor.user() } `

frog/imports/api/users.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+/**
+ * Returns whether the given user is verified i.e. has an email. If no user is provided then will return whether the current user is verified.
+ *
+ * @param: {User=} user
+ */
+
+export const isVerifiedUser = (user: User) => {
+  if (!user) {
```
houshuang: Why don't you use the getUser function here?

frog/imports/client/App/TopBar.js
```diff
@@ -184,9 +184,7 @@ class TopBarController extends React.Component<
             </Tabs>
             <h3>
               {LocalSettings.researchLogin ? ' * ' : ''}
-              {Meteor.user().isAnonymous
-                ? 'Anonymous Visitor'
-                : Meteor.user().username}
+              {getUsername()}
```
houshuang, line 187: I like this simplification :)

frog/imports/client/Preview/ConfigPanel.jsx
```diff
@@ -130,7 +130,7 @@ class ConfigPanel extends React.Component<*, *> {
       setActivityTypeId
     } = this.props;
     if (
-      metadatas.owner_id === Meteor.user().username &&
+      metadatas.owner_id === getUsername() &&
```
houshuang, line 133: All of this logic is even worse now that multiple users can have the same display name - you don't even need a separate server to impersonate another user :) But that's not your problem, we'll have to rethink this completely at some point, but we'll leave it for now. Anyway you can't access another users sessions and student data, only the publicly stored graph.

frog/imports/client/StudentView/Runner.jsx
```diff
@@ -60,7 +61,7 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
     (isTeacher && activity.plane === 2 ? '' : groupingValue) +
     ')';
   if (activity.plane === 1) {
-    title = `(individual/${Meteor.user().username})`;
+    title = `(individual/${getUsername})`;
```
houshuang: getUsername()?

frog/imports/client/StudentView/Runner.jsx
```diff
@@ -117,7 +118,7 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
           readOnly
         }}
         activityId={activity._id}
-        username={Meteor.user().username}
+        username={getUsername}
```
houshuang: getUsername()?

frog/imports/client/StudentView/SessionBody.jsx
```diff
@@ -82,7 +83,7 @@ const StudentView = withRouter(({ activities, session, classes, history }) => (
               color="inherit"
               className={classes.flex}
             >
-              {Meteor.user().username}
+              {getUsername}
```
houshuang: getUsername()?

houshuang (changes requested):
If you can address these issues, ping me again and I will do a functional test. 

Here's what I would test: 
for the three kinds of accounts (old account created with ?login=name, anonymous account, and verified account):

displays correctly in:
- single activity
- graph editor
- dashboard
- when teacher joins a class (chat for example, projector view)
- when student joins a class
- when data gets merged (for example from ac-brainstorm to ac-ck-board)
- in the wiki (top bar)
- in the wiki (instance list for individual pages)
- when creating/editing graphs/activities from the server
- when exporting session data

run_cypress.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
houshuang: Why do you introduce this in this PR? ALso why use such an old version of Cypress?

frog/imports/api/users.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param: {User=} user
+ */
+
+export const getUsername = (user: User) => {
+  const selectedUser = getUser(user);
+  if (selectedUser) {
+    if (selectedUser.isAnonymous) return 'Anonymous User';
+    else if (selectedUser.emails) return selectedUser.profile.displayName;
+    else if (selectedUser.username) return selectedUser.username;
+  }
```
qgolsteyn, line 21: NIT: else statement

run_cypress.sh
```diff
@@ -0,0 +1,18 @@
...
+    if [ -f "$FROG/node_modules/.bin/cypress" ] && [[ `"$FROG/node_modules/.bin/cypress" --version` =~ $CYPRESS_VERSION ]]; then 
+        echo 'Using pre-installed local Cypress'; CYPRESS="$FROG/node_modules/.bin/cypress"
+    else
+        echo 'Installing Cypress'; npm install cypress@$CYPRESS_VERSION && CYPRESS="$FROG/node_modules/.bin/cypress"
+    fi
+fi
+echo "Cypress: $CYPRESS"
+
+"$CYPRESS" run --record --key e92a866f-0cde-45be-9cd8-72f9ed6650f3
+exit 0
```
qgolsteyn: This file doesn't seem related to this PR...

frog/imports/api/users.js
```diff
@@ -0,0 +1,50 @@
...
+  return 'No user logged in';
+};
+
+/**
+ * Returns the appropriate user object based on the type of user. If no user is passed as args then will return the current user object.
+ *
+ * @param: {User=} user
+ */
+const getUser = (user: User) => {
```
qgolsteyn: I don't really understand this function:
- in the case the user has an id: we return a user that has the id. Couldn't we simply have a function called findUserById? In which scenario would we have an id but no userObj?
- in the case the user has an obj but no id: We already have the obj, why do we need this function?
- in the case we have nothing: we return the current user. Would suggest creating a function called getCurrentUser

qgolsteyn (changes requested):
Left some comments, I have some confusion around the `getUser` function

frog/imports/api/users.js
houshuang: This is a helper function for the main function (note that it is not exported), which makes it easier to call the main function in three different ways (with current user, with id, with object). I asked for this, and I think it's OK.

houshuang (changes requested):

frog/imports/api/users.js
RachitMalik12: The idea behind this API was to  get the correct display name for each type of user by either passing in a user object or a userid and thus reduce the number of checks we have and also the number of calls to Meteor.user().username. 
getUser is not being exported but just as a helper to getUsername. In different parts of the FROG codebase we want to get the username by either passing the user id or user object or nothing.

run_cypress.sh
RachitMalik12: no idea how it got here 🤷🏻‍♂️

PR #1803, opened 2019-07-24
Page Settings

ajainuary:
**Description**
This PR introduces Page-wise Settings: 
1. It makes the checkboxes in the New Page Modal functional
2. Allows users to change page settings after creating a page
3. Some settings are conditional on the plane of the page.

**How to test?**
Create a new page and choose suitable settings, alternatively change settings after creating a new page.


frog/imports/api/wikiDocHelpers.js
```diff
@@ -9,6 +12,7 @@ export const addNewWikiPage = (
   liType = 'li-richText',
   liId,
   plane,
+  pageSettings,
```
qgolsteyn: What type is this??

frog/imports/api/wikiDocHelpers.js
```diff
@@ -22,6 +26,7 @@ export const addNewWikiPage = (
     liId,
     liType,
     plane,
+    pageSettings,
     instances,
     noNewInstances
   };
```
qgolsteyn, line 29: Remove if statement below this comment

frog/imports/api/wikiDocHelpers.js
```diff
@@ -197,6 +202,25 @@ export const updateSettings = (wikiDoc: Object, settings: WikiSettingsT) => {
   wikiDoc.submitOp(op);
 };
 
+export const updatePageSettings = (
+  wikiDoc: Object,
+  pageId: string,
+  socialPlane: number,
+  pageSettings: PageSettingsT
```
qgolsteyn, line 206: <3

frog/imports/api/wikiDocHelpers.js
```diff
@@ -1,6 +1,9 @@
 // @flow
 import { uuid } from '/imports/frog-utils';
-import { type WikiSettingsT } from '/imports/client/Wiki/types';
+import {
+  type WikiSettingsT,
+  type PageSettingsT
+} from '/imports/client/Wiki/types';
```
qgolsteyn: Consider moving types to the api folder...

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -34,19 +35,23 @@ type StateT = {
...
-  operatorConfig?: Object
+  operatorConfig?: Object,
+  error: ?string
 };
 
 type PropsT = {
   classes: Object,
-  onCreate: Function,
-  setModalOpen: Function,
+  onSubmit: Function,
```
qgolsteyn: Define that function

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -34,19 +35,23 @@ type StateT = {
...
+  hideModal: Function,
```
qgolsteyn: Define that function

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -34,19 +35,23 @@ type StateT = {
   socialPlane: number,
   open: boolean,
   expanded: boolean,
-  allowView: boolean,
-  allowEdit: boolean,
+  pageSettings: PageSettingsT,
   activityConfig?: Object,
-  operatorConfig?: Object
+  operatorConfig?: Object,
```
qgolsteyn, line 44: Use type mixed if we do not know the structure of this object

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -26,6 +26,7 @@ import IconButton from '@material-ui/core/IconButton';
 import ApiForm from '../GraphEditor/SidePanel/ApiForm';
```
qgolsteyn, line 27: Rename component to PageSettingsModal

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -34,19 +35,23 @@ type StateT = {
...
+  error: ?string
 };
 
 type PropsT = {
   classes: Object,
-  onCreate: Function,
-  setModalOpen: Function,
+  onSubmit: Function,
+  hideModal: Function,
   clearError: Function,
```
qgolsteyn: Remove this

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -34,19 +35,23 @@ type StateT = {
...
-  errorDiv: any,
-  wikiId: string
+  isOwner: boolean,
+  wikiId: string,
+  pageSettings?: PageSettingsT,
```
qgolsteyn, line 60: Comment comment comment

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -81,72 +86,79 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
-      pageTitle: '',
+      pageTitle: this.props.title || '',
       pageTitleValid: true,
       open: true,
       expanded: false,
-      socialPlane: 3,
-      allowView: true,
-      allowEdit: true
+      socialPlane: this.props.socialPlane || 3,
+      pageSettings: this.props.pageSettings || {
```
qgolsteyn: Should move to a constant called `DEFAULT_SETTINGS`

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -81,72 +86,79 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
       pageTitle,
       socialPlane,
       activityConfig,
-      operatorConfig
+      operatorConfig,
+      pageSettings
     } = this.state;
-    this.props.onCreate(pageTitle, socialPlane, activityConfig, operatorConfig);
+    this.props
+      .onSubmit(
```
qgolsteyn, line 151: onSubmit does not have a type definition specific enough

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -81,72 +86,79 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
+        pageTitle,
+        socialPlane,
+        activityConfig,
+        operatorConfig,
+        pageSettings
+      )
+      .then(error => {
```
qgolsteyn: Do we expect props.onSubmit to return a Promise? If so, specify that in the type. If the parent does not respect this invariant, this will lead to an exception

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -158,19 +170,10 @@ class NewPageModal extends React.Component<PropsT, StateT> {
         operatorTypesList.push('op-aggregate');
       }
     }
-
     return (
       <Dialog
```
qgolsteyn, line 179: We should use the Modal object and the Modal Controller

frog/imports/client/Wiki/ModalPageSettings.js
qgolsteyn: Using Modal, we do not need passing this function

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -201,64 +208,101 @@ class NewPageModal extends React.Component<PropsT, StateT> {
                 }}
                 data-testid="wiki_page_title_editor"
               />
-              {errorDiv !== null && (
-                <FormHelperText error>{errorDiv}</FormHelperText>
+              {this.state.error !== null && (
```
qgolsteyn: You already destructured `state`

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -201,64 +208,101 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
-              </Tabs>
-            </AppBar>
+          <Collapse
+            in={expanded || this.props.action === 'edit'}
+            timeout="auto"
+          >
+            {this.props.action === 'create' && (
+              <AppBar position="static">
+                <Tabs
+                  value={this.state.currentTab}
```
qgolsteyn, line 222: You use both destructuring and access the state attributes directly, choose one

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -201,64 +208,101 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
-                  </FormControl>
-                  <FormGroup row>
-                    <FormControlLabel
-                      control={
-                        <Checkbox
-                          checked={this.state.allowView}
-                          onChange={this.handleChangeAllowView}
-                          color="primary"
+                <>
+                  <FormGroup>
```
qgolsteyn, line 235: We should consider using react-jsonschema-form as this view logic is quite complex

frog/imports/client/Wiki/ModalPageSettings.js
```diff
@@ -311,22 +355,16 @@ class NewPageModal extends React.Component<PropsT, StateT> {
...
-                this.props.clearError();
-              }}
-              color="primary"
-            >
+            <Button onClick={this.props.hideModal} color="primary">
               Cancel
             </Button>
             <Button
-              onClick={() => this.handleCreate()}
+              onClick={() => this.handleSubmit()}
```
qgolsteyn, line 363:
```suggestion
              onClick={this.handleSubmit}
```

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -61,6 +54,13 @@ class WikiContentComp extends React.Component<> {
         error: null
       });
     }
+    if (this.state.docMode === 'edit') {
+      this.setState({ docMode: 'view' }, () => {
+        this.props.checkEdit().then(x => {
```
qgolsteyn: What is x, specify x

frog/imports/client/Wiki/WikiContentComp.js
```diff
@@ -61,6 +54,13 @@ class WikiContentComp extends React.Component<> {
         error: null
       });
     }
+    if (this.state.docMode === 'edit') {
```
qgolsteyn, line 68: Simplify, or comment comment comment

frog/imports/client/Wiki/index.js
```diff
@@ -303,22 +302,58 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
     }
 
-    const instanceId =
+    if (
+      fullPageObj.pageSettings !== undefined &&
+      fullPageObj.pageSettings.hidden &&
+      this.state.privilege !== PRIVILEGE_OWNER
+    ) {
+      this.props.showModal(
+        <AlertModal
```
qgolsteyn, line 334: Use the standard AlertModal

frog/imports/client/Wiki/index.js
```diff
@@ -303,22 +302,58 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+          title="Unable to view Page"
+          callback={() => {
+            this.props.hideModal();
+          }}
+        >
+          This page is hidden by the owner.
+        </AlertModal>
+      );
+      fullPageObj = parsedPages['home'];
```
qgolsteyn, line 343: What happens if home is deleted?

frog/imports/client/Wiki/index.js
```diff
@@ -475,11 +510,46 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
...
+    }
     const instanceId = foreignInstanceId || this.getInstanceId(fullPageObj);
     const newCurrentPageObj = this.getProperCurrentPageObj(
       fullPageObj,
       instanceId
     );
+    if (
+      fullPageObj.pageSettings !== undefined &&
+      !fullPageObj.pageSettings.allowView &&
+      instanceId !== Meteor.userId() &&
```
qgolsteyn, line 552: User ID can be called once when the component is created

qgolsteyn (changes requested):
Some logic should be refactored to reduce code duplication, improve type safety, reduce the number of callback nesting, remove asynchronous logic inside UI components, reduce the number of parameters passed around in each function.

Some ideas:
- Use a context object, this removes the need to pass props to each modal, making the code more maintainable. Contacting the server could be done here.
- Keep parameters inside their parent objects. With that many fields, destructuring makes the code harder to read.
- Group error modals (can't read, can't write, can't see instance) into one modal, that conditionally shows error based on pageObj.
- Could we use MobX?

RachitMalik12 (commented):
Can you fix eslint?

RachitMalik12 (changes requested):
Couple of things functionality wise: 

- Create a new wiki and go to the page settings menu. Delete the page title in the modal and click Apply. Page disappears from the sidebar. Since there is no error validation I can name it % and it crashes FROG. 
- The arrow at the bottom of the page settings modal is a bit confusing since it doesn't do anything. Maybe remove it? 
- If I disable allowing others to view in the Individual plane and then change the social plane to Everyone, the check boxes show Allow others to view checked. This makes me think that option is enabled but in reality it is not. 
- If I create a page, allow it to be read only and then another user navigates to the same page they can see it. But then if the wikiOwner makes the page hidden, the page disappears from the sidebar but I can still see the content. On further discussion noted that the changes are not instant and take into effect after reload but this could be something that could worked on in the future. 
- If I create a page say Page1 and make it hidden and another user tries to go this page by URL , after the alert that says Page is hidden pops up I click ok and then I get a blank screen. I think we should reload the wiki and go back to home in that case. 
- If the wikiOwner creates a page that is read only and then another user goes to that page and then tries to create a new page, it blocks creating a page by saying cannot edit the page. Create a page shouldn't be blocked in this case. 
Other than that looks fine :). Everyone else feel free to test further..

frog/imports/client/Wiki/index.js
ajainuary: If users are able to delete home then we have a different problem.

frog/imports/client/Wiki/index.js
ajainuary: Please elaborate on the standards.

frog/imports/client/Wiki/ModalPageSettings.js
ajainuary: In this case, we need to have custom styling which would not be possible by the Modal Object

PR #1804, opened 2019-07-24
Toast notifications on sign up/login 

RachitMalik12:
This PR adds toast notifications on success/error while signing up or logging in to FROG. 

**Testing instructions**
Create a new account. If it is created successfully then you will see a success toast. If there are errors an error toast will be displayed which should disappear on closing the modal. Same for logging in 

**Asana task**
https://app.asana.com/0/1129712605359574/1132009327400800/f


houshuang (changes requested):
I rewrote the two modals to functions with hooks, because it seemed cleaner. Can you double-check that it's still working as planned? 

I prefer having the error "Incorrect password" show up in the login box, rather than as a toast.

Same with "Please fill out all fields". I think toasts should only be used after the modal has closed? (This is the way it works with Create new account modal)

Also, could the toasts be top-right instead of bottom-right? I think that's more natural?

houshuang (changes requested):
I upped the auto-dismiss time to 5 seconds. However, there's an issue with toasts never dismissing if you close the modal.

houshuang (changes requested):
Success toast after logging in to the wiki still never disappears.

RachitMalik12:
> Success toast after logging in to the wiki still never disappears.

I thought it might be a good idea to let just the success toast be there as a notification to the user of successful account creation. Perhaps I think it is better to reduce the duration to about 3000 ms.

houshuang (approved):

PR #1805, opened 2019-07-26
Added sign up and login modal to the wiki 

RachitMalik12:
This PR adds the ability for the user to login or sign up from the wiki.

**Testing instructions**

- Create a wiki page and click on the gear on the right. A drop down with an option to create an account or login or logout will be displayed based on the user logged in. 
- Creating an account and logging in should have expected behaviour 
- The rest of the wiki functionality should be unchanged


houshuang (changes requested):
I pushed a commit to resolve merge conflict.

- the pulldown menu in the wiki doesn't update after you've logged in (if you were anonymous, and then log in as a wiki owner, the wiki settings menu should appear, but it doesn't without a browser reload).
- change upgrade to "Create account"
- open the create account modal when clicking on Anonymous User (only anonymous, not when someone is logged in).

houshuang (approved):

PR #1810, opened 2019-07-30
Blocks api login for verified users and fixes small issue in users api

RachitMalik12:
This PR blocks the ability for users to login through url by typing `?login=username` for verified accounts.

**Testing instructions**

- Create a verified account with the sign up modal. Look at the account in the database, grab the username field and try to login with ?login=username. A message saying could not login will be displayed 
- Try to login with an old account and it should still work 

**Asana task**
https://app.asana.com/0/1129712605359574/1132009327400810/f

### DIFF
--- a/frog/imports/api/users.js
+++ b/frog/imports/api/users.js
@@ -52,8 +52,6 @@ const getUser = (user: User): ?userObj => {
 
 export const isVerifiedUser = (user: User) => {
   const selectedUser = getUser(user);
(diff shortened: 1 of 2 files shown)


houshuang (changes requested):
This makes anonymous login to landing page or wiki impossible (it tries sending username.login null null null wiki).

qgolsteyn (approved):
LGTM :+1:

PR #1813, opened 2019-07-31
Upgrading legacy accounts + Wiki compatibility with legacy accounts 

RachitMalik12:
This PR adds functionality to upgrade legacy accounts to have a password and display name. Also fixes the logout not being displayed when a legacy account logs into the wiki. 

**Testing instructions**

- Login to a legacy account by using ?login=someusername. Note the user id

- Click upgrade your account on the single activity TopBar or the wiki topbar. Enter some display name, email, password. Check the database and the same user id should have the email associated with it. 
- Try to login using the email you entered 

**Asana task**
https://app.asana.com/0/1121331595459324/1133484398049560/f
https://app.asana.com/0/1129712605359574/1133484398049561/f

### DIFF
--- a/frog/imports/api/users.js
+++ b/frog/imports/api/users.js
@@ -36,6 +36,7 @@ export const getUsername = (user: User, wiki: boolean = false): ?string => {
  */
 export const userType = (user: User): string => {
   const selectedUser = getUser(user);
+  // if there is no user logged in then we treat it the same as anonymous
   if (!selectedUser) return 'Anonymous';
   else if (selectedUser.isAnonymous) return 'Anonymous';
   else if (isVerifiedUser({ selectedUser })) return 'Verified';
--- a/frog/imports/api/users.js
+++ b/frog/imports/api/users.js
@@ -49,7 +49,6 @@ export const getUserType = (user?: UserObj): UserType => {
   const selectedUser = getUser(user);
   // if there is no user logged in then we treat it the same as anonymous
   if (!selectedUser || selectedUser.isAnonymous) return 'Anonymous';
-  // eslint-disable-next-line no-extra-boolean-cast
   else if (isVerifiedUser({ meteorUser: selectedUser })) return 'Verified';
   else if (selectedUser.username) return 'Legacy';
   else return 'No user logged in';
--- a/frog/imports/api/users.js
+++ b/frog/imports/api/users.js
@@ -30,7 +30,17 @@ export const getUsername = (user: User, wiki: boolean = false): ?string => {
   }
   return undefined;
 };
-
+/**
+ * Returns the type of the given user if the user is passed or the current user
+ * @param: {User=} user
+ */
+export const userType = (user: User): string => {
+  const selectedUser = getUser(user);
+  if (!selectedUser) return 'Anonymous';
+  else if (selectedUser.isAnonymous) return 'Anonymous';
+  else if (isVerifiedUser({ selectedUser })) return 'Verified';
+  else if (selectedUser.username) return 'Legacy';
+};
 /**
  * Returns the appropriate user object based on the type of user. If no user is passed as args then will return the current user object.
  *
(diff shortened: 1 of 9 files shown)


houshuang (changes requested):
Wording: Add email and password to your account. "Upgrade your account" (instead of create etc). 

Making it work in wiki.

frog/imports/api/users.js
```diff
@@ -30,7 +30,18 @@ export const getUsername = (user: User, wiki: boolean = false): ?string => {
...
-
+/**
+ * Returns the type of the given user if the user is passed or the current user
+ * @param: {User=} user
+ */
+export const userType = (user: User): string => {
+  const selectedUser = getUser(user);
+  // if there is no user logged in then we treat it the same as anonymous
+  if (!selectedUser) return 'Anonymous';
+  else if (selectedUser.isAnonymous) return 'Anonymous';
```
qgolsteyn: The two if statements can be merged using an OR operator

frog/imports/api/users.js
```diff
@@ -30,7 +30,18 @@ export const getUsername = (user: User, wiki: boolean = false): ?string => {
...
+  else if (isVerifiedUser({ selectedUser })) return 'Verified';
```
qgolsteyn: BUG: `selectedUser` is not a valid property of type `User`. Why not pass selectedUser directly?

frog/imports/api/users.js
```diff
@@ -30,7 +30,18 @@ export const getUsername = (user: User, wiki: boolean = false): ?string => {
   }
   return undefined;
 };
-
+/**
+ * Returns the type of the given user if the user is passed or the current user
+ * @param: {User=} user
+ */
+export const userType = (user: User): string => {
```
qgolsteyn:
```suggestion
export const getUserType = (user: User): string => {
```

frog/server/accountManagement.js
```diff
@@ -43,10 +44,10 @@ export const createAccount = (
...
     } else if (!Accounts.findUserByEmail(email)) {
       const user = Meteor.user();
-
-      if (user?.isAnonymous) {
-        // checks for duplicate email and displays error on the console.
-
+      // if the user is anonymous or a legacy user we want to simply add the email and password to their
+      // account. This allows them to keep the graphs and activities they created when they didn't have an
+      // account. Else we just create a new account
+      if (user?.isAnonymous || (user.username && !isVerifiedUser())) {
```
qgolsteyn: We should check the user we have

```suggestion
      if (userType(user) === 'Anonymous' || userType(user) === 'Legacy') {
```

qgolsteyn (changes requested):
Left some comments

houshuang (changes requested):
An upgraded account is still able to log in using ?login=. It looks like the problem is in api/users isVerifiedUser, it uses getUser, but although it passes a user object, it gets back an anonymous account (I used console log to see what goes in and what comes out). There might be some confusion with the id/obj parameters.

RachitMalik12:
> An upgraded account is still able to log in using ?login=. It looks like the problem is in api/users isVerifiedUser, it uses getUser, but although it passes a user object, it gets back an anonymous account (I used console log to see what goes in and what comes out). There might be some confusion with the id/obj parameters.

Yes the parameters were passed the wrong way. On further thought since we have the getUserType function I've decided to keep isVerifiedUser() as a helper. Replaced every instance of isVerifiedUser with getUserType() === 'Verified'.

frog/imports/client/Wiki/index.js
```diff
@@ -1035,8 +1035,7 @@ class WikiComp extends React.Component<WikiCompPropsT, WikiCompStateT> {
         }
       });
     }
-
-    if (Meteor.user().isAnonymous || !isVerifiedUser() || !Meteor.user()) {
+    if (getUserType() === 'Anonymous') {
```
houshuang, line 1038: I like this change :)

houshuang (approved):

PR #1818, opened 2019-08-06
Change password form 

RachitMalik12:
**Note 1 :** Please review and merge #1813 first
 **Note 2 :** I have merged the content of #1813 since I needed some functionality from it. Hence the large number of files changed. 

**Description**

This PR introduces a modal and functionality for a user to change their password. In #1821 a common modal is introduced in which this form will be integrated. 

**Testing instructions**
- From the TopBar of single activity or the overflow menu from the wiki click change password. 
- This should open a modal. Try invalid inputs like: Empty old password, incorrect old password, new password less than 5 characters, new password not matching each other. All this should be handled. Once you change the password successfully, logout and try logging in with the new password.


houshuang:
Should say "Change password" rather than reset (reset is when you've forgotten). Why do we make them repeat the password twice, when they only need to enter it once when they sign up? :)

RachitMalik12:
Closing. Please look at #1821

PR #1821, opened 2019-08-08
Change display name and personal profile modal for FROG users 

RachitMalik12:
**Please look at this PR after #1813  and #1818 in that order. The large number of files changed are due to merging the work of the above 2 PRs since some of the functionality was required to implement this PR**

This PR introduces a user profile modal and functionality to change the users display name. In addition another function to get the email of the verified user was introduced in the users API. 

**Testing instructions**

- From the top bar on single activity or wiki click the overflow menu and select the View/Edit profile option. This brings up a modal and with user info and a pen icon to edit. 
- Change the display name to nothing. Error handling should be taken care of. No success toast is displayed since the user sees the name change on the profile information modal


houshuang (changes requested):
I don't think you should be able to change your email, but right now there is a pen icon next to it (although it doesn't work). Just removing the icon would do it.

RachitMalik12:
> I don't think you should be able to change your email, but right now there is a pen icon next to it (although it doesn't work). Just removing the icon would do it.

Changed that to an inbox icon instead..

houshuang (approved):

PR #1824, opened 2019-08-13
Wiring up orchestration view

houshuang:
This is very far from ready, but I think it will be useful for you to have a look, especially on how I'm wiring up the Storybook components. Going forwards, I'm for example thinking about whether each view should be pulling in the TopBar, or whether that should be done centrally, etc?

A lot of mucking around to try to fix things, not done, will try to clean up and maybe it will be separate PRs etc in the end... 

Given how convoluted a lot of the central logic is, and our desire to not rewrite more than necessary, some of the wiring code will be quite ugly. I think my main concern is that we keep the separation that makes it easy for you three to continue iterating on the design, moving components around etc, even after I have begun wiring up things, and we get a better sense of how they work in practice. Happy to discuss tomorrow.

### DIFF
--- a/frog/imports/client/StudentView/Runner.jsx
+++ b/frog/imports/client/StudentView/Runner.jsx
@@ -61,7 +61,7 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
     (isTeacher && activity.plane === 2 ? '' : groupingValue) +
     ')';
   if (activity.plane === 1) {
-    title = `(individual/${getUsername()||''})`;
+    title = `(individual/${getUsername() || ''})`;
   }
 
   const config = activity.data;
@@ -256,7 +256,6 @@ export class RunActivity extends React.Component<PropsT, {}> {
 }
 
 export default withTracker(({ activity }) => {
-  console.log(activity);
   const object = Objects.findOne(activity._id);
   return { object, activity };
 })(Runner);
--- a/frog/imports/client/StudentView/Runner.jsx
+++ b/frog/imports/client/StudentView/Runner.jsx
@@ -61,30 +61,34 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
     (isTeacher && activity.plane === 2 ? '' : groupingValue) +
     ')';
   if (activity.plane === 1) {
-    title = `(individual/${getUsername()})`;
+    title = `(individual/${getUsername()||''})`;
   }
 
   const config = activity.data;
 
   // if teacher is previewing p1/p2 activity, grab data from first instance
   const activityStructure = getStructure(activity);
+  let activityData;
   if (
     groupingValue !== 'all' &&
     isTeacher &&
     object.globalStructure.studentIds.length === 0
   ) {
-    return <h1>Cannot preview activity when no students are in session</h1>;
+    activityData = {
+      data: activityTypesObj[activity.activityType].dataStructure || {},
+      config: activity.data
+    };
+  } else {
+    activityData = getMergedExtractedUnit(
+      config,
+      object.activityData,
+      activityStructure,
+      groupingValue !== 'all' && isTeacher
+        ? Object.keys(object.activityData.payload)[0]
+        : groupingValue,
+      object.socialStructure
+    );
   }
-
-  const activityData = getMergedExtractedUnit(
-    config,
-    object.activityData,
-    activityStructure,
-    groupingValue !== 'all' && isTeacher
-      ? Object.keys(object.activityData.payload)[0]
-      : groupingValue,
-    object.socialStructure
-  );
 
   const stream = value => {
     Meteor.call('stream', activity, groupingValue, value);
@@ -252,6 +256,7 @@ export class RunActivity extends React.Component<PropsT, {}> {
 }
 
 export default withTracker(({ activity }) => {
+  console.log(activity);
   const object = Objects.findOne(activity._id);
   return { object, activity };
 })(Runner);
--- a/frog/imports/client/StudentView/SessionBody.jsx
+++ b/frog/imports/client/StudentView/SessionBody.jsx
@@ -51,7 +51,10 @@ const styles = {
   }
 };
 
-const ActivityContainer = ({ activities, sessionId }) => {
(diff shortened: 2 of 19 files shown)


frog/imports/client/StudentView/Runner.jsx
```diff
@@ -61,31 +61,35 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
     (isTeacher && activity.plane === 2 ? '' : groupingValue) +
     ')';
   if (activity.plane === 1) {
-    title = `(individual/${getUsername()})`;
+    title = `(individual/${getUsername() || ''})`;
```
qgolsteyn, line 64: A comment here would be useful, I am not sure what this change does

frog/imports/client/StudentView/Runner.jsx
```diff
@@ -61,31 +61,35 @@ const Runner = ({ path, activity, sessionId, object, single }) => {
...
   if (
     groupingValue !== 'all' &&
     isTeacher &&
     object.globalStructure.studentIds.length === 0
   ) {
-    return <h1>Cannot preview activity when no students are in session</h1>;
+    activityData = {
+      data: activityTypesObj[activity.activityType].dataStructure || {},
+      config: activity.data
+    };
```
qgolsteyn, line 80: What does this code do?

frog/imports/client/StudentView/SessionBody.jsx
```diff
@@ -51,7 +51,10 @@ const styles = {
   }
 };
 
-const ActivityContainer = ({ activities, sessionId }) => {
+export const ActivityContainer = ({ activities, sessionId }) => {
```
qgolsteyn, line 54: If exporting ActivityContainer, do we still need StudentView?

frog/imports/client/StudentView/SessionBody.jsx
```diff
@@ -166,7 +169,7 @@ class SessionBodyController extends React.Component<
   }
 }
 
-const SessionBody = compose(
+export const SessionBody = compose(
```
qgolsteyn, line 172: Just in case, this PR does not make use of the exported SessionBody. This could be a typo?

frog/imports/client/TeacherView/Components/OrchestrationContainer.js
```diff
@@ -0,0 +1,139 @@
...
+import { Logo } from '/imports/ui/Logo';
+import { OverflowMenu } from '/imports/ui/OverflowMenu';
+import { ActivityStatus } from '/imports/ui/ActivityStatus';
+import StepsContainer from './StepsContainer';
+import { nextActivity } from '/imports/api/engine';
+import { ActivityContainer } from '/imports/client/StudentView/SessionBody';
+import { Activities } from '/imports/api/activities';
+import { restartSession } from '/imports/api/sessions';
+
+const SimpleWrapper = ({ activities, session }) => {
```
qgolsteyn, line 26: Consider renaming to OrchestrationContainer

frog/imports/client/TeacherView/Components/OrchestrationContainer.js
```diff
@@ -0,0 +1,139 @@
...
+              <Button
+                icon={<SupervisedUserCircle fontSize="small" />}
+                rightIcon={<ArrowDropDown fontSize="small" />}
+              >
+                Filter by group
+              </Button>
+            }
+          />
+          {ready && (
+            <div style={{ height: 'calc(100% - 100px)' }}>
```
qgolsteyn, line 126: We can wrap this screen's content inside a div with `display: flex` to avoid the use of `calc(...)`.

frog/imports/client/TeacherView/Components/OrchestrationContainer.js
qgolsteyn, line 26: As we work on this component, we should try to move most of the api logic to a mobx store and split this component into smaller containers

frog/imports/client/TeacherView/Components/Steps.js
```diff
@@ -0,0 +1,63 @@
...
+          if (onClick) {
+            onClick(step._id);
+          }
+        }}
+      />
+    ))}
+  </Panel>
+);
+
+// const StepsExample = () => (
```
qgolsteyn, line 53: Consider moving the example code to a Storybook story

frog/imports/client/TeacherView/Components/StepsContainer.js
```diff
@@ -0,0 +1,48 @@
...
+    : timeInGraph > 1
+    ? 999
+    : 0;
+  const maxSeq = Object.keys(sequence).reduce(
+    (acc, x) => (sequence[x] > acc ? sequence[x] : acc),
+    0
+  );
+
+  const steps = [];
+  for (let i = 0; i < maxSeq + 1; i++) {
```
qgolsteyn, line 24: NIT: Consider replacing `i < maxSeq + 1` with `i <= maxSeq`

frog/imports/client/TeacherView/Components/StepsContainer.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+import * as React from 'react';
+import { isEmpty } from 'lodash';
+
+import { Activities } from '/imports/api/activities';
+import { getActivitySequence } from '/imports/api/graphSequence';
+import Steps from './Steps';
+
+const StepsContainer = ({ activities, openActivities, timeInGraph }) => {
```
qgolsteyn, line 9: I am a bit concerned with doing repetitive work as this logic will run every time the container rerenders. We could make use of `React.useMemo`.

qgolsteyn (changes requested):
Left a couple of comments.

Generally, I think we should take the opportunity to:
- Split the UI into small containers which rely on a single source of truth (a mobx store?)
- this store can specify the shape of the view model and be responsible for performing the data manipulation to convert the server model to the view model
- this should help with limiting the number api functions used in React components, and allow us to iterate on the UI with more confidence

Of course, I understand that this might take more work, so this is something we can do later if we do not have the time right now

PR #1828, opened 2019-08-15
Feature/single activity ui

qgolsteyn:
This PR updates single activity to make use of the new FROG UI components. It groups all data manipulation logic for SingleActivity inside a single folder to reduce the amount of work done inside React components. It also updates Storybook stories to improve navigation (UI components under the UI "folder", App components inside the App "folder).

Feedback, especially from @AbhinavThukral97 would be appreciated. The UI has room to improve, but this work should be done once we agree on the FROG branding and after having created other critical UI components (especially typography related components)


frog/package.json
```diff
@@ -56,8 +56,8 @@
     "mathjs": "^5.4.2",
     "meteor-node-stubs": "^0.4.1",
     "minio": "^7.0.5",
-    "mobx": "5.9.0",
-    "mobx-react": "^5.2.3",
+    "mobx": "5.13.0",
```
houshuang, line 59: Was this needed to get it to work?

houshuang (dismissed):

frog/package.json
qgolsteyn: Yup, was needed to support hooks

qgolsteyn:
@houshuang Ready for review:
- Large images no longer break the UI
- Set a max width on the UI
- Fixed some errors React was complaining about

houshuang (approved):
Great work!

PR #1832, opened 2019-08-20
Improving Graph Editor's Design

AbhinavThukral97:
This PR makes minor colour and design changes to improve the look of the Graph Editor


houshuang:
Quite nice. The connection lines are still off, but I can fiddle with that. My one concern is that the level lines (which are quite important) are almost invisible... Either making them more visible, or perhaps just having three "swim lanes" with different colors could be an idea. We should also probably add an icon for the three different levels (many of the previous design proposals had this). These icons can be used in many places. We have some for preview right now, but I'm not sure they're the best... One head for single, two heads for group but how to do class (right now it's an icon that looks like a school building)...

I also would like to shrink the space above and beyond, but we can see how much when we put it on the bottom and try putting the sidepanel on top.

houshuang:
The one advantage of having a lot of space above/beyond is that there is more space to put the operators to get them out of the way :)

Also another new feature is that we'd like a small area to the left for "pre-class" - any operators placed there would be executed immediately when you start the graph (so that you can use for example your group modal to edit student groups before they start etc)...

houshuang (approved):

houshuang:
Great start! Thanks!

PR #1834, opened 2019-08-21
Landing page design (Icons, Video Tutorial and Login Modal Design)

AbhinavThukral97:
This PR adds relevant icons replacing the placeholder images, video tutorial, and colour scheme changes in the Login Modal.


houshuang:
Looks good to me. @RachitMalik12 can you review the changes to the login modal?

frog/imports/client/AccountModal/Login.js
```diff
@@ -20,17 +22,28 @@ const styles = (theme: Object) => ({
...
+    backgroundColor: primaryColor
   },
   form: {
     width: '100%', // Fix IE 11 issue.
     marginTop: theme.spacing(1)
   },
   submit: {
-    margin: theme.spacing(2, 0, 2, 0)
+    background: primaryColor,
+    margin: theme.spacing(2, 0, 2, 0),
```
qgolsteyn: NIT: `theme.spacing(2, 0)`

frog/imports/client/AccountModal/Login.js
```diff
@@ -20,17 +22,28 @@ const styles = (theme: Object) => ({
...
   avatar: {
     margin: theme.spacing(1),
-    backgroundColor: theme.palette.secondary.main
+    backgroundColor: primaryColor
   },
   form: {
     width: '100%', // Fix IE 11 issue.
     marginTop: theme.spacing(1)
   },
   submit: {
```
qgolsteyn, line 31: I suggest using the FROG button instead of defining an additional style

frog/imports/client/AccountModal/Login.js
```diff
@@ -20,17 +22,28 @@ const styles = (theme: Object) => ({
...
-    margin: theme.spacing(2, 0, 2, 0)
+    background: primaryColor,
+    margin: theme.spacing(2, 0, 2, 0),
+    padding: theme.spacing(2),
+    boxShadow: '0 5px 20px rgba(0,0,0,0.05)',
+    transition: '.5s ease',
```
qgolsteyn: Generally, we should aim for shorter transitions to make the interface feel more responsive

frog/imports/ui/LandingPage/index.js
```diff
@@ -20,25 +20,30 @@ const useStyle = makeStyles(theme => ({
     width: '100%',
     maxWidth: '100%',
     padding: theme.spacing(10, 6),
-    color: blueGrey[800],
+    color: blueGrey[800]
+  },
+  introSection: {
+    padding: theme.spacing(12, 0),
```
qgolsteyn: This spacing is quite large, we should make sure it still works on smaller form factors like mobile

frog/imports/ui/LandingPage/index.js
```diff
@@ -137,39 +148,52 @@ const LandingPage = ({ history }) => {
...
-          onClick={() => history.push('/wizard')}
-        >
-          Try it out now
-        </Button>
+        <div className={classes.alignCenterDiv}>
+          <Button
+            disableFocusRipple
+            disableRipple
+            className={classes.lineButton}
+            onClick={() => history.push('/wizard')}
```
qgolsteyn, line 156: Consider moving `wizard` to a constant as it is used multiple time in this file

frog/imports/ui/LandingPage/index.js
```diff
@@ -137,39 +148,52 @@ const LandingPage = ({ history }) => {
...
-        <Button
-          disableFocusRipple
-          disableRipple
-          className={classes.lineButton}
-          onClick={() => history.push('/wizard')}
-        >
-          Try it out now
-        </Button>
+        <div className={classes.alignCenterDiv}>
+          <Button
```
qgolsteyn, line 152: Consider using the FROG Button

qgolsteyn (changes requested):
General comments:
- We should move all color and typography decisions to a theme object (found in `/client/App`)

RachitMalik12 (commented):
So one aspect is using the common design elements that we have created like the Button etc.. 
We should also probably change the Sign Up modal also to use these common elements. 
Other than that all good!

frog/imports/client/AccountModal/Login.js
AbhinavThukral97: Was part of the old code. Somehow missed this change :) Thanks. Resolved.

frog/imports/ui/LandingPage/index.js
AbhinavThukral97: Changed it to 10, but works well on mobile.

frog/imports/ui/LandingPage/index.js
AbhinavThukral97: Just tried to replace it. However `<Button>` is used at other places with some custom styles and replacing it causes problems. I do agree though, we should be using the FROG button as a standard. I'll try to replace all the buttons on the page soon when redesigning. 
Ps. Suggestion: I find that using the FROG button causes a line break. I suggest adding a display: inline-block.

frog/imports/ui/LandingPage/index.js
AbhinavThukral97: The other links have sub routes after '/wizard'. Would it be a good idea to replace it? As any change would probably affect the sub routes as well unless we're changing /wizard only to say /activity and the url after that remains unchanged. Let me know what you think.

PR #1835, opened 2019-08-21
Warning Light Position Fix

AbhinavThukral97:
The warning light now placed at the top-right area of the config form instead of the page by changing the position to relative for its split parent.


houshuang (approved):

houshuang:
Great, thanks!

PR #1836, opened 2019-08-21
Feature/orchestration_view

qgolsteyn:
This PR updates the UI of Orchestration View.

Important consideration taken in this PR:
- Rather than performing data operations in the view, these operations have been moved to a React Context object. This object provides data in the correct format for the view, as well as a number of functions that mutate the current session (moves to next activity, pause session, etc.).
- The view has been divided into a number of containers responsible for displaying one "domain" (controlling the session, displaying the session steps, etc.)

More work is needed on the following items:

- [ ] Displaying the preview or the dashboard
- [ ] Displaying the session graph or instructions
- [ ] Displaying the overflow menu


houshuang:
Cool! Didn't have a chance to look at your code yet (let me know if there's specific stuff you want feedback on), but your approach sounds interesting.

I think we'll get a much better feeling for how everything works together when we get the central view if you're able to lift it from my messy PR. It would also be cool to see how the graph editor (now improved by Abhinav) looks at the bottom, as well as instructions...

I can do some work on bringing in the dashboards, but did we decide exactly how they will be triggered?

Also need a clear X to get back to the logged in page. We need some kind of settings and links - not sure if this is should be one modal or two... I'm thinking it could be something that was automatically shown when you first enter a session, with links for students, and some quick settings. (Also need to decide about asking people to name their sessions or not)... Also thinking of some kind of button that will show the student URL in really big letters, for teachers to quickly flash on the projector...

Probably lot's I'm missing. Again, the earlier we have something hacky that we can actually try out in practice (even just with our own templates etc) we'll get a better feel.

Thanks again!! Very exciting

PR #1850, opened 2019-08-26
Fixing Operator Movement

AbhinavThukral97:
A quick fix for the operator movement due to the change in position of the graph editor


houshuang (changes requested):
Close, but not quite. First of all, operator is supposed to be on the mouse, not 20 px below, and second once you click, it doesn't stick - you have to click multiple times (maybe the listener is not correct?), and then it disappears. I think the Y written to the database is not correct (needs -300 maybe)?

PR #1851, opened 2019-08-26
Simple Date Formatting for Draft/Session List Items

AbhinavThukral97:
This PR provides a simple function to convert the date formatted as 'DD/MM/YYYY' to one these relevant strings: 'Today', 'x days ago', 'x months ago' or 'x years ago'.
Note: 'x minutes/x seconds' cannot be calculated yet as the only input received is the formatted date of creation.

### DIFF
--- a/frog/imports/client/GraphEditor/Graph.js
+++ b/frog/imports/client/GraphEditor/Graph.js
@@ -22,8 +22,8 @@ const scrollMouse = e => {
 };
 
 const mousemove = e => {
-  // Subtracting Graph Editor's position from the top of the screen
-  store.ui.socialMove(e.clientX, e.clientY - (screen.height - 300 - 100));
+  // We do -100 here because there is 100px above the graph editor
+  store.ui.socialMove(e.clientX, e.clientY - 100);
 };
 
 const Graph = connect(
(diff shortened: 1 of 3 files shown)


frog/imports/client/GraphEditor/Graph.js
```diff
@@ -22,8 +22,8 @@ const scrollMouse = e => {
 };
 
 const mousemove = e => {
-  // We do -100 here because there is 100px above the graph editor
-  store.ui.socialMove(e.clientX, e.clientY - 100);
+  // Subtracting Graph Editor's position from the top of the screen
```
houshuang: This should not be part of this PR.

houshuang (changes requested):

houshuang:
@AbhinavThukral97 	Note that currently, the parsing of dates is done in UserDashboard/data-utils/helpers.js:

```
export const parseDate = (date): Date => {
  return `${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}`;
};
```

this function takes a Date object that also has time. I think it would make sense to move your logic here, and also include minutes ago?

frog/imports/client/GraphEditor/Graph.js
AbhinavThukral97: Ah yeah I branched this by mistake. I'll fix this.

AbhinavThukral97:
I see, I'll check this out.

PR #1858, opened 2019-08-27
Time Formatting for Sessions/Drafts

AbhinavThukral97:


houshuang (approved):

houshuang:
@AbhinavThukral97 	It's OK for now, but I think in the future it's probably better to move this kind of logic to the container - it could also be extracted to Frog-utils, but then called from the container, rather than the presentational component.

PR #1862, opened 2019-08-27
Enabling Publish Validator Hover

AbhinavThukral97:
This PR enables the hover on the Publish button. A lot of components were changed to enable this. Reasons listed as following: 
- The button component was changed to add a 'disabled hover' variant to allow hovering for the Validator component. The hover event for the validator wasn't triggering due to the internal hover component covering the validator.
- Shifting the ErrorList component from the Graph component to the Top Panel component.
- Adding variants to the ErrorList component to retain old error list functionality and to style the new list associated with the TopBar.


houshuang (changes requested):
Looks good for the whole graph, but if you open an activity and make a mistake in the config (for example ac-quiz removing the question), the error message is painted on top of the red button, and flickers very quickly. In the advanced preview (/teacher/preview) it doesn't trigger at all. In the wiki, it looks OK.

Remember we use this component in a lot of places :)

AbhinavThukral97:
I was worried about missing a view. Will fix this.

houshuang (changes requested):
Nothing is displayed in advanced preview
Screenshot 2019-08-28 at 15 17 20

PR #1878, opened 2019-09-03
Guest Session Sign In

AbhinavThukral97:
This PR implements the following task of inserting a pop-up asking "guest" teachers to sign in/continue as guest after reaching the Orchestration View.

Asana task reference: https://app.asana.com/0/1138038533534121/1137467645571898/f

### DIFF
--- a/frog/imports/client/AccountModal/AccountModal.js
+++ b/frog/imports/client/AccountModal/AccountModal.js
@@ -8,7 +8,13 @@ import SignUp from './SignUp';
 import Login from './Login';
 import { resetShareDBConnection } from '/imports/client/App/resetShareDBConnection';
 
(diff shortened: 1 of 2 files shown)


houshuang (changes requested):
The “Continue as guest” button doesn’t do anything (it should dismiss the wizard). Also, we should show the “Create an account” modal as the first view, not the logged in.

An additional requirement I just thought of - if they say “Continue as guest”, we should append ?u={userId} to the URL bar (using the withRouter history) - this means that if they copy the URL they’ll always be able to get in, even if they’re accessing from another computer. And if we detect ?u= in the URL bar, we should not prompt them to sign in (to avoid them having to get this modal every time they access the session).

AbhinavThukral97:
I had to redo changes on another branch, so closing this PR for a new PR. Thanks.

PR #1880, opened 2019-09-04
Legacy Accounts Upgrade on Login

AbhinavThukral97:
This PR requires legacy account users to upgrade their account on Login.

Asana task reference: https://app.asana.com/0/1138038533534121/1137834559839297/f

### DIFF
--- a/frog/imports/client/AccountModal/AccountModal.js
+++ b/frog/imports/client/AccountModal/AccountModal.js
@@ -15,7 +15,7 @@ const AccountModal = ({
 }: {
   formToDisplay: string,
   closeModal?: () => void,
-  variant?: 'guest' | 'default'
+  variant?: 'guest' | 'legacy' | 'default'
 }) => {
   const [form, setForm] = React.useState(null);
   const [showToast, hideToast] = useToast();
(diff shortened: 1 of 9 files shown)


houshuang (changes requested):
Don't move to finished until it has been reviewed :) Not sure how this is supposed to work - if I log

 in using ?login=, which is what we mean by legacy accounts, I go directly to the wizard, and am able to create activities/graphs, when I close it I go back to the wizard (not the UserDashboard, but also not the Learn Landing page). I'm never a single time asked to upgrade my account. 

The desired functionality is that anyone logging in with ?username=... should see a modal asking them to upgrade their account. When they do so, they will have access to all the graphs/sessions etc that already were associated with that account. If they don't do so, they won't be able to proceed. Once they have upgraded, they won't be able to log in with ?login= anymore of course.

houshuang:
Also add for /t/SLUG URLs and good to merge

AbhinavThukral97:
Added the upgrade modal on the top app switch level. Should work on all routes now. Let me know if there are any changes. Thanks!

houshuang (approved):

PR #1881, opened 2019-09-18
Graph Templates

AbhinavThukral97:
This PR implements the functionality of creating, editing templates for graphs.

### DIFF
--- a/frog/imports/api/graphs.js
+++ b/frog/imports/api/graphs.js
@@ -145,7 +145,12 @@ export const addGraph = (graphObj?: Object): string => {
     _id: graphId,
     name,
     ownerId: Meteor.userId(),
-    createdAt: new Date()
+    createdAt: new Date(),
+    templateSource: graphObj
+      ? graphObj.templateSource
+        ? graphObj.templateSource
+        : null
+      : null
   });
   if (!graphObjTmp) {
     return graphId;
--- a/frog/imports/api/graphs.js
+++ b/frog/imports/api/graphs.js
@@ -138,7 +138,12 @@ export const findOneGraphMongo = (id: string) => {
 export const addGraph = (graphObj?: Object, nameVal?: String): string => {
   const graphObjTmp = graphObj && graphObj.graph && upgradeGraph(graphObj);
   const graphId = uuid();
-  let name = graphObjTmp.graph.name ? graphObjTmp.graph.name : 'Unnamed';
+  let name =
+    graphObjTmp && graphObjTmp.graph
+      ? graphObjTmp.graph.name
+        ? graphObjTmp.graph.name
+        : 'Unnamed'
+      : 'Unnamed';
   name = nameVal ? nameVal : 'Unnamed';
   insertGraphMongo({
     ...((graphObjTmp && graphObjTmp.graph) || {}),
--- a/frog/imports/api/graphs.js
+++ b/frog/imports/api/graphs.js
@@ -135,11 +135,11 @@ export const findOneGraphMongo = (id: string) => {
   }
 };
 
-export const addGraph = (graphObj?: Object): string => {
+export const addGraph = (graphObj?: Object, nameVal?: String): string => {
   const graphObjTmp = graphObj && graphObj.graph && upgradeGraph(graphObj);
   const graphId = uuid();
-  const name =
-    (graphObjTmp && graphObjTmp.graph && graphObjTmp.graph.name) || 'Unnamed';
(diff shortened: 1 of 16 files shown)


AbhinavThukral97:
Work in progress

frog/imports/api/graphs.js
```diff
@@ -145,7 +145,12 @@ export const addGraph = (graphObj?: Object): string => {
     _id: graphId,
     name,
     ownerId: Meteor.userId(),
-    createdAt: new Date()
+    createdAt: new Date(),
+    templateSource: graphObj
```
houshuang: why not just put graphObj.templateSource, it will just be undefined if it doesn't exist?

frog/imports/api/templates.js
```diff
@@ -0,0 +1,61 @@
...
+  Templates.insert({
+    name: name,
+    graph: graph,
+    _id: templateId,
+    ownerId: Meteor.userId(),
+    createdAt: new Date()
+  });
+};
+
+export const findTemplate = id => {
```
houshuang: Not sure this function is really needed... And if you do findOne, and there is none, it will just return undefined anyway, so no need to have the if statement etc?

frog/imports/api/templates.js
```diff
@@ -0,0 +1,61 @@
...
+  if (id) {
+    const template = Templates.findOne(id);
+    return template;
+  }
+  return null;
+};
+
+export const updateTemplate = (id, graph) => {
+  try {
```
houshuang: Under which condition would this fail? Silently swallowing errors is dangerous (although I know I do it some places :))

frog/imports/api/templates.js
```diff
@@ -0,0 +1,61 @@
...
+    const doc = Templates.findOne(id);
+    const template = Templates.update(
+      { _id: doc._id },
+      { $set: { graph: graph } }
+    );
+    return template;
+  } catch (err) {
+    console.warn(err);
```
houshuang: Please give more context, like "Failed updating template etc etc".

houshuang (approved):
Hi. This looks quite good. A few general comments:

When you do “Save as template”, it should also set the templateSource of that graph, so that the menu now says “Update template” instead of “Save template” while on that graph.

On the Template page I think we need some text near the top like “Create new graph based on template” otherwise people will think they’re opening and editing the actual template.

Templates should also appear on the Recents page (under Drafts)

houshuang (changes requested):
Sorry I approved by mistake.

frog/imports/api/templates.js
AbhinavThukral97: Ah, this was when I was debugging. I forgot to revert back.

frog/imports/api/templates.js
AbhinavThukral97: Sure!

frog/imports/api/templates.js
AbhinavThukral97: I had planned to do more with this function. Just in case we need to make any changes in all the graphs created using the template, so I was planning to fetch the graphs and add a graphs object containing ids. I don't think we need this as of now though, so I'll minify this!

AbhinavThukral97:
> Templates should also appear on the Recents page (under Drafts).

Just to clarify, adding the list of templates after the list of drafts? Under the Drafts heading itself?

Sure, I'll add a message bar component for this then.

frog/imports/api/graphs.js
AbhinavThukral97: True! I'll fix this.

houshuang:
> > Templates should also appear on the Recents page (under Drafts).
> 
> Just to clarify, adding the list of templates after the list of drafts? Under the Drafts heading itself?
> 
> Sure, I'll add a message bar component for this then.

No, a new heading "Templates" below the two existing columns.

houshuang (changes requested):
We should automatically set the name of the generated graph to the template name as a default.

Why do you use blue for "Graph based on template", and green for "Template updated"?

houshuang (changes requested):
For the removeTemplate stuff, you need to make sure the code only runs on the server, so you use if Meteor.isServer(). Otherwise, even when you try to do a Meteor.call, it will run the code on the client to simulate, and that will produce an error.

AbhinavThukral97:
> Why do you use blue for "Graph based on template", and green for "Template updated"?

This is because the "graph based on template" is an info message, while "template update" is a success message.

houshuang (approved):

PR #1882, opened 2019-09-24
Adding Archive Functionality

AbhinavThukral97:
This PR allows Drafts, Sessions, and Templates to be archived and creates a section for archived items for other actions.

### DIFF
--- a/frog/imports/client/TeacherView/index.js
+++ b/frog/imports/client/TeacherView/index.js
@@ -14,6 +14,7 @@ import { Sessions } from '/imports/api/sessions';
 const TeacherViewRunner = withRouter(
   withTracker(({ match, history }) => {
     const user = Meteor.user();
+    let session;
     if (match?.params?.slug) {
       session = Sessions.findOne({
         slug: match.params.slug
--- a/frog/imports/client/TeacherView/index.js
+++ b/frog/imports/client/TeacherView/index.js
@@ -14,7 +14,6 @@ import { Sessions } from '/imports/api/sessions';
 const TeacherViewRunner = withRouter(
   withTracker(({ match, history }) => {
     const user = Meteor.user();
-    let session;
     if (match?.params?.slug) {
       session = Sessions.findOne({
         slug: match.params.slug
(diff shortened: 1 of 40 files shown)


AbhinavThukral97:
WIP

AbhinavThukral97:
This PR also contains changes from the Graph Templates PR as we wanted to add the archiving functionality to templates as well.

houshuang:
Merge conflicts

lfaucon:
@AbhinavThukral97 After merging there is a problem with a double definition of the function `parseTemplateData` in the file `imports/client/UserDashboard/data-utils/helpers.js`. Can you fix it ?

frog/imports/client/TeacherView/index.js
```diff
@@ -14,7 +14,6 @@ import { Sessions } from '/imports/api/sessions';
 const TeacherViewRunner = withRouter(
   withTracker(({ match, history }) => {
     const user = Meteor.user();
-    let session;
```
lfaucon: Why this change? Doesn't it create syntax errors?

frog/imports/client/UserDashboard/components/ArchivesPage/index.js
```diff
@@ -0,0 +1,179 @@
...
+                  );
+                }
+              )}
+            </List>
+          </Paper>
+        </Grid>
+      )}
+    </Grid>
+  );
+};
```
lfaucon: I don't like this file much. It seems to me that it is a near identical copy-paste of the `RecentPage` component. For code quality we should avoid such things. Do you see a way to reuse the same component instead of duplicating the code?

lfaucon (changes requested):
Seems like a great feature to add. Thanks for this work!

Looks like this PR contains all the commits of #1883 and additional commits. I suggest merging #1883 first for this reason.

frog/imports/client/UserDashboard/components/ArchivesPage/index.js
AbhinavThukral97: The only difference between this and the `RecentPage` component is the display limit of 6 data rows. We can definitely add a variant in the RecentPage to accommodate this but my reasoning for keeping the two components separate was the future scope of expanding the RecentPage to add more functionality.

lfaucon (approved):

PR #1886, opened 2019-10-03
Does not filter out published graphs from draft list

lfaucon:


houshuang:
Can you please think a bit about this and discuss with Jenny? We spent quite a bit of time trying to think about how we could make the workflow much easier than what it was (with all the Unnamed #29 etc). I wrote it up here https://docs.google.com/document/d/1xrfEJdoGi7umlqEHvvCUdqXdB3opPfxYs-Ffyv7qR-w/edit?usp=drive_web&ouid=107702703184747130690

However, there were two components that still haven't been implemented. One was the ability to clone a session (create a new graph from a session). This will soon be merged. The other, much more complex thing, was moving all the testing functionality into the graph editor, meaning that you wouldn't have to publish a graph just to test it. I'd still like to do this sometime, but no idea when. 

So in the meantime, if you discuss with Jenny and think this is the best way to solve the issue short-term, I'm OK with merging.

lfaucon:
I think one problem we add previously was also because graphs and session graphs were mixed within the same lists. With the change I just made you will actually have only one non-session graph and if you click several times on Publish several sessions. I find the workflow not so bad this way. We could had the possibility to delete a session directly. 

I did this change as I heard Jenny complain multiple times of losing her graphs because of publishing them. The same things will happen to CS411 students next week. I agree that a button "Go back to graph and edit" from the session view would be as good or better a solution as what I implemented in this PR.

We can easily filter again the graphs labelled "published" in the future, for now I think that is a positive improvement.

lfaucon:
The problem right now is:

- spend 15min designing your graph
- publish it
- want to make some modification? No, you have to create a brand new graph

That is why I made this change which we need before Tuesday. If the other solution is coming before that deadline I am ok to wait for and not merge this PR at all.

houshuang:
Makes perfect sense for now.

PR #1889, opened 2019-10-07
Show the content area full screen in PopupLayout

lfaucon:
In the orchestration view the teachers sees the SidePanel (Left), the OG (Bottom) and the Content (Top-right) which is either dashboards or activity previews. This PR add a button to make the content full screen. It is needed when looking at dashboard on a small screen.


houshuang (dismissed):
This looks good.

houshuang:
Note that we now have one button in the top right to open in a new tab, and one in the bottom left to make full screen (toggle). Also, the first button is always there, whereas the second only appears once the session has started... Maybe we can consolidate these two buttons and make them more consistent? Main point is not to confuse teacher. @AbhinavThukral97 thoughts? 
Screenshot 2019-10-09 at 10 10 17

PR #1892, opened 2019-10-08
Wiring up ErrorPage if SLUG not found

AbhinavThukral97:
This PR wires up the Error page when accessing undefined sessions using 't/SLUG', with and without login for TeacherView.


lfaucon:
How does the page actually comes up? When typing the wrong url as a teacher, FROG redirects me towards the previous sections that was opened. Can you tell me steps to reproduce?

AbhinavThukral97:
The error page is rendered if the user account has no sessions to redirect to or an attempt to access the 't/SLUG' is made without login.

houshuang:
Looks good to me. I think we should probably also fail if people try to go to the wrong slug instead of defaulting to a random slug the user has - that was useful before when we were switching users all the time. @lfaucon what do you think?

houshuang (approved):

lfaucon:
I agree with @houshuang on showing the error page instead of redirecting to an existing previous slug 

I have not been able to replicate triggering the error page. I must be missing something?...

AbhinavThukral97:
I wasn't sure where the redirect functionality was needed. New PR up to fix this.
@lfaucon This can be replicated with any wrong slug now.

PR #1894, opened 2019-10-09
Hide graphs generated by wizards

houshuang:


lfaucon (approved):
Are you sure about this change ? 
In the class yesterday, I was actually thinking that when a teacher uses a template for a session, this same template could be a good starting point to start using the graph editor. Does this PR's change goes against this?

houshuang:
The current graph editor is very difficult to use for people who don't know it, and it's easy to create invalid graphs. Therefore, I do not want normal teachers who click on FROG because they saw a link on Twitter (yes I know I'm optimistic) to ever say "Oh, what's this". I'm basically "hiding" the graph editor for people who don't know it exists (people in CS411 etc). 

Another issue is that it should be possible to use the wizards to create the starting point for a graph that you edit yourself, I totally agree, although that should be in the graph editor somewhere. OR maybe on the overflow menu it could have "Create new blank graph / create new graph from template" and it would directly take you to the graph editor.

PR #1895, opened 2019-10-09
Removing the default session redirect for unknown slugs

AbhinavThukral97:
Removed the functionality that redirected to an existing session for unknown slugs and modified the Error Page a little bit. Now shows the error message for any wrong slug.


lfaucon:
@AbhinavThukral97 please fix the linting error from Eslint

lfaucon (approved):

PR #1901, opened 2019-10-10
Feature/control from quiz

lfaucon:


houshuang:
We should really have tests for all operators, they are quite quick to write.

PR #1912, opened 2019-11-05
Component for replacing the React Mosaic Library

AbhinavThukral97:


houshuang:
@lfaucon Can you have a look at this? The idea is to replace React Mosaic with something very simple, since it's currently increasing our bundle size with .5 MB, and has a lot of functionality (like resizing) which we never use... 

I think we still want some kind of simple title bar, similar to what we have now. Also not sure about the odd-number layouts, like 3 - should we have 3 side by side or one taking half, and the two others taking two quarters like now (downside is that we currently can't choose which goes where). 

Screenshot 2019-11-15 at 08 35 52

lfaucon:
@AbhinavThukral97 This looks great! Good work

For the view with 3 components, I would prefer one big (half of the screen) and two small (quarter of the screen). Similarly to what we have now already.

Screenshot from 2019-11-15 09-49-31

Also, when we replace ReactMosaic, it would be good to have the same topbar on all the children (Like on the screenshot: "Chat (individual/Chen Li)")

@houshuang Are we sure that the students never used the resizing?

Future improvements could be:
- Children can be toggled into full screen
- Children can be minimized (For example if we have 4 windows and 2 are minimized, then the 2 left should fit as a two-children layout)

houshuang:
Rather than focusing on resizing and minimizing, I'd rather we move towards
what I've been hoping for for a very long time, which is to stop treating
all activity types the same, but instead special case instructions, social
presence and chat, etc.

AbhinavThukral97:
I've changed the 3-window view as mentioned and added the functionality for title/topbar using a window component similar to the one currently being used. 
I'll start replacing the Mosaic component with this and test it out quickly.
Let me know if there are any other changes/suggestions :)

houshuang:
Looks good to me. Perhaps a different color for the top bar to distinguish? You might also need to add some margin to the body, but you'll see once you try implementing it and rendering real activity types (the preview window is the quickest way of seeing it in action with many different activity types/window configurations).

houshuang:
@AbhinavThukral97 I tried cleaning up a tiny bit, removing the node libraries and the commented code. Currently the Preview page crashes immediately, not quite sure why? If we can get everything to work well and the checks to be green, we can merge.

AbhinavThukral97:
I've fixed the Preview section and cleaned the rest of the code. I think we are good to merge now. Let me know if there are any further changes.

houshuang:
@lfaucon This looks good to me, can you also review since this touches on very crucial code? The advantage of this will be that we can get rid of a 0.5 MB dependency, and it might also make it easier to customize the interface more in the future.

frog/imports/client/Preview/Content.jsx
```diff
@@ -233,51 +251,23 @@ const ContentController = ({
...
-                ]
-              : users.map((name, idx) => [name, instances[idx]])
-          )}
-        />
+        <ActivitySplitWindow>
+          {showDash
+            ? [
+                ['dashboard', 'dashboard'],
+                ...users.map((name, idx) => [name, instances[idx]])
+              ].map(arr => {
```
lfaucon: To be more easily read. This could be re-written 
`([name, instance ]) => { ... } `

frog/imports/client/Preview/index.js
```diff
@@ -155,8 +154,4 @@ class PreviewPage extends React.Component<any, any> {
 
 PreviewPage.displayName = 'PreviewPage';
 
-export default (props: any) => (
-  <ErrorWrapper>
-    <PreviewPage {...props} />
-  </ErrorWrapper>
-);
```
lfaucon, line 162: Why did you remove the `ErrorWrapper`? Does this have something to do with the windowing component?

frog/imports/client/StudentView/Runner.jsx
```diff
@@ -163,15 +163,12 @@ const Runner = ({ path, activity, sessionId, object, paused }) => {
   );
 
   return (
-    <MosaicWindow
-      toolbarControls={[<div key={1} />]}
-      draggable={false}
+    <ActivityWindow
       key={activity._id}
-      path={path}
```
lfaucon, line 170: What was this props `path` that you remove?

frog/imports/client/Preview/Content.jsx
lfaucon: I did it

frog/imports/client/StudentView/Runner.jsx
AbhinavThukral97: 'path' was a prop for React Mosaic's component that's used to expand split windows.

frog/imports/client/Preview/index.js
AbhinavThukral97: I don't remember removing this. Maybe this was a change merged after I started this branch.

frog/imports/client/StudentView/Runner.jsx
lfaucon: ok :)

lfaucon (approved):

PR #1915, opened 2019-11-05
Cloning Sessions

AbhinavThukral97:
Adding the session cloning functionality


houshuang:
@AbhinavThukral97 Status?

frog/imports/api/sessions.js
```diff
@@ -231,6 +232,39 @@ export const updateOpenActivities = (
...
+        ? 'Template'
+        : 'Graph';
+      if (sessionType === 'Single Activity') {
+        store.setWizardConfig(session.simpleConfig.config);
+        history.push(`/wizard/${session.simpleConfig.activityType}`);
+      } else if (sessionType === 'Template') {
+        store.setWizardConfig(session.simpleConfig.config);
+        history.push(`/wizard/${session.simpleConfig.activityType}`);
+      } else {
+        const graphObj = JSON.parse(graphToString(session.fromGraphId));
```
houshuang: We are very careful about what activities/operators/connections that we publish from the backend to the frontend, otherwise you would end up downloading 20,000 operators whenever you logged into chilifrog.ch as a teacher. All this is determined through server/teacherImports and teacher/main.

So the simple answer is that the client does not have the activities/operators etc of a random session without explicitly subscribing to them (which is done using setId in the store etc). However, the much simpler solution is to move all this logic to the backend, which always has access to the entire database. So there should be a call Meteor.call('cloneSession', session.id), which returns the new session id, and all the copying etc should happen on the backend.

houshuang (changes requested):

frog/imports/api/sessions.js
AbhinavThukral97: Thanks for this! My latest commit should resolve this issue. Currently,
- Clone session for a session created using graphs redirects to a duplicate source graph
- Clone session for a session created using Single Activities/Templates redirects to the wizard configuration view with the pre-filled values from the session. 
The user can review/tweak these and publish to run a new (almost identical) session. (I believe this was the required functionality.)
This however, confuses me about returning new session ids. Could you please explain a bit more on what is required after this?

PS. I think if we want this action to redirect to duplicate graph/configuration views, we could rename the action from 'Clone Session'? Maybe "Duplicate Source Graph" for graphs or "Duplicate Session Configuration" for Single Activities and Templates would be more appropriate?

PR #1917, opened 2019-11-15
Compute score for quiz leaderboard based on configured correct answers

lfaucon:
Before this PR the quiz leaderboard would only count the student progress (useless). Now it works by counting the number of correct answers as specified by the teacher in the quiz config.

### DIFF
--- a/frog/imports/internalActivities/ac-quiz/utils.js
+++ b/frog/imports/internalActivities/ac-quiz/utils.js
@@ -78,6 +78,23 @@ export const computeProgress = (
   return nAnsweredQuestions / questions.length;
 };
 
+export const isCorrect = (
+  qData: { [idx: string]: boolean, text: string },
+  qConfig: { answers: Object[], text: boolean }
+) => {
+  return !qConfig.answers.some((x, i) => !!x.isCorrect !== !!qData[`${i}`]);
+};
+
+export const computeScore = (
+  questions: { answers: Object[], text: boolean }[],
+  form: { [qIdx: number]: { [idx: string]: boolean, text: string } }
+) => {
+  return questions.reduce(
+    (acc, q, qIdx) => (isCorrect(form[qIdx], q) ? acc + 1 : acc),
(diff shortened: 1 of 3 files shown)


frog/imports/internalActivities/ac-quiz/utils.js
```diff
@@ -78,6 +78,23 @@ export const computeProgress = (
   return nAnsweredQuestions / questions.length;
 };
 
+export const isCorrect = (
+  qData: { [idx: string]: boolean, text: string },
+  qConfig: { answers: Object[], text: boolean }
+) => {
+  return !qConfig.answers.some((x, i) => !!x.isCorrect !== !!qData[`${i}`]);
```
houshuang, line 85: I'm trying to parse this line, but find it hard. Why are you turning both x.isCorrect and qData... into booleans before comparing them?

houshuang (changes requested):
This would really benefit from a test or two, should be very quick to write, just feed in some form configs with example data. My main concern is that it should never crash a form no matter what the form config is, and of course secondarily, it should ideally display the correct score :) Unfortunately the logic is too dense for me to be able to tell if it's correct/safe just by looking at it.

frog/imports/internalActivities/ac-quiz/utils.js
lfaucon: `isCorrect ` is either false of undefined for wrong answers (configured by teachers)
`qData[`${i}`]` is either false or undefined for not selected answers (by students from ActivityRunner) 

They are undefined by default and take value false only if for example the user selects, then unselects the answer or the teacher ticks then unticks the configuration box.

`!qConfig.answers.some((x, i) => !!x.isCorrect !== !!qData[`${i}`])` 
It says if there is "some" in the config answers for which the boolean "isCorrect" is not the same as the boolean "reactive data" then return false

Thanks for always reminding me to add tests. You're completely right :)

